PR #5, opened 2017-02-01
Added lines between steps and images

KennethBates:
Formatting had images beside text instead of beneath text.


johnsimons:
Thank you 😄

PR #11, opened 2017-02-02
PowerShell for registry key removal

dragon788:
This prevents hitting the wrong one and breaking something.


johnsimons:
Thanks for the PR 😄

PR #15, opened 2017-02-10
Example Octopus script option for Service Accounts

dragon788:
This applies to any type of account, not necessarily just a service account.

After beating my head against my desk for a week or so being frustrated that our CI pipeline relied on somebody manually entering the credentials to grant accounts the ability to run scheduled tasks or logon as a service, I came across Carbon and after a bit of poking around figured out how to script granting  these permissions. The only "caveat" is that the account granting the permission needs to be a member of Local Administrators (I think) in order to have the "delegation" permission.


PaulStovell:
Thanks for this, I'm sure people will appreciate it!

PR #22, opened 2017-02-22
fix for #18 clarifications on octo exe step version options

### LINKED ISSUE #18: Contradictory and confusing detail in 'Creating releases' page
See page: https://octopus.com/docs/api-and-integration/octo.exe-command-line/creating-releases

This page has lead to some confusion and a fair bit of lost effort and time, there are 2 issues.

1. The [options] section states that switches are provided in `--SWITCH=VALUE` when the examples describe `--SWITCH _space_ VALUE`
2. The [options] section states that the --package switch can override the "Version number to use for a package in the release" however it overrides the package for a **STEP** in the release process.

The former issue certainly added some confusion as I was unsure as to whether iI was passing the switches to octo.exe correctly, however the latter issue is by far the most obscure and confusing.

Please can the documentation be clarified to prevent further confusion by people using octo.exe, including your support staff (see http://help.octopusdeploy.com/discussions/questions/7042-defaultpackageversion-package).

Also is it possible to explain/describe the reasons for why the "name of the step" needs to be specified; when it could reference a step which has nothing todo with deploying packages (e.g. a custom deployment step). When creating a release through the UI you select the **packages** not the packages for each step, hence the confusion here.

NickJosevski:
Fixes #18

### DIFF
--- a/docs/api-and-integration/octo.exe-command-line/creating-releases.md
+++ b/docs/api-and-integration/octo.exe-command-line/creating-releases.md
@@ -149,14 +149,21 @@ octo create-release --project HelloWorld --version 1.0.3 --server http://octopus
 
 ## Specifying the package version {#Creatingreleases-Specifyingthepackageversion}
 
-The step name and version pair will take the highest precedence supplied via `--package stepName:1.0.3`, if that version cannot be found then the version supplied by the `--packageversion` will be next, finally falling back to the what is found specified as specified by `--packagesFolder`.
+For each step that has a package, the version is determined in the following order:
 
-### Option --packageversion
+ 1. The step name matches a --packageVersion parameter or a file filename found by --packagesFolder
+ 1. The package id matches a --packageVersion parameter or a file found by --packagesFolder
+ 1. The value from the --package parameter
+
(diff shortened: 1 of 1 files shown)


docs/api-and-integration/octo.exe-command-line/creating-releases.md
```diff
@@ -149,18 +149,25 @@ octo create-release --project HelloWorld --version 1.0.3 --server http://octopus
 
 ## Specifying the package version {#Creatingreleases-Specifyingthepackageversion}
 
+The step name and version pair will take the highest precedence supplied via `--package stepName:1.0.3`, if that version cannot be found then the version supplied by the `--packageversion` will be next, finally falling back to the what is found specified as specified by `--packagesFolder`.
```
droyad: Not quite right. Also the --packagesFolder parameter may override an earlier parameter if specified last.

It's more like: 
For each step that has a package, the version is determined in the following order:
1) The step name matches a `--packageVersion` parameter or a file filename found by `--packageFolder`
2) The package id matches a `--packageVersion` parameter or a file found by `--packageFolder`
3) The value from the `--package` parameter

If there are duplicate names/ids as a result of the `--packageVersion` and `--packageFolder` parameters, the last one specified is used.

PR #32, opened 2017-02-27
Update ssh-keygen command in format Octopus is looking for

TimGebhardt:
The updated command is what works on Ubuntu 16.04.  I think it's also a bit more future-proof assuming that Octopus itself isn't flexible on the format of the thumbprint.


zentron:
Looks good! Thanks for the contribution @TimGebhardt. Making these documented commands as cross-platform as possible is always a bit of an art. 👍

PR #38, opened 2017-03-08
Added info on passwords when adding new user

KennethBates:
Clarified the need for a password when creating a new user in Octopus - that it's the UsernamePassword provider password only, and that it isn't the user's AD password.


johnsimons:
@KennethBates @slewis74 is this ready to be merged ?

PR #39, opened 2017-03-14
Point links to headings in this document

flcdrg:
Otherwise they aren't really doing anything.

Also add list items for Azure Web App and Cloud Services

### DIFF
--- a/docs/deployment-targets/index.md
+++ b/docs/deployment-targets/index.md
@@ -12,41 +12,43 @@ Deployment targets are added to an environment from the Environments page.
 
 A deployment target can be one of the following types:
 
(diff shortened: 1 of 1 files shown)


johnsimons:
Hi @flcdrg 

Thanks heaps for the PR, we actually have support for auto magically generate ToCs using the `!toc` syntax, see https://github.com/OctopusDeploy/docs#toc
This way we don't need to maintain the toc :)

Here is the fix - https://github.com/OctopusDeploy/docs/commit/2463918fabe2dd1ac82300b4ecb302b949762bfc

So thank you for pointing out the issue, really appreciated.

Cheers
John

PR #45, opened 2017-03-24
Update names for Package Feed, Id and Version vars

alfhenrik:
No longer named `Octopus.Action.Package.NuGet*`


johnsimons:
@alfhenrik when was this changed ? Do we need to use partials here ?

PR #53, opened 2017-04-06
had to add path for other 3 directories to path

jonathanmccracken:
I ran into this problem when moving my Home folder from C:\ to D:\ ... seems like I need to modify the path for each of these before I was able to get this to work


johnsimons:
@thebigmoosey could you please review this PR ?

johnsimons:
I have manually merged this with a few modifications, see https://github.com/OctopusDeploy/docs/commit/e42e2e040cbfc63eddc5c4440d30862ce711779e

Thanks @jonathanmccracken

jonathanmccracken:
no problem thanks for updating the docs!

PR #58, opened 2017-04-18
Update Octopus.Server.exe configure args

matt-richardson:
As it was missing `xFrameOptionAllowFrom`, but also a lot more.


matt-richardson:
@NickJosevski - not sure if there was a specific reason that this wasn't up to date?
@slewis74 - not sure if we want to mention the `azureAD*` and `googleApps*` properties in this one?

NickJosevski:
I don't know what happened there, possibly where I stopped 1 day, and went on to the next one, not having finished it.

Thanks for finding it.

PR #63, opened 2017-04-24
Add page about deploying via DSC

matt-richardson:


slewis74:
Would be worth adding info about 

`az account set --subscription 'xxxxxxxxxxx'` in case the user has multiple subscriptions

Also, in the sample vmDnsName is TentacleVM, but the value isn't allowed to have upper case characters so we should use a better example.

PR #72, opened 2017-05-07
Adding missing publicSettings properties for AzureRm VM extension

paulmarsy:
From decompiling the extension I believe these are the only missing properties:
- MachinePolicy
- Tenants
- TenantTags

The values used for the example should be consistent with the rest of the documentation.


johnsimons:
Thank you heaps @paulmarsy

PR #79, opened 2017-05-24
Use correct default port for polling tentacles

mrled:
As far as I can tell from other documentation, such as https://octopus.com/docs/installation/installing-tentacles/polling-tentacles, the default port for polling tentacles is 10943, and this was the case for my Octopus server, but this document was using 10934 (with the 3 and 4 transposed).

(Interestingly, the 10934 port is used elsewhere in the documentation, for instance, in this document: https://octopus.com/docs/administration/upgrading/upgrading-from-octopus-2.6/in-place-upgrade-install-over-2.6. There, the tentacle was a *listening* tentacle, and the 10934 port may have been chosen by incrementing the default listening tentacle port of 10933 by 1... perhaps this was the cause of the confusion?)

### DIFF
--- a/docs/how-to/troubleshoot-polling-tentacles.md
+++ b/docs/how-to/troubleshoot-polling-tentacles.md
@@ -4,7 +4,7 @@ description: How to troubleshoot problems with polling Tentacles.
 position: 5
 ---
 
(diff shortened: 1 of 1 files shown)


johnsimons:
Hi @mrled 
Thanks for fixing this up 😄

PR #96, opened 2017-06-22
Add docs for new certificate commands

matt-richardson:
Related to OctopusDeploy/Issues#3574
Related to OctopusDeploy/OctopusTentacle#10


stopthatastronaut:
Looks good to me - the meat of it is pulled straight from the /help content anyway, right?

matt-richardson:
Yeah, it is pulled from the help text.
I really want us to have version specific help text here, as it changes all the time, but that sounds like a reasonable amount of automation pain... Probably needs to be using docker, which means we need [this trello card](https://trello.com/c/LMuUF2GO/312-create-docker-images-for-all-versions-from-30-onwards) to be done first.

PR #101, opened 2017-07-10
Explain how to disable progression

pawelpabich:


pawelpabich:
[LifecycleProgressionEvaluator throws when the Lifecycle only has optional Phases](https://trello.com/c/JGRAcHbl/663-lifecycleprogressionevaluator-throws-when-the-lifecycle-only-has-optional-phases)

PR #104, opened 2017-07-12
Include escape sequences

Myster:
Include escape sequences so slashes & square brackets show correctly in paths


johnsimons:
Thanks @Myster

PR #107, opened 2017-07-13
Copy home folder including sub folders

jmalczak:
Hi Guys

I was running move tentacle script but it didn't copy any of the sub folders of home folder. I have added -Recurse switch for copy-item command.


johnsimons:
Thanks @jmalczak

PR #133, opened 2017-10-17
[WIP] 4.0 lifecycle docs

alfhenrik:


robpearson:
I tested this locally and the images aren't coming through correctly. Something seems amiss. Content wise it seems fine but I think there might be a problem w/ the file naming etc. 

@alfhenrik Did you get the same experience?

PR #153, opened 2017-12-07
Enh integration bits

Dalmirog:


Dalmirog:
[Add reference to Swagger in our docs](https://trello.com/c/3VXW5Mly/483-add-reference-to-swagger-in-our-docs)

PR #164, opened 2018-01-18
A bit of simplification of getting a snapshot

IlyaAI:
The points 3 and 4 in section 'Get a snapshot from your running Octopus Server' might be merged in signe step:
`dotmemory.exe get-snapshot Octopus.Server`


droyad:
Awesome, thanks!

PR #172, opened 2018-02-06
Add new variable for machine names

hnrkndrssn:
Waiting for https://github.com/OctopusDeploy/OctopusDeploy/pull/1670 to be merged first

- [ ] Update **actual** version variable was introduced in

### DIFF
--- a/docs/deployment-process/variables/system-variables.md
+++ b/docs/deployment-process/variables/system-variables.md
@@ -26,7 +26,7 @@ Deployment-level variables are drawn from the project and release being deployed
 |**`Octopus.Deployment.Id`** <br/>The ID of the deployment | *deployments-123*|
 |**`Octopus.Deployment.Name`** <br/>The name of the deployment | *Deploy to Production*|
 |**`Octopus.Deployment.PreviousSuccessful.Id`** <br/>The ID of the previous successful deployment of this project in the target environment | *deployments-122*|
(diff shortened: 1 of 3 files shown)


docs/deployment-process/variables/system-variables.md
```diff
@@ -26,7 +26,7 @@ Deployment-level variables are drawn from the project and release being deployed
 |**`Octopus.Deployment.Id`** <br/>The ID of the deployment | *deployments-123*|
 |**`Octopus.Deployment.Name`** <br/>The name of the deployment | *Deploy to Production*|
 |**`Octopus.Deployment.PreviousSuccessful.Id`** <br/>The ID of the previous successful deployment of this project in the target environment | *deployments-122*|
-|**`Octopus.Deployment.Machines`** <br/>Ids of machines being targeted by the deployment. | machines-123,machines-124|
+!partial <deploymentMachineVariables>
```
droyad: I think the current approach is to not use partials anymore, rather just note the version number. Check with @leecarlon

alfhenrik: Oh, OK. I was not aware of that

leecarlon: Rob's correct, but a bit ahead of the curve. We won't need to use partials for old versions when the new versioned version of the docs go live (ETA next week, with details to follow), and in fact, at that point, using partials will only really be useful for pre-release content.

alfhenrik: OK, I'll revert it to just mention the version it was introduced in in the descripition

PR #173, opened 2018-02-07
Updated usage stats and when added

thekatstevens:
Not sure if it's too specific/scary. We could probably ditch a lot of the specific lines with some wording like:

> We also track usage of some features (such as multi-tenancy, LetsEncrypt, guest accounts et. al.) to see what user uptake is like. These are only true/false values of whether you are using the feature, and contain no identifying information about _how_ you are using the feature.

And just use that to capture new inclusions that fall under that wording (like how adding new step types falls under the "the types of step you use and how many" wording).

### DIFF
--- a/docs/administration/security/outbound-requests.md
+++ b/docs/administration/security/outbound-requests.md
@@ -32,7 +32,7 @@ During installation, Octopus will ask you whether you want to check for new rele
 
 ![](/docs/images/3048073/3277613.png "width=500")
 
-When the checkbox "Automatically check for new Octopus releases" is checked during installation, Octopus will make a HTTPS request to the `octopusdeploy.com` domain every 8 hours. This request includes:
+When the checkbox "Automatically check for new Octopus releases" is checked during installation, Octopus will make a HTTPS request to the `octopus.com` domain every 8 hours. This request includes:
 
 1. The current Octopus Deploy version number that you are running
 2. A unique installation ID (read more below)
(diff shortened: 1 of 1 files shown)


docs/administration/security/outbound-requests.md
```diff
@@ -50,15 +50,41 @@ In addition, if you also check the box "Help improve Octopus by sending usage st
 4. The number of each of the different types of deployment steps that you have
 5. The number of releases and deployments that you have done in the last 90 days
 6. The number of users that you have
-7. (As of 2.6) A SHA1 hash of your license key serial number
-8. (As of 2.6) The amount of memory Octopus is using
-9. (As of 2.6) The % of CPU Octopus is using on average
-10. (As of 3.12) The time to your first deployment
+
```
matt-richardson: Maybe make this a table, rather than headings?

Or should this be partials, based on the version?

docs/administration/security/outbound-requests.md
```diff
@@ -32,14 +32,14 @@ During installation, Octopus will ask you whether you want to check for new rele
 
 ![](/docs/images/3048073/3277613.png "width=500")
 
-When the checkbox "Automatically check for new Octopus releases" is checked during installation, Octopus will make a HTTPS request to the `octopusdeploy.com` domain every 8 hours. This request includes:
+When the checkbox "Automatically check for new Octopus releases" is checked during installation, Octopus will make a HTTPS request to the `octopus.com` domain every 8 hours. This request includes:
 
 1. The current Octopus Deploy version number that you are running
 2. A unique installation ID (read more below)
```
matt-richardson: markdown spec says that all entries in a numbered list should be `1.`

docs/administration/security/outbound-requests.md
thekatstevens: Once docs are snapshotted we can just edit this rolling forward and trust the archives to be accurate for old versions. For now, I might follow your other suggestion of making it a table and then we can snapshot it off this month, tidy it up and power to the future. I'll amend this shortly

PR #177, opened 2018-02-26
formatting change

NickJosevski:


NickJosevski:
it still didn't fix the intending problem I noticed

PR #232, opened 2018-07-05
Clarify use of variable syntax in email subject and body

danielfischeroctopus:
You can use basic variable substitution syntax in the email subject and the extended syntax in email body.
Provided some explicit clarification for this and provided link.


leecarlon:
Hi Daniel, the build failed on this one. It looks like you've linked to the GitHub content instead of the content on the docs site. You should be able to fix the link by removing "docs/blob/master" from the link "docs/blob/master/docs/development-process/variables/variables-substitution-syntax.md".

PR #239, opened 2018-08-06
[WIP] security review doc

leecarlon:
Added text to the top level security doc to make it as discoverable as possible to people looking at the security docs. I'll end the actual link when we know the best way to host the PDF.


thekatstevens:
Yeah I think this is fine, add the doc to `/wwwroot/documents/security-review.pdf` or something, and link to it 👍

PR #244, opened 2018-08-16
Add more worker info to performance pages

MichaelJCompton:


docs/administration/performance.md
```diff
@@ -57,24 +57,28 @@ Octopus Servers do quite a lot of work during deployments, mostly around package
...
 
 - scale up by controlling the **task cap** and providing more server resources as required
 - scale out using [Octopus High Availability](/docs/administration/high-availability/index.md)
+- scale out using [Workers](/docs/administration/workers.index.md)
 
-We are planning a third option for scaling your Octopus Server:
+We are planning a fourth option for scaling your Octopus Server:
 
-- sharding across multiple Octopus Servers using [Spaces](https://octopus.com/blog/octopuses) - planned for Octopus 4.x
+- dividing up your Octopus environment using [Spaces](https://octopus.com/blog/octopus-spaces-blog-series-kick-off) - planned for Octopus 2018.9.0
```
michaelnoonan: This will probably be 2018.11.0 planned for release in October.

docs/deployment-process/performance.md
```diff
@@ -141,4 +142,8 @@ Steps and actions in your deployment process can be configured to start after th
 
 Similarly to [parallel targets](#parallel-targets), running too many steps in parallel can cause your Octopus Server to become the bottleneck and make your deployments take longer overall.
 
+### Consider how much deployment work the Octopus server is doing {#server-work}
+
+Some steps, like Azure deployments and AWS steps, [run on a worker](docs/administration/workers.md#where-steps-run).  By default, thats the [built-in worker](/docs/administration/workers#built-in-worker) in the Octopus server.  That means the step invokes a (or many) Calamari processes on the server machine to do the deployment work.  That workload can be shifted off the server and onto [workers](/docs/infrastructure/workers.md).  See this [blog post](https://octopus.com/blog/workers-performance) for a way to begin looking at workers for performance.
```
michaelnoonan: thats -> that's
a (or many) -> could invoke one or more Calamari processes
That workload -> These workloads?

docs/infrastructure/workers/index.md
```diff
@@ -47,4 +47,6 @@ Default pools attached to cloud targets allow co-location of workers and targets
...
+
+The number of parallel worker invocations any one deployment task will make is controlled by the [Octopus System Variable](/docs/deployment-process/variables/system-variables.md) `Octopus.Action.MaxParallelism`.  For example, if your server task cap is 2 and the `Octopus.Action.MaxParallelism` variable is at its default value of 10, then you could be running 20 instances of the built-in worker concurrently (it's actually more because there could also be parallel steps within those deployment tasks).  If external workers are added to the default pool, then the workload is shared across those workers: a single worker will get the whole workload that the built-in worker used to do, two workers might get half each, etc.  Remember that in an HA setup each node has a task cap and can invoke a built-in worker locally, so you'll need to provision workers to at least match your server nodes, otherwise, you'll be asking each worker to do the sum of what the HA nodes were previously doing.
```
michaelnoonan: Oh no! I didn't realised we'd further overloaded this variable with more meanings. I think we should change that design.

michaelnoonan: Or at least talk it out! :)

docs/deployment-process/performance.md
leecarlon: > That workload can be shifted off the server and onto [workers]
That should be either: 
That workload can be moved off the server onto a [Worker]...
or:
These workloads can be moved off the server onto [Workers]...

docs/infrastructure/workers/index.md
leecarlon: It might be worth breaking out the HA info into a subsection with a heading so it stands out for HA users.

MichaelJCompton: It's not really overloaded further.  It's just the same options as we always had.  In terms of how many worker invocations can run at once it's equivalent - just that where the workers are running has changed.

I put up an issue about splitting out the meanings of this var https://github.com/OctopusDeploy/Issues/issues/4667 and this one about controlling worker parallelism https://github.com/OctopusDeploy/Issues/issues/4815

It'll be quite the thing to change to allow something like task caps specified on workers, or task caps specified per pool (meaning the workers in this pool each have a task cap of X).  There'd potentially be complications around how we spin up the parallel using octo parallel (cause that's way up in DeploymentTaskController, but at the moment we don't make worker choices till way down in the ActionDispatcher) and we might have to consider how we distribute packages otherwise the workers with packages will end up being a bottleneck.  But the tricky bit might be the waiting cause we have to implement some sort of wait queue for threads waiting for workers, but then that'd have to be responsive to cancellation, potential reordering if we go with task priorities, etc.  

I think we should look at it, but I expect it will be subtle to get really robust.

michaelnoonan: Of course! I see what you mean about the variable: we've just moved the location from the Octopus Server to a Worker, and the variable does just what it did before. Let's take this conversation offline. :)

PR #247, opened 2018-08-31
Fixing broken Bootstrap Switch link in credits.md

draevin:
Used github link instead of website (http://bootstrapswitch.site/) based on precedent set by others, such as Font Awesome


droyad:
Thanks @draevin

PR #257, opened 2018-10-10
Added advice on updating SSL\TLS Certificates

jburger:
I recently worked with a customer on updating an existing web portal binding SSL\TLS certificate, and we both found that the documentation didn't really make it clear what the correct approach was. 

This PR adds a link-able subheading close to the place in the documentation that myself and the customer naturally gravitated towards. The content under this heading outlines the approach to tackling this scenario.


jburger:
[Updating the SSL Cert for the Web portal](https://trello.com/c/7r1U5BPf/271-updating-the-ssl-cert-for-the-web-portal)

PR #282, opened 2018-11-05
Update credits / third party dependencies

matt-richardson:
* includes all nuget packages from (non-test) octopus.sln projects
* includes all nodejs packages used in the newportal 
* includes a link to the license

outstanding questions:
* We used to list `wix` - should we still?
* We used to list `Microsoft Windows, Visual Studio and .NET`. Should we still? If so, we should probably add `.NET Core`
* We used to list `TeamCity, ReSharper and WebStorm` - should we still? If so, we should probably add `JetBrains Rider`
* any third party CSS we are using?
* any third party icons? images? we are using?
* we should probably add Tentacle, Calamari, Halibut, Nevermore. Anything else?


wordlee:
A while ago there was some discussion around moving this list into the product itself instead of hosting it on the docs. I don't have a preference, but thought I'd mention it so it isn't forgotten.

PR #283, opened 2018-11-05
And add angular dependencies as well (the old portal)

matt-richardson:


zentron:
Waiting for an excuse to do a minor bump since it removes functionality 👍

PR #300, opened 2018-12-20
Renovating upgrading guidance

michaelnoonan:
This is the first step in some work I plan to do over the next day or so. Firstly I wanted anyone coming here via the LTS blog posts etc to have something closer to what they expect to see.

Next step is to clarify all the version numbers, most of which incorrectly refer to `3.x` nowadays. :)

@wordlee if this is a close enough starting place, can you merge it down?


wordlee:
Looks great, @michaelnoonan . My only comment is, upgrading feels like such a top level concern (to me at least), that all on-premises customers will need at some point, I wonder if it's worth giving it a separate area in the docs. so /docs/upgrading/ instead of /docs/administration/upgrading/

PR #321, opened 2019-01-25
Updates to show compatible versions for 2019.1 and greater

jburger:
- [x] Octopus server
- [x] Octopus.Client
- [x] Calamari
- [x] Tentacle
- [x] Team City Plugin

### DIFF
--- a/docs/api-and-integration/compatibility.md
+++ b/docs/api-and-integration/compatibility.md
@@ -15,11 +15,14 @@ The table below outlines the backward compatibility between Octopus Server and r
 | 2018.2 ➜ 2018.12 | 4.30.7 ➜ latest        | 3.5 ➜ latest    | 3.0 ➜ latest | 3.3 ➜ latest    |
 | 2019.1* ➜ latest | 5.0.0 ➜ latest         | 3.5 ➜ latest    | 4.0 ➜ latest | coming soon      |
 
-> &ast; **Note**: older versions of some libraries and plugins _may_ work with **2019.1** and higher _only if_ the [default space](https://g.octopushq.com/default-space) is enabled and such integrations are only used against that space. To make use of other spaces, please upgrade.
-> Forwards compatible versions are: 
-> - Octopus client: 4.30.7 ➜ 4.47.0
-> - Calamari: 3.5 ➜ 4.12.0
-> - Tentacle: 3.0 ➜ 3.25 
-> - TeamCity Plugin: 3.3 ➜ latest
-> 
-> However, **2019.1** also carries some breaking changes to the API so care must be taken when upgrading from a prior version that existing integrations are not effected. There is a comprehensive rundown of these changes available in our [downloads section](https://octopus.com/downloads/compare?from=2018.12.1&to=2019.1.0).
+### &ast; Partial forwards compatibility
+
+Older versions of some libraries and plugins _may_ work with **2019.1** and higher _only if_ the [default space](https://g.octopushq.com/default-space) is enabled and such integrations are only used against that space. To make use of other spaces, please upgrade.
+
+However, **2019.1** also carries some breaking changes to the API so care must be taken when upgrading from a prior version that existing integrations are not effected. There is a comprehensive rundown of these changes available in our [downloads section](https://octopus.com/downloads/compare?from=2018.12.1&to=2019.1.0).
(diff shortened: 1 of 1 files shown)


docs/api-and-integration/compatibility.md
```diff
@@ -6,10 +6,18 @@ position: 130
...
+| Octopus Server    | Octopus.Client Octo.exe | Calamari         | Tentacle      | TeamCity Plugin  |
+| --------------    | ----------------------- | ------------     | ------------  | ---------------  |
+| 3.2               | 3.2                     | 3.0 ➜ latest    | 3.2           |                  |
+| 3.3               | 3.3 ➜ latest           | 3.3              | 3.0 ➜ latest | 3.3 ➜ latest    |
+| 3.4               | 3.3 ➜ latest           | 3.4              | 3.0 ➜ latest | 3.3 ➜ latest    |
+| 3.5 ➜ 2018.2     | 3.3 ➜ latest           | 3.5 ➜ latest    | 3.0 ➜ latest | 3.3 ➜ latest    |
+| 2018.2 ➜ 2018.12 | 4.30.7 ➜ latest        | 3.5 ➜ latest    | 3.0 ➜ latest | 3.3 ➜ latest    |
+| 2019.1 ➜ latest  | 5.0.0 ➜ latest         | 4.12.1 ➜ latest | 4.0 ➜ latest | coming soon      |
+
+> Note: older versions of some libraries and plugins _can_ work with 2019.1 and higher _only if_ the [default space](https://g.octopushq.com/default-space) is enabled and such integrations are only used against that space. To make use of other spaces, please upgrade.
```
jburger: Not sure that this section is entirely clear

docs/api-and-integration/compatibility.md
```diff
@@ -6,10 +6,18 @@ position: 130
...
-| 3.5 ➜ 2018.2    | 3.3 ➜ latest           | 3.5 ➜ latest | 3.0 ➜ latest | 3.3 ➜ latest    |
-| 2018.2 ➜ latest | 4.30.7 ➜ latest        | 3.5 ➜ latest | 3.0 ➜ latest | 3.3 ➜ latest    |
+| Octopus Server    | Octopus.Client Octo.exe | Calamari         | Tentacle      | TeamCity Plugin  |
+| --------------    | ----------------------- | ------------     | ------------  | ---------------  |
+| 3.2               | 3.2                     | 3.0 ➜ latest    | 3.2           |                  |
+| 3.3               | 3.3 ➜ latest           | 3.3              | 3.0 ➜ latest | 3.3 ➜ latest    |
+| 3.4               | 3.3 ➜ latest           | 3.4              | 3.0 ➜ latest | 3.3 ➜ latest    |
+| 3.5 ➜ 2018.2     | 3.3 ➜ latest           | 3.5 ➜ latest    | 3.0 ➜ latest | 3.3 ➜ latest    |
+| 2018.2 ➜ 2018.12 | 4.30.7 ➜ latest        | 3.5 ➜ latest    | 3.0 ➜ latest | 3.3 ➜ latest    |
+| 2019.1 ➜ latest  | 5.0.0 ➜ latest         | 4.12.1 ➜ latest | 4.0 ➜ latest | coming soon      |
```
TomPeters: Maybe put an asterisk next to the Octopus Server version on this row, and an asterisk next to the Note below to tie them together.

TomPeters: `Calamari: 4.12.1 ➜ latest`. Shouldn't this be the same as `2018.2 -> 2018.12`? We haven't changed calamari in spaces so I wouldn't expect this version requirement to have changed either.

docs/api-and-integration/compatibility.md
TomPeters: Probably always worth highlighting that there are a small number of other breaking changes that breaks forwards compatibility (I'm thinking about things like the Teams API). People should check the breaking changes in `2019.1` to see if they would be affected and therefore need to upgrade.

TomPeters: Maybe this would be clearer if it were under a different section with a header like `Partial Compatibility`. Then you could list the versions in a table like the above compatibility matrix, but it becomes obvious that you get limited compatibility if your versions match the second table.

PR #334, opened 2019-02-14
Spaces migrator

NickJosevski:


docs/api-and-integration/octo.exe-command-line/import.md
```diff
@@ -43,6 +43,8 @@ Common options:
                              enabled, a key of API-GUEST can be used. This
                              key can also be set in the OCTOPUS_CLI_API_KEY
                              environment variable.
+      --destinationPackageFeedSpaceId=VALUE
```
wordlee: @NickJosevski This one will get overwritten by @Octobob the next time it runs, unless the changes have also been made in OctopusClients. If the changes have been made in OctopusClients, they'll get pulled in. Also, I had to double take on this one and the next and suggest a slight change to make the meaning clearer [Optional if not using the Spaces Feature].

PR #339, opened 2019-02-21
Update credits / license attribution

Octobob:
An automated update of the credits / license attribution file, based on the latest code in https://github.com/OctopusDeploy/OctopusDeploy. 

Please review for
* syntactical correctness (ie, markdown looks correct)
* logical correctness (ie, when reading it, it makes sense)
* license conformance (we agree with the license in question and we are able to redistribute it)

Created by TeamCity project 'Automated Documentation Updates::Update credits.md (OSS license attribution file)', build 41


docs/credits.md
```diff
@@ -982,6 +983,7 @@ Octopus Deploy is made possible thanks to the following great third-party produc
...
 | timers-browserify                                 | J. Ryan Stinnett                                                          | [https://github.com/jryans/timers-browserify](https://github.com/jryans/timers-browserify)                                                                                                                                  | [MIT](https://opensource.org/licenses/MIT)                                                                          |
+| TimeZoneConverter                                 | Matt Johnson                                                              | [https://github.com/mj1856/TimeZoneConverter](https://github.com/mj1856/TimeZoneConverter)                                                                                                                                  | [Undefined]()                                                                                                       |
```
michaelnoonan: This is a bit interesting @matt-richardson - it's actually MIT, but this says undefined.

matt-richardson: Its because he's [using](https://github.com/mj1856/TimeZoneConverter/blob/master/src/TimeZoneConverter/TimeZoneConverter.csproj#L10) the `PackageLicense` tag, which nuget has [stuffed around with right royally](https://github.com/NuGet/Home/issues/7509). Also, we're using a really old nuget, as upgrading to a new version of nuget is a _nightmare_. Think we'll need to tackle it though :(

matt-richardson: Manually tweaked, to give me a weeks grace

PR #362, opened 2019-03-22
Static IP Update

alexrolleyoctopus:
Fixing some out of date instructions and some general wording updates.


alexrolleyoctopus:
That's a good point. We do have a [pricing page](https://octopus.com/pricing/overview) but it doesn't mention Static IP. I've reached out to the cloud pricing team to get that added there, and I've removed the price from this page.

PR #366, opened 2019-03-27
Add UDP 1434 to outbound rules

ryanrousseau:
Configurations using named instances with the SQL Server Browser Service make outbound requests to UDP 1434.


OctopusBob:
LGTM

PR #369, opened 2019-04-04
Updating cli docs for octo.exe

Octobob:
An automated update of the command line docs for octo.exe.

Created by TeamCity project 'Automated Documentation Updates::Update CLI Docs', build 40

### DIFF
--- a/docs/api-and-integration/octo.exe-command-line/push-metadata.md
+++ b/docs/api-and-integration/octo.exe-command-line/push-metadata.md
@@ -1,0 +1,79 @@
+---
+title: push-metadata
+description: Pushes package metadata to Octopus Server.
+---
+
+Pushes package metadata to Octopus Server.
+
+**push-metadata options**
+
+```text
+Usage: octo push-metadata [<options>]
+
+Where [<options>] is any of:
+
+Package metadata pushing:
+
+      --package-id=VALUE     The ID of the package; e.g. MyCompany.MyApp
+      --version=VALUE        The version of the package; defaults to a
+                             timestamp-based version
+      --metadata-file=VALUE  Octopus Package metadata Json file.
+      --replace-existing     If the package metadata already exists in the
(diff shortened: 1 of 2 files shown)


docs/api-and-integration/octo.exe-command-line/push-metadata.md
```diff
@@ -0,0 +1,79 @@
+---
+title: push-metadata
+description: Pushes package metadata to Octopus Server.
+---
```
wordlee: To keep the sorting straight in the TOC:

```suggestion
position: 285
---
```

slewis74:
I hadn't added this documentation page because the octo.exe command is really only for use by the build server extensions. We can add it for completeness if we have to, but I expect nobody would ever use it directly. Especially without knowing the JSON format of the file is requires. I would not bother with this page, or if we do go ahead with it include a link to the OctopusClient code repo for the object structure the file must contain.

matt-richardson:
This gets auto-generated from the command line help, on the theory that if its worth having command line help, its worth having in the docs

wordlee:
We shouldn't add anything to the docs unless we think it will be helpful. Matt can you exclude this page from being auto-generated?

slewis74:
Thought more about this after a conversation with AppVeyor just now. We've been the ones build the build server plugins so far, but if we're not the ones building the build server plugin this doc has value.

slewis74:
I think we should also add a link to https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Client/Model/PackageMetadata/OctopusPackageMetadata.cs in the OctopusClient code for the description about the file, so it's clear what JSON has to be in the file.

wordlee:
I'm going to close this one as we have another PR that supersedes this one (I'll make the same edits in the new PR). And when the link has been added to OctopusClient code it'll get picked up next time OctoBob does his thing.

PR #392, opened 2019-05-08
Added initial API examples

ryanrousseau:
I saw that our docs site supports tabbed code examples and that makes it really enticing to host the examples here.


wordlee:
Currently waiting on a fix for the code sample switcher on docs to be picked up by docsync. This can be merged at any time, but note, switching between samples won't work until that fix is in place.

PR #399, opened 2019-05-10
Add some information about script module usages

matt-richardson:
Relates to https://github.com/OctopusDeploy/Issues/issues/5557


docs/deployment-examples/custom-scripts/script-modules.md
```diff
@@ -68,7 +68,11 @@ Make sure to select a **Role**, an **Environment** and to put a **Step Name**
 ![](script-modules-deployment-release.png)
 
 
-## PowerShell Script Modules{#ScriptModules-PowerShell}
+## Using Script Modules{#ScriptModules-using}
```
wordlee: It might nice to add a list of links directly to the different languages here. It would save some scrolling :)

docs/deployment-examples/custom-scripts/script-modules.md
```diff
@@ -68,7 +68,11 @@ Make sure to select a **Role**, an **Environment** and to put a **Step Name**
...
+
+Each language has a slightly different syntax for using the Script Module. Please see the language specific section below.
+
+### PowerShell Script Modules{#ScriptModules-PowerShell}
```
wordlee: While technically subsections of the Using Script Modules section. I try to keep major headings at level two because of the way they present on the docs site, and I try to avoid level 4s altogether because the heading sizing is small than the body text.

matt-richardson: I'll fix here... But we should probably fix the heading sizes, rather than work around it in the docs.

PR #410, opened 2019-05-28
Fix migrator partial-export command options

Waldo000000:
- Remove the `--sourceSpaceId` option from the doco for the migrator's `partial-export` command. This option doesn't exist.
- Raised [via support](https://help.octopus.com/t/partial-export-with-sourcespaceid/23182)
- Also tweaked a related option description


Waldo000000:
An alternative is to actually implement support for such a partial export, i.e. see https://github.com/OctopusDeploy/issues/issues/5593.

Waldo000000:
Alternatively, @michaelnoonan is having a look at getting these created automatically: https://github.com/OctopusDeploy/docs/pull/411

PR #413, opened 2019-05-28
User feedback re Accessing ARM Template Output

wordlee:
Octopus.Action[ArmTemplateStepName].Output.AzureRMOutputs[Foo]

Should be:

Octopus.Action[Arm Template Step Name].Output.AzureRmOutputs[Foo]

Two mistakes, the spaces in the step name, and the casing of the m in RM. 

@robpearson or @benPearce1, if either one of you could confirm this, that would be very helpful.


wordlee:
Thanks, Ben! To make it even clearer, I've added:

Note, you need to replace **Arm Template Step Name** with the name of your ARM step template.

PR #426, opened 2019-06-12
Add create and deploy release examples

ryanrousseau:


ryanrousseau:
> 
> 
> Looks good, Ryan. Just one thought, do you think we need to provide any guidance/warnings about running these scripts?

I was thinking that too. I'll send a separate PR to add some instructions to the index of the examples folder.

ryanrousseau:
> 
> 
> Looks good, Ryan. Just one thought, do you think we need to provide any guidance/warnings about running these scripts?

I might have been thinking of something else. Either way I've added both to this PR. I added some info about getting the api key, using Octopus.Client, and running the c# scripts to the index as well as some usage notes and warnings to the example pages. Is that in line with what you were thinking?

PR #446, opened 2019-06-26
Update index.md

vivianpiketmedia:
Added link to broader description on firewall.


wordlee:
Thanks for the PR. However, our users are familiar with this term. If at any point we think adding a definition will help them, we'll add it to the documentation rather than linking to an external website that could change without notice.

PR #451, opened 2019-07-02
Update title of deregister-worker.md

ryanrousseau:


ryanrousseau:
> 
> 
> LGTM.
> 
> I was initially concerned that this part would be automatically updated, but I think your changes are in the part that is kept as is - it's just the first code block that gets replaced.

Ah, good to know.  I didn't realize those pages had generated content. I just noticed two Deregister From links on the nav menu when I was tracking down something for a customer today.

PR #465, opened 2019-07-15
Updated wording and explained Octopus Server

michaelnoonan:
Trying to help a customer who got confused when they couldn't target the Octopus Server directly. https://help.octopus.com/t/script-console-suggestion-run-on-octopus-server-option/23522/3

### DIFF
--- a/docs/administration/managing-infrastructure/script-console.md
+++ b/docs/administration/managing-infrastructure/script-console.md
@@ -48,6 +48,13 @@ Besides making it easy to run a script on many servers, the other advantage of u
 
 You cannot target the Octopus Server with the Script Console. If you want to run ad-hoc tasks on your Octopus Server, you should install a Tentacle agent on your Octopus Server and register that Tentacle as a worker or deployment target.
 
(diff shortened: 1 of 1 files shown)


docs/administration/managing-infrastructure/script-console.md
```diff
@@ -42,6 +44,10 @@ Besides making it easy to run a script on many servers, the other advantage of u
 
 ![](images/3277919.png)
 
+## Targeting the Octopus Server
+
+You cannot target the Octopus Server with the Script Console. If you want to run ad-hoc tasks on your Octopus Server, you should install a Tentacle agent on your Octopus Server and register that Tentacle as a worker or deployment target.
```
wordlee: Are there any caveats we should add to this? i.e., is it a good idea to install Tentacle on the Octopus Server?

alexrolleyoctopus: @wordlee Yep, that is our recommended approach as it is more secure than running processes on the built in worker.

Ideally you want it off server entirely, however this is a good baseline/starting point without getting too complex

michaelnoonan:
I added some security notes. Thanks for the review!

PR #467, opened 2019-07-18
Linux tentacle setup script

tw17:
Added instructions for running the setup script, which in turn means we dont need to run a whole bunch of commands manually.


wordlee:
The QuickStart Scripts could be moved into an other document that way advanced users can still access them and reuse/modify them. I'm always wary of putting scripts in with the basic configuration instructions as some new users (hopefully not many) might just scroll down and copy paste the script in. It would be good to add caveats/assumptions about when/how to use the scripts too.

tw17:
I think the QuickStart scripts could be given a different name, or just a bit of a description that they are examples of how to automate the setup.

PR #477, opened 2019-07-23
further explanation for deployment time

benPearce1:


docs/deployment-examples/azure-deployments/deploying-a-package-to-an-azure-web-app/using-deployment-slots-with-azure-web-apps.md
```diff
@@ -53,7 +53,7 @@ New-AzureRMWebAppSlot -Name #{WebSite} -Slot Staging -ResourceGroupName #{Resour
 ```
 
 :::hint
-The reason for the first line, which removes the Staging Slot, is to ensure we are deploying to a clean slot. This can significantly reduce the time taken for deployments with a large number of files.
+The reason for the first line, which removes the Staging Slot, is to ensure we are deploying to a clean slot. This can significantly reduce the time taken for deployments with a large number of files as it avoids having to calculate the set of files which have acutally changed by comparing time stamps or checksums. For deployments with a smaller number of files, this will may not be any quicker.
```
wordlee: Minor tweak to the language, @benPearce1 :)
```suggestion
The first line of the script removes the Staging Slot to ensure we are deploying to a clean slot. 
This can significantly reduce the time taken for deployments with large numbers of files as it avoids having to calculate the set of files which have actually changed by comparing time stamps or checksums. For deployments with a smaller number of files, this will not be any quicker.
```

PR #478, opened 2019-07-24
Update apt + repo urls

tothegills:
- Did a rename of "Quick start scripts" -> "Automation scripts"
- Removed net core instructions, I don't think it's actually a pre-req because we bundle
- Removed blurb about the portal not having a way to add linux Tentacle

### DIFF
--- a/docs/infrastructure/deployment-targets/linux/tentacle/index.md
+++ b/docs/infrastructure/deployment-targets/linux/tentacle/index.md
@@ -10,7 +10,7 @@ Welcome to Linux Tentacle early access. This page provides information about get
 
 - Octopus Server 2019.5.7 or newer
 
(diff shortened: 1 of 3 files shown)


tw17:
We still need to keep the bit about netcore, we still rely on it being installed, even though we bundle the app we still use it to handle the framework dependencies so we dont have 10+ different .deb and .rpm version per release
in the configure bash script we link to https://dotnet.microsoft.com/download/linux-package-manager/rhel/runtime-current

docs/infrastructure/deployment-targets/linux/tentacle/index.md
```diff
@@ -9,13 +9,11 @@ Welcome to Linux Tentacle early access. This page provides information about get
 ## Requirements
 
 - Octopus Server 2019.5.7 or newer
-- [.NET Core 2.x](https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=netcore2x) installed on the Linux machine
```
tw17: we need to keep this bit.

tothegills: Updated based on our discussion

PR #500, opened 2019-08-19
Add comma in simple variable transform example

istrupin:
There was a comma missing in the json example for the simple variable transform -- I added it here.


matt-richardson:
Thanks @istrupin!

PR #501, opened 2019-08-20
Update to unit file

twerthi:
Following the doco to set up the Linux tentacle as a service, I found that the service would not start because of the working directory didn't exist.  Changing the working directory from default to tentacle allowed my Linux tentacle to start correctly as a service.  Being a Linux noob, it is entirely possible that I did something incorrect during the installation process, which is why I'm requesting a review of the change.


wordlee:
I'll let @tw17 take this one :)

PR #512, opened 2019-09-02
Pre-releases and deployment notes

slewis74:
Updates to better explain why changes are not accumulated across pre-releases.

Source 

https://help.octopus.com/t/release-changes/23784/6

### DIFF
--- a/docs/deployment-process/releases/deployment-notes.md
+++ b/docs/deployment-process/releases/deployment-notes.md
@@ -12,7 +12,7 @@ Deployment notes summarize these changes by rolling up the [release notes](/docs
 
 ## Versioning and Pre-releases  {#Versioning}
 
-When it comes to release changes and deployment notes, [pre-release versions](/docs/packaging-applications/create-packages/versioning.md#semver) are handled differently to other versions. Deployments of a pre-release will only ever include changes for that single version, they do not accumulate across versions like other scenarios.
+[Pre-release versions](/docs/packaging-applications/create-packages/versioning.md#semver) are handled differently to other versions for release changes and deployment notes. Deployments of a pre-release will only ever include changes for that single version, they do not accumulate across versions like other scenarios.
 
 The reason for this is that in some scenarios different pre-release versions could be getting deployed to an environment where they co-exist. Consider deployments for an application installer or a NuGet package. You could have multiple branched versions of those co-existing in an environment at any given point, as an example, so the accumulation of work across releases has to be considered differently.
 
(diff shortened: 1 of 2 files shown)


docs/deployment-process/releases/deployment-notes.md
```diff
@@ -10,6 +10,14 @@ Deployment notes summarize these changes by rolling up the [release notes](/docs
 
 ![Deployment notes](images/deployment-notes.png)
 
+## Versioning and Pre-releases  {#Versioning}
+
+When it comes to release changes and deployment notes, [pre-release versions](/docs/packaging-applications/create-packages/versioning.md#semver) are handled differently to other versions. Deployments of a pre-release will only ever include changes for that single version, they do not accumulate across versions like other scenarios.
```
wordlee: I'd change the structure a little bit here, to put the subject first:
```suggestion
[Pre-release versions](/docs/packaging-applications/create-packages/versioning.md#semver) are handled differently to other versions for release changes and deployment notes. Deployments of a pre-release will only ever include changes for that single version, they do not accumulate across versions like other scenarios.
```

docs/deployment-process/releases/deployment-notes.md
```diff
@@ -10,6 +10,14 @@ Deployment notes summarize these changes by rolling up the [release notes](/docs
...
+
+The reason for this is that in some scenarios different pre-release versions could be getting deployed to an environment where they co-exist. Consider deployments for an application installer or a NuGet package. You could have multiple branched versions of those co-existing in an environment at any given point, as an example, so the accumulation of work across releases has to be considered differently.
```
wordlee: Slight tweak (reject if this doesn't help):
```suggestion
This is because in some scenarios different pre-release versions could be getting deployed to an environment where they co-exist. As an example, consider deployments for an application installer or a NuGet package, you could have multiple branched versions of those co-existing in an environment at any given point, so the accumulation of work across releases has to be treated differently.
```

docs/deployment-process/releases/deployment-notes.md
```diff
@@ -10,6 +10,14 @@ Deployment notes summarize these changes by rolling up the [release notes](/docs
...
+
+Octopus therefore takes the position that pre-releases are a "work in progress" and will fail under the banner of "full" release when that are completed/merged. At that point the changes are then considered for accumulation across the "full" versions.
```
wordlee:
```suggestion
Octopus takes the position that pre-releases are a *work in progress* and will fail under the banner of *full* release when that are completed/merged. At that point, the changes are then considered for accumulation across the *full* versions.
```

slewis74: Updated, and fixed a couple of spelling mistakes we'd both missed before

PR #514, opened 2019-09-04
udpate url for tar file

NickJosevski:


NickJosevski:
Hi @wordlee thanks for checking on that.

Here's the next PR after we ship Tentacle, that'll always point to a "latest" link, and no versions in the other commands so will be evergreen. https://github.com/OctopusDeploy/docs/pull/515

PR #516, opened 2019-09-05
Explain how Linux Tentacle auto upgrade works

tothegills:


docs/infrastructure/deployment-targets/linux/tentacle/index.md
```diff
@@ -102,6 +101,14 @@ Start the Tentacle interactively by running:
     sudo systemctl enable tentacle
     ```
 
+## Automatic Tentacle upgrades
+Linux Tentacle can be upgraded via the Octopus portal from the Infrastructure > Deployment Targets screen. The upgrade attempts to find a package manager capable of performing the upgrade, and then falls back to extracting a `tar.gz` archive to the Tentacle installation folder.
```
wordlee: Being picky, is upgraded or updated? (It probably is upgraded, but I thought I'd check).

wordlee: adding formatting:
```suggestion
Linux Tentacle can be upgraded via the Octopus portal from the {{Infrastructure > Deployment Targets}} screen. The upgrade attempts to find a package manager capable of performing the upgrade, and then falls back to extracting a `tar.gz` archive to the Tentacle installation folder.
```

tothegills: I think update is more accurate, but we call it "Upgrade" in the portal.

PR #533, opened 2019-09-26
Added `-SkipDependencies` to prevent Powershell from returing a circular reference error

droyad:


droyad:
[Octopus.Client can't be installed via PowerShell](https://trello.com/c/EqwtKNLP/1527-octopusclient-cant-be-installed-via-powershell)

PR #539, opened 2019-10-04
Introducing Runbooks

LittleBigMoosey:
This adds some basic documentation for the initial Runbooks feature that we're planning to release early next week (behind a feature toggle).

Decided to list this under the `Deployment Process` section, as this contains a lot of the shared docs and mentions Projects/Variables etc. that are all related to Runbooks. We can move this around once we get more feedback from the early-access customers.

### DIFF
--- a/docs/deployment-process/index.md
+++ b/docs/deployment-process/index.md
@@ -38,7 +38,7 @@ This is an example of a very simple process, the following sections go into more
 
 ## Projects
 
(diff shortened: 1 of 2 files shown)


LittleBigMoosey:
There's no way to really import a sample runbook at this stage, so instead I've added a quick example line for now, and we can iterate on this once we start getting some feedback from customers and we have a better idea of how they might best use this feature.

PR #541, opened 2019-10-09
Patch to move the powershell core content into a new section

wordlee:


wordlee:
Created against the wrong branch....

PR #586, opened 2019-12-04
WIP - Runbook Triggers Documentation

alialmaktoum:
# Background
Adds a page for Scheduled Runbook Triggers and cleans up various sections of documentation to do with Runbooks and Triggers.

# Results

image

# Pre-requisites
- [x] I have considered whether this should be a patch or wait for a minor.
- [x] I have considered appropriate testing for my change.

PR #576, opened 2019-11-18
Adding Runbooks Publishing section

MarkSiedle:
This PR adds a publishing section to Runbooks and cleans up some limitations that have since been actioned.

PR #588, opened 2019-12-06
Enh dtargettriggers

taniarichardsonoctopus:
**Related to bug bash card for Triggers: https://trello.com/c/0quWXT2a/16-inconsistent-terminology-for-triggers**

We're trying to keep the naming in product the same as in the docs.

@wordlee  I wanted to ask whether we can set up a redirect for this page as I've changed the path.  Also, is there anything else I may have missed.  Thanks


wordlee:
I'll add the redirect and update internal links to the page.

wordlee:
I've added the redirect and fixed the internal links. I wonder if this PR includes the commits from #576 and #586 should we close those PRs? cc @MarkSiedle @alialmaktoum

PR #589, opened 2019-12-13
Re-wording Basic Examples

twerthi:
Previous wording made it look like when using create-release without specifying --version, it would use whatever the highest NuGet package version was in the project.  As we know, the Release Versioning project setting specifies a specific step in the process and uses that steps package for the version number.  A customer contacted us about the confusion, they expected whatever NuGet package in the process had the highest number, that was the number that got used.

I'm not married to my wording, just trying to clarify the documentation 😄


twerthi:
Good suggestion, @matt-richardson.  I've added that part.

matt-richardson:
Nice one. :shipit:

PR #595, opened 2020-01-16
Add a short explanation and workaround warning alert for Azure Cloud Services Timeout issue

ankithkonda:
Related to https://github.com/OctopusDeploy/Issues/issues/6111 

We will not be able to fix that issue, as the change would need to happen on Azure Cloud Services. This docs update will inform customer who encounter Timeout Issues or slow re-packaging of a work around, however as this work around disables variable substitution, we're using a warning alert.


ankithkonda:
[Exclude /useCtpPackageFormat switch](https://trello.com/c/ExDmu7SO/2737-exclude-usectppackageformat-switch)

wordlee:
Sorry for the late response. I pushed a couple of minor line edits on master.

PR #598, opened 2020-01-17
Structure docs for deployments, release management, and ops

wordlee:
With the introduction of operations processes, we need to structure sections (specifically the deployment process sections) of the docs so they are from a just deployment process perspective.
This PR:
- Updates the Getting Started page.

Splits the Deployment Process section into:
- Projects
- Deployment Process
- Release Management
- Operations Runbooks


wordlee:
The structure is in place. Merging to master and will keep working on the language in master.

PR #602, opened 2020-01-28
Add hint to rolling-deployments.md to mention use of Window size

harrisonmeister:
Add hint to rolling-deployments.md to mention use of Window size with Octopus.Action.MaxParallelism

### DIFF
--- a/docs/deployment-patterns/rolling-deployments.md
+++ b/docs/deployment-patterns/rolling-deployments.md
@@ -34,7 +34,7 @@ The window size controls how many deployment targets can be deployed to at once.
 
 :::hint
 **Window size with Octopus.Action.MaxParallelism**
-If you include the variable `Octopus.Action.MaxParallelism` in your Project with a value higher than your **Window size** set in a rolling deployment, you will find that its value is no longer respected. This is expected behavior as Octopus also uses this variable to limit the number of deployment targets on which the rolling deployment step will run concurrently. A warning will also be printed in the Task Log.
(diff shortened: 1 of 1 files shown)


docs/deployment-patterns/rolling-deployments.md
```diff
@@ -32,6 +32,11 @@ The window size controls how many deployment targets can be deployed to at once.
 - A window size of 1 will deploy to a single deployment target at a time. Octopus will wait for the step to finish running on deployment target A before starting on deployment target B
 - A window size of 3 will deploy to a three deployment targets at a time. Octopus will wait for the step to finish running on deployment targets A, B *or* C before starting on deployment target D
 
+:::hint
+**Window size with Octopus.Action.MaxParallelism**
+If you include the variable `Octopus.Action.MaxParallelism` in your Project with a value higher than your **Window size** set in a rolling deployment, you will find that its value is no longer respected. This is expected behavior as Octopus also uses this variable to limit the number of deployment targets on which the rolling deployment step will run concurrently. A warning will also be printed in the Task Log.
```
wordlee: Slight tweak to remove ambiguity (I had to read the sentence twice to understand which *it* was being referred to.
```suggestion
If you include the variable `Octopus.Action.MaxParallelism` in your Project with a value higher than your **Window size** set in a rolling deployment, you will find that the `Octopus.Action.MaxParallelism` value is no longer respected. This is expected behavior as Octopus also uses this variable to limit the number of deployment targets on which the rolling deployment step will run concurrently. A warning will also be printed in the Task Log.
```

harrisonmeister: Thanks @wordlee, that suggestion is great (Ironically, I had that _originally_ but removed as I thought referencing the variable twice in quick succession might be a bit too wordy, but it removes the ambiguity) 👍

PR #605, opened 2020-01-31
update docker image tag

emilol:
Relates to OctopusDeploy/Issues#4878

Our docs point to a specific tag (over a year old) that had a bug, a client recently raised said bug after following the docs.

We're looking into automating this for future releases of OctopusCLI


emilol:
[Bitbucket and Octopus](https://trello.com/c/vO6Xv6o9/2821-bitbucket-and-octopus)

PR #606, opened 2020-02-10
move image version to include

emilol:
Relates to OctopusDeploy/Issues#4878 and #605

Moving the image version to an include so we can update it without needing to know about any specific usages.

TODO: modify the build/deploy process to update the version in this file when a new image is pushed to docker hub.


emilol:
I've added a test to octofront for it, just about to push that too 👍

emilol:
[Bitbucket and Octopus](https://trello.com/c/vO6Xv6o9/2821-bitbucket-and-octopus)

PR #696, opened 2020-04-14
Increase SQL Server reqs

michaelnoonan:
Lines up with our intent for 2020: https://octopus.com/blog/raising-minimum-requirements-for-octopus-server#question-will-i-be-affected

See https://octopusdeploy.slack.com/archives/CDANN5QLT/p1586905758124900


michaelnoonan:
Thanks @wordlee - I did check for that. It felt strange to omit certain details from one section, when it was helpful in both sections.

PR #729, opened 2020-05-06
Call out that that Export -> Modify -> Import isn't supported

jburger:
On occasion, customers are tempted to do this, but it lands them in hot water, sometimes bricking their instance entirely. I think it is a pretty typical thing for an engineer to attempt though, so I'm hoping that some explicit wording here should prevent some confusion for them and our engineers in support.

For example: [Internal tracking](https://secure.helpscout.net/conversation/1151986741/63948?folderId=557082)

### DIFF
--- a/docs/administration/data/data-migration.md
+++ b/docs/administration/data/data-migration.md
@@ -34,7 +34,7 @@ The data migration tools are not suitable for every imaginable scenario. In thes
 
 Sometimes using the data migration tool may look like it could solve a problem, but in fact will make things worse. Here are some scenarios we've seen that are explicitly not supported.
 
-### Export -> Modify -> Import
+1. **{{ Export , Modify , Import }}**
 
 Unfortunately, since the import isn't running all of the same validation checks as the API, using an `export` -> `modify` -> `import` can modify your data in such a way that is invalid for the API. Some scenarios _might work_ but because at this point you're effectively hand editing your data, it isn't something we support.
 :::
--- a/docs/administration/data/data-migration.md
+++ b/docs/administration/data/data-migration.md
@@ -31,7 +31,7 @@ The data migration tools are not suitable for every imaginable scenario. In thes
 
 ## Unsupported scenarios
 
-Sometimes using the data migration tool may look like it could solve a problem, but in fact will make things worse. Here are some scenario's we've seen that are explicitly not supported.
+Sometimes using the data migration tool may look like it could solve a problem, but in fact will make things worse. Here are some scenarios we've seen that are explicitly not supported.
 
 ## Export -> Modify -> Import
 
--- a/docs/administration/data/data-migration.md
+++ b/docs/administration/data/data-migration.md
@@ -33,7 +33,7 @@ The data migration tools are not suitable for every imaginable scenario. In thes
 
 Sometimes using the data migration tool may look like it could solve a problem, but in fact will make things worse. Here are some scenarios we've seen that are explicitly not supported.
 
-## Export -> Modify -> Import
+### Export -> Modify -> Import
 
 Unfortunately, since the import isn't running all of the same validation checks as the API, using an `export` -> `modify` -> `import` can modify your data in such a way that is invalid for the API. Some scenarios _might work_ but because at this point you're effectively hand editing your data, it isn't something we support.
 
--- a/docs/administration/data/data-migration.md
+++ b/docs/administration/data/data-migration.md
@@ -37,6 +37,7 @@ Sometimes using the data migration tool may look like it could solve a problem, 
 ### Export -> Modify -> Import
 
 Unfortunately, since the import isn't running all of the same validation checks as the API, using an `export` -> `modify` -> `import` can modify your data in such a way that is invalid for the API. Some scenarios _might work_ but because at this point you're effectively hand editing your data, it isn't something we support.
+:::
 
 ## Tips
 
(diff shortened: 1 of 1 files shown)


NickJosevski:
I'm ok with that, would like to see where @wordlee stands on general warnings to avoid unsupported actions in our docs.

docs/administration/data/data-migration.md
```diff
@@ -29,6 +29,14 @@ The data migration tools are not suitable for every imaginable scenario. In thes
 1. To move your entire Octopus Server from a self-hosted installation to Octopus Cloud, see [migrating from self-hosted to Octopus Cloud](/docs/octopus-cloud/migrations.md).
 1. To move your entire Octopus Cloud instance to a self-hosted Octopus Server please [contact our support team](mailto:support@octopus.com) for advice and assistance.
 
+## Unsupported scenarios
+
+Sometimes using the data migration tool may look like it could solve a problem, but in fact will make things worse. Here are some scenarios we've seen that are explicitly not supported.
+
+## Export -> Modify -> Import
```
jburger, line 36:
```suggestion
### Export -> Modify -> Import
```

docs/administration/data/data-migration.md
```diff
@@ -29,6 +29,14 @@ The data migration tools are not suitable for every imaginable scenario. In thes
 1. To move your entire Octopus Server from a self-hosted installation to Octopus Cloud, see [migrating from self-hosted to Octopus Cloud](/docs/octopus-cloud/migrations.md).
 1. To move your entire Octopus Cloud instance to a self-hosted Octopus Server please [contact our support team](mailto:support@octopus.com) for advice and assistance.
 
+## Unsupported scenarios
+
+Sometimes using the data migration tool may look like it could solve a problem, but in fact will make things worse. Here are some scenarios we've seen that are explicitly not supported.
+
+### Export -> Modify -> Import
```
paulegradie, line 36: An absolute nitpick on formatting that is totally up to you -> I might chuck at least one more hashtag on there to give a nice clear visual difference that we're reading inside a subsection. (.e.g. since right now we've got the one example, it kinda looks like `## tips` is part of the unsupported scenarios).

Nice explicit message! I can see how this would be tempting for users to do, so the warning is a good idea! Does hand manipulating data void warranty or anything severe like that? If so, might be good to mention. 🤷

wordlee:
> I'm ok with that, would like to see where @wordlee stands on general warnings to avoid unsupported actions in our docs.

In general, people assume "If it's in the docs, it's supported" and I think they're right to do so. If we can warn people **not to take** particular actions we should do so, AND we should be very clear so that people who scan the text don't see a series of steps and mistakenly think "that's what I was looking for" and then copy -> paste -> disaster :)

docs/administration/data/data-migration.md
wordlee: Note on the heading level, docsync will fail validation if the heading levels skip a level, i.e., going from ## (h2) to #### (h4), but I agree it would be good to make this stand out more, perhaps with :::warning which will put the section inside a red callout. I'll modify and commit.

docs/administration/data/data-migration.md
```diff
@@ -29,6 +29,14 @@ The data migration tools are not suitable for every imaginable scenario. In thes
 1. To move your entire Octopus Server from a self-hosted installation to Octopus Cloud, see [migrating from self-hosted to Octopus Cloud](/docs/octopus-cloud/migrations.md).
 1. To move your entire Octopus Cloud instance to a self-hosted Octopus Server please [contact our support team](mailto:support@octopus.com) for advice and assistance.
 
+## Unsupported scenarios
```
wordlee, line 32: Converting the section to a callout (problem not warning - problem is red, warning yellow) and removing the headings as they won't validate inside a callout:
```suggestion
:::problem
**Unsupported scenarios**
```

docs/administration/data/data-migration.md
```diff
@@ -29,6 +29,14 @@ The data migration tools are not suitable for every imaginable scenario. In thes
...
+
+Sometimes using the data migration tool may look like it could solve a problem, but in fact will make things worse. Here are some scenarios we've seen that are explicitly not supported.
+
+### Export -> Modify -> Import
+
+Unfortunately, since the import isn't running all of the same validation checks as the API, using an `export` -> `modify` -> `import` can modify your data in such a way that is invalid for the API. Some scenarios _might work_ but because at this point you're effectively hand editing your data, it isn't something we support.
```
wordlee, line 38: closing the callout
```suggestion
Unfortunately, since the import isn't running all of the same validation checks as the API, using an `export` -> `modify` -> `import` can modify your data in such a way that is invalid for the API. Some scenarios _might work_ but because at this point you're effectively hand editing your data, it isn't something we support.
:::
```

docs/administration/data/data-migration.md
wordlee, line 36: removed the heading from the callout and converted to markdown to achieve a nicer look :)
```suggestion
1. **{{ Export , Modify , Import }}**
```

docs/administration/data/data-migration.md
```diff
@@ -29,6 +29,16 @@ The data migration tools are not suitable for every imaginable scenario. In thes
...
 1. To move your entire Octopus Cloud instance to a self-hosted Octopus Server please [contact our support team](mailto:support@octopus.com) for advice and assistance.
 
+:::problem
+**Unsupported scenarios**
+
+Sometimes using the data migration tool may look like it could solve a problem, but in fact will make things worse. Here are some scenarios we've seen that are explicitly not supported.
+
+1. **{{ Export , Modify , Import }}**
+
+Unfortunately, since the import isn't running all of the same validation checks as the API, using an `export` -> `modify` -> `import` can modify your data in such a way that is invalid for the API. Some scenarios _might work_ but because at this point you're effectively hand editing your data, it isn't something we support.
```
wordlee, line 39: slight tweak:
```suggestion
Unfortunately, since the import isn't running all of the same validation checks as the API, using an **{{ export, modify, import}}** can modify your data in such a way that is invalid for the API. Some scenarios _might work_ but because at this point you're effectively hand editing your data, this isn't something we support.
```

jburger:
Thanks all :+1:

PR #736, opened 2020-05-12
Making security content a top level concern in the docs

wordlee:
Security content in the docs is spread across a few different locations. To make it easier for customers to access, it will move to docs/security.

Content that was moved from docs/administration to docs/security:
- docs/administration/security/*
- docs/administration/authentication/*
- docs/administration/users-and-teams/*

- [x] Update OctoPurl with the new links


wordlee:
FYI @NickJosevski I'm merging this into master, but wanted to give you a heads up, just in case it overlaps with your work.

PR #745, opened 2020-05-21
Add list of tools for container-worker images

harrisonmeister:


matt-richardson:
I'm concerned this will get out of date pretty quickly.
I wonder if we should just point them to the dockerhub repo. I'm pretty sure the team was looking at automatically updating the readme on there during build. (I think @emilol might have more info about whether that actually happened)

harrisonmeister:
> I'm concerned this will get out of date pretty quickly.
> I wonder if we should just point them to the dockerhub repo. I'm pretty sure the team was looking at automatically updating the readme on there during build. (I think @emilol might have more info about whether that actually happened)

That's a fair point. I'd be happy to point users to the dockerhub repo if it provides additional context. My use-case was originally trying to find out if the gcloud cli was in the image for a sample project I'm working on

robpearson:
I think it's a balance. I like the idea of including the list here as it's far more helpful for people who read the page. I think it's probably a good idea to add a blurb saying this toolset was added at the time of writing but the repository is the best resource for the up to date list. /cc @wordlee

wordlee:
Maintenance is always an issue when for items like this that go out of date quickly. We usually link to the authoritative source, for instance, `for a complete list of the pre-installed tools see the [Docker Hub repo](link)`. If it's possible to automatically update the list of pre-installed tools, that would be great to include, but only if it is automatic, otherwise it will date and customers will be confused by the differences :)

emilol:
We update the docker image in the docs automatically, is that maybe what you're thinking of @matt-richardson? I'm not aware of any plans to update tooling in readmes automatically, but that sounds pretty neat!

harrisonmeister:
I think these comments resonate where my main concern is now. I'd rather not link to the dockerhub repo if it doesnt _actually_ contain a list of tools (version or not). I appreciate that automatically updating it is additional work (co-incidentally I do something similar to the [octopus-cli-pipe README](https://bitbucket.org/octopusdeploy/octopus-cli-run/src/master/ci-scripts/bump-version.sh#lines-17) as part of it's CI process - though that's just for the version of the docker image itself.

Where do we currently maintain the versions of tooling in the docker images - is it the `ARG` parameters [here](https://github.com/OctopusDeploy/WorkerTools/blob/master/ubuntu.18.04/Dockerfile#L3) ?

Perhaps this PR should change to:
- link to DockerHub
- The DockerHub readme should link to the Dockerfile's for each supported version

That is more than one click to get to the information, so we could just link direct to the Dockerfiles?

harrisonmeister:
Closing this PR as the docs have since been updated

PR #748, opened 2020-05-27
Vm extension wording

mcasperson:
Noted that the VM extension is depreciated and will be unsupported from 2021.


docs/infrastructure/deployment-targets/windows-targets/azure-virtual-machines/configuration-structure.md
```diff
@@ -4,6 +4,10 @@ description: Detailed file format information for the configuration files requir
 position: 7
 ---
 
+:::warning
+The VM extension is depreciated and will be unsupported from 2021.
```
MJRichardson, line 8: `Deprecated` rather than `depreciated`

Same thing below in a few places

PR #755, opened 2020-06-01
Add more details about Service Principal roles

ryanrousseau:


ryanrousseau:
Thanks, @wordlee. I added another quick change to let people know there might be a time delay between adding the service principal and it being available to Octopus.

PR #756, opened 2020-06-02
Reporting

DevOpsDerek:
Split up the original page and adding some details, and a new image.


docs/administration/reporting/report-on-deployments-excel-template.md
```diff
@@ -0,0 +1,89 @@
+---
+title: Reporting
+description: Octopus reporting with excel template
```
wordlee, line 3:
```suggestion
description: Octopus reporting with an excel template
```

PR #765, opened 2020-06-08
Update pictures and add warning about Subversion

millerjn21:
The pictures needed to be updated as one of them contains "Metadata" which isnt what our current version shows. A user got concerned that they had setup build info correctly since theirs didnt say Metadata. I've taken 3 new screenshots for consistency. I also added a warning that subversion doesnt work for commit messages, only github.


wordlee:
Hi @millerjn21, updates look good. You still need to update the images. When you do, make sure the image filenames are all in lowercase or docsync will fail.

PR #777, opened 2020-06-12
Preventing progression

wordlee:
I made some further changes to the doc so that the language is based around preventing progression rather than blocking. I also update the page title, file name, and added redirects.


adamoctoclose:
LGTM

PR #782, opened 2020-06-16
Change recommendations for Octopus.Client usage in steps

TomPeters:
I think we want to help customers shy away from using the version of Octopus.Client we ship with Octopus Server or Tentacle. Instead, we should recommend they obtain Octopus.Client using the Packages feature.

### DIFF
--- a/docs/octopus-rest-api/octopus.client.md
+++ b/docs/octopus-rest-api/octopus.client.md
@@ -222,21 +222,18 @@ var projects = repositoryForSpace.Projects.GetAll();
 
 ## Loading in an Octopus Step {#Octopus.Client-Loadinginanoctopusstep}
 
-You can use Octopus.Client from inside Octopus (for example in a script step, a package install script, or the script console) by loading it from the server or Tentacle application directory. The credentials would still need to be supplied to establish the connection. For example:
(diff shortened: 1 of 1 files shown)


docs/octopus-rest-api/octopus.client.md
```diff
@@ -222,20 +222,21 @@ var projects = repositoryForSpace.Projects.GetAll();
 
 ## Loading in an Octopus Step {#Octopus.Client-Loadinginanoctopusstep}
 
-You can use Octopus.Client from inside Octopus (for example in a script step, a package install script, or the script console) by loading it from the server or Tentacle application directory. The credentials would still need to be supplied to establish the connection. For example:
+You can use Octopus.Client from inside Octopus (for example in a script step or a package install script) by referencing it as an package. You can configure [nuget.org](https://api.nuget.org/v3/index.json) as the [External Feed](https://octopus.com/docs/packaging-applications/package-repositories/nuget-feeds) that provides this package. Octopus will automatically extract this package for you, allowing your script to reference the .dll file it contains using a relative path. For example:
```
matt-richardson, line 225:
```suggestion
You can use Octopus.Client from inside Octopus (for example in a script step or a package install script) by referencing it as an package. You can configure [nuget.org](https://api.nuget.org/v3/index.json) as an [External Feed](https://octopus.com/docs/packaging-applications/package-repositories/nuget-feeds) that provides this package. Octopus will automatically extract this package for you, allowing your script to reference the .dll file it contains using a relative path. For example:
```

PR #792, opened 2020-06-24
Add telemetry and clusterShared path commands

tothegills:
Added two new path commands.

`telemetry` sets the storage location of temporary files used when collecting telemetry. This should be path shared by all nodes in a cluster.
`clusterShared` sets the root path shared by nodes in a cluster. This is a shorthand way to configure a cluster instead of setting each path individually, previously each of artifacts, built-in repository and task logs would need to be configured.


wordlee:
Hi @tothegills, these edits look good and we can merge them. Note for future though, we have a tool that runs about once a week and updates these docs automatically with any new commands in the CLI. Next times it runs, it should these edits have already been added and will skip these (unless there are any differences).

PR #803, opened 2020-07-01
Quickstart guides

wordlee:
A series of guides that extend the Getting Started page and provide step by step instructions to familiarize new users with the fundamentals.


wordlee:
@OctopusBob, this PR is only an initial pass at adding some concrete getting started procedural content. Some of the content already exists on the index pages of the other sections, but collecting it in one place and adding additional content should make it more discoverable, and allow us to add similar content. 

Re our Slack conversation, this PR isn't intended to be the final form, but a starting place to build from - where that's in the formal docs or somewhere else.

PR #815, opened 2020-07-10
Updating AWS section for file storage

twerthi:
Updated AWS section in regards to file storage options.


docs/administration/high-availability/configuring-octopus-for-high-availability.md
```diff
@@ -67,7 +67,7 @@ Once you have [created your File Share](https://docs.microsoft.com/en-us/azure/s
...
+If your Octopus Server is running in Amazon AWS you can use [AWS FSx](https://aws.amazon.com/fsx/).  FSx can be referenced by your HA nodes via UNC path or can be configured to use [Microsoft DFS](https://en.wikipedia.org/wiki/Distributed_File_System_(Microsoft)).  Use of FSx requires an Active Directory (AD) instance running in AWS.  If you don't have an AD instance, or don't want one, you will need to configure your own file share using something like [Microsoft DFS](https://en.wikipedia.org/wiki/Distributed_File_System_(Microsoft)). Unfortunately [Amazon Elastic File System (EFS)](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AmazonEFS.html) is not supported on Windows, and [Amazon Elastic Block Store (EBS) volumes](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumes.html) can only be mounted on a single instance.
```
wordlee, line 70:
```suggestion
If your Octopus Server is running in Amazon AWS you can use [AWS FSx](https://aws.amazon.com/fsx/).  FSx can be referenced by your HA nodes via UNC path or it can be configured to use [Microsoft DFS](https://en.wikipedia.org/wiki/Distributed_File_System_(Microsoft)).  Use of FSx requires an Active Directory (AD) instance running in AWS.  If you don't have an AD instance, or don't want one, you will need to configure your own file share using something like [Microsoft DFS](https://en.wikipedia.org/wiki/Distributed_File_System_(Microsoft)). Unfortunately [Amazon Elastic File System (EFS)](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AmazonEFS.html) is not supported on Windows, and [Amazon Elastic Block Store (EBS) volumes](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumes.html) can only be mounted on a single instance.
```

PR #839, opened 2020-07-23
Harden iis

twerthi:


wordlee:
This PR has been merged, but I'm adding @OctopusDeploy/school-security retroactively to review the content from a security perspective.

jburger:
@twerthi this is great work - I like the example in that it feels relevant and useful. I also like that you've shown the sources of your script too, that engenders trust with the reader. 

My only question here would be that things like ciphersuite recommendations from NIST and other compliance frameworks are going to change over time. Is it your intent that we'd stay across that for our customers and keep this page up to date?

If not, then some alternatives to that approach might be to make it clear that this is an example and may become out of date with current recommendations. Or we could comment out parts of the example that are likely to change (like the cipher suites) which would force the customer think a bit harder about those.

I'll leave that decision with you, and I'll support whatever direction you want to take that.

docs/runbooks/runbook-examples/routine/hardening-iis.md
```diff
@@ -0,0 +1,236 @@
...
+	  'TLS_DHE_DSS_WITH_AES_256_CBC_SHA256',
+	  'TLS_DHE_DSS_WITH_AES_256_CBC_SHA',
+	  'TLS_DHE_DSS_WITH_AES_128_CBC_SHA256',
+	  'TLS_DHE_DSS_WITH_AES_128_CBC_SHA',
+	  'TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA',
+	  'TLS_RSA_WITH_AES_256_CBC_SHA256',
+	  'TLS_RSA_WITH_AES_256_CBC_SHA',
+	  'TLS_RSA_WITH_AES_128_CBC_SHA256',
+	  'TLS_RSA_WITH_AES_128_CBC_SHA',
+	  'TLS_RSA_WITH_3DES_EDE_CBC_SHA'
```
jburger, line 215: This doesn't look right to me @twerthi 

`TLS_RSA_WITH_3DES_EDE_CBC_SHA`
`TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA`

AFAIK Triple Data Encryption Standard is not a perfect forward secrecy cipher suite and is no longer recommended by NIST

https://csrc.nist.gov/News/2017/Update-to-Current-Use-and-Deprecation-of-TDEA

The source of the script that you may have used appears to be a little bit different now https://www.hass.de/content/setup-microsoft-windows-or-iis-ssl-perfect-forward-secrecy-and-tls-12 ?

twerthi: Ah!  @wordlee , good call having @jburger review this!  @jburger Thanks for catching that, I've removed the triple des stuff and submitted another PR.  I did find this buried in the comments of the script, `# - 3DES: It is recommended to disable these in near future` So it looks like it is just out of date.  

In regards to your question about whether regularly updating it, I didn't really have any plans.  This was to be provided as an example of what is possible.  I'm not against keeping it up to date, but not sure where the responsibility would lie.  I'm certainly up for discussion.

wordlee: > In regards to your question about whether regularly updating it, I didn't really have any plans. This was to be provided as an example of what is possible. I'm not against keeping it up to date, but not sure where the responsibility would lie. I'm certainly up for discussion.

We don't currently have any formal processes around updating content. My preference is to include the security disclaimer from #857, and make sure users understand, the guidance is for demonstration purposes only, and they should check ... still think of the words :)

PR #840, opened 2020-07-23
Harden windows

twerthi:


wordlee:
Looks good, Shawn. I made a couple of edits, but I think you should check these ones :)

wordlee:
This PR has been merged, but I'm adding @OctopusDeploy/school-security retroactively to review the content from a security perspective.

PR #860, opened 2020-07-30
Suggested minor improvements to Structured Configuration Variables docs

thedewi:
I have a few minor suggestions I hope can simplify wording and improve flow, so I'm raising this as a PR.

There is only one real correction here, which was that the final YAML example didn't contain the `baz`. However, I am suggesting here to adjust the example using the familiar structure from a previous example, because I thought having a property key called `key` might be confusing.

I also would remove "the variables section of your deployment process" because variables don't live under Deployment Process at this point.

I've folded the final sentence about variable substitution syntax into the previous paragraph about that, so that we discuss the basic usage and then the extended syntax.

### DIFF
--- a/docs/shared-content/structured-configuration-variables.include.md
+++ b/docs/shared-content/structured-configuration-variables.include.md
@@ -1,10 +1,10 @@
 :::info
-This Configuration Feature used to be called JSON Configuration Variarbles but has been re-named with the added support for YAML
+This Configuration Feature used to be called JSON Configuration Variarbles but has been re-named with the added support for YAML.
 :::
 
-With the Structured Configuration Variables feature you can define [variables](/docs/projects/variables/index.md) in Octopus for use in the JSON and YAML configuration files of your applications. This lets you define different values based on the scope of the deployment. This feature uses a matching syntax so you can update configuration nested in JSON and YAML objects and array literals.
+With the **Structured Configuration Variables** feature you can define [variables](/docs/projects/variables/index.md) in Octopus for use in the JSON and YAML configuration files of your applications. This lets you define different values based on the scope of the deployment. This feature uses a matching syntax so you can update configuration nested in JSON and YAML objects and array literals.
 
-This is designed to work natively with [.NET Core Structured configuration files](/docs/deployment-examples/asp.net-core-web-application-deployments/index.md) but works equally well with any JSON files.
+This is designed to work natively with [.NET Core Structured configuration files](/docs/deployment-examples/asp.net-core-web-application-deployments/index.md), but works equally well with any JSON or YAML files.
 
 ## Configuring the structured configuration variables feature {#StructuredConfigurationVariablesFeature-Configuringthestructuredconfigurationvariablesfeature}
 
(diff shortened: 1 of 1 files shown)


docs/shared-content/structured-configuration-variables.include.md
```diff
@@ -71,7 +71,7 @@ Given this example of a target config file:
 }
 ```
 
-If you define the [variables](/docs/projects/variables/index.md) `weatherApiUrl`, `weatherApiKey`, `port`, and `debug` in the variables section of your [deployment process](/docs/projects/variables/index.md) with the values `test.weather.com`, `TEST7654321`, `80`, and `false`, the target config file in your packaged application is updated to become:
+If you define [variables](/docs/projects/variables/index.md) in your Octopus Project called `weatherApiUrl`, `weatherApiKey`, `port`, and `debug` with the values `test.weather.com`, `TEST7654321`, `80`, and `false`, the target config file is updated to become:
```
wordlee, line 74:
```suggestion
If you define [variables](/docs/projects/variables/index.md) in your Octopus project called `weatherApiUrl`, `weatherApiKey`, `port`, and `debug` with the values `test.weather.com`, `TEST7654321`, `80`, and `false`, the target config file is updated to become:
```

docs/shared-content/structured-configuration-variables.include.md
```diff
@@ -108,7 +108,7 @@ weatherApi:
   key: DEV1234567
 ```
 
-You can also replace an entire object. For the example above you could set Octopus Variable `weatherApi` to a value of `{"url":"test.weather.com","key":"TEST7654321"}` which will result in this:
+You can also replace an entire object. For the example above, you could set Octopus Variable `weatherApi` to a value of `{"url":"test.weather.com","key":"TEST7654321"}`, which will result in this:
```
wordlee, line 111:
```suggestion
You can also replace an entire object. For the example above, you could set Octopus variable `weatherApi` to a value of `{"url":"test.weather.com","key":"TEST7654321"}`, which will result in this:
```

docs/shared-content/structured-configuration-variables.include.md
```diff
@@ -173,7 +173,7 @@ foo:
     - qux
 ```
 
-It's possible to replace an entire array or sequence too. With the previous example above, if the Octopus Variable `foo:bar` was to a value of `["baz","qux"]` it would create outputs like:
+It's possible to replace an entire array or sequence too. In the previous example above, if the Octopus Variable `foo:bar` was set to `["baz","qux"]`, it would create outputs like:
```
wordlee, line 176:
```suggestion
It's possible to replace an entire array or sequence too. In the previous example, if the Octopus Variable `foo:bar` was set to `["baz","qux"]`, it would create outputs like:
```

docs/shared-content/structured-configuration-variables.include.md
```diff
@@ -21,41 +21,41 @@ or
...
 ```
 
-**Match any .yaml files in specified folder or deeper**
+**Match any .yaml files in the specified directory or deeper**
 ```
 Application/**/*.yaml
 ```
 
-The Target File field also supports [extended template syntax](/docs/projects/variables/variable-substitutions.md#VariableSubstitutionSyntax-ExtendedSyntax) which can allow conditionals and loops to be used to specify files.
+The Target File field also supports [Variable Substitution Syntax](/docs/projects/variables/variable-substitutions.md), to allow things like referencing environment-specific files, or conditionally including them based on scoped variables. [Extended template syntax](/docs/projects/variables/variable-substitutions.md#VariableSubstitutionSyntax-ExtendedSyntax) allows conditionals and loops to be used.
```
wordlee, line 52:
```suggestion
The **Target File** field also supports [Variable Substitution Syntax](/docs/projects/variables/variable-substitutions.md), to allow things like referencing environment-specific files, or conditionally including them based on scoped variables. [Extended template syntax](/docs/projects/variables/variable-substitutions.md#VariableSubstitutionSyntax-ExtendedSyntax) allows conditionals and loops to be used.
```

thedewi:
> Looks good. I committed a couple of minor edits.

Thanks @wordlee. I like all these. 👍

PR #866, opened 2020-07-31
Harden linux

twerthi:


wordlee:
I've added school of security to look at this from a security perspective :)

PR #867, opened 2020-08-03
Adding basic sample

twerthi:


docs/octopus-rest-api/octopus-cli/allow-releaseprogression.md
```diff
@@ -73,6 +73,12 @@ Common options:
                              fatal. Defaults to 'debug'.
 ```
 
+## Basic examples
```
wordlee, line 76:
```suggestion
## Basic example
```

harrisonmeister: @wordlee on this, would you pluralize only when there is more than one (just for future reference)

PR #868, opened 2020-08-04
Add new filters.

andrew-moore-octo:
(draft until the changes have been merged to Octostache, and integrated into Calamari)

This PR adds information about four new filters supported in Calamari:

- `YamlSingleQuoteEscape`
- `YamlDoubleQuoteEscape`
- `PropertiesKeyEscape`
- `PropertiesValueEscape`

### DIFF
--- a/docs/projects/variables/variable-filters.md
+++ b/docs/projects/variables/variable-filters.md
@@ -37,7 +37,7 @@ The filters provided by Octopus are for use with trusted input; don't rely on th
 Octopus provides the following filters:
 
 | Name                                    | Purpose                                           | Example input           | Example output                   |
-| --------------------------------------- | ------------------------------------------------- | ----------------------- | -------------------------------- |
+| --------------------------------------- |-------------------------------------------------- | ----------------------- | -------------------------------- |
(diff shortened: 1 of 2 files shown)


andrew-moore-octo:
[Update docs](https://trello.com/c/a9Jo3p1w/56-update-docs)

andrew-moore-octo:
@wordlee I've been thinking whether it's worth adding some additional info to these new filters so users know which version of Octopus Server will include them. Screenshots taken using VS Codes markdown preview, which is why they look odd, but hopefully it's enough as a mockup:

**Option 0: how it is now - don't show the version**
image

**Option 1: Add a new column**
image

**Option 2: Version in the Name column**
image

**Option 3: Version in the Purpose column**
image

What do you think? Do you think this is a good path to go down, and if so, do any of the options above work better (or something else entirely)?

wordlee:
Hi @andrew-moore-octo, I think it is worth it now that we have multiple supported versions and the change is happening between those versions. We're somewhat limited by the format (I try to avoid too many columns because the user's screen width might squash things up a bit too much), so option 2 or 3 makes sense to me, and of the two I think the version stands out the most in option 2.

andrew-moore-octo:
Thanks @wordlee! I've updated the table to include the version numbers like in option 2.

PR #873, opened 2020-08-04
Move autocomplete from general settings to shared and reference include

harrisonmeister:
Moved autocomplete to a shared include and referenced in both the octopus cli index page and the `install-autocomplete` cli page


harrisonmeister:
@wordlee - do you think this makes sense to add to an include? Discussed a bit and it felt like the best approach, but happy for you to make any changes :+1:

wordlee:
> @wordlee - do you think this makes sense to add to an include? Discussed a bit and it felt like the best approach, but happy for you to make any changes 👍

Yep. I think this is the perfect use case.

PR #897, opened 2020-08-06
Prompted variables in runbooks

MJRichardson:
It is common when defining prompted variables to want them to apply only to the deployment process, or only to specific runbooks.

This only becomes important once you have both prompted variables and runbooks in a project.


MJRichardson:
Lovely!  Thanks @wordlee!

PR #993, opened 2020-09-03
Forward slash breaking in-page links

danefalvo:
The '/' symbol within Document chapters titles are breaking the inpage Targets


wordlee (approved):
This should do it :)

PR #1008, opened 2020-09-09
Add more REST api examples

harrisonmeister:
Removed a bunch of positions for child pages which were added before folder structure was added

Also added examples:

- Create API Key
- Deploy existing release
- Add target role
- Delete targets by name


twerthi (approved):
LGTM

wordlee (commented):
👍

PR #1051, opened 2020-09-17
Add publish-runbook python example

harrisonmeister:


ryanrousseau (changes requested):
I think I've found an easier way to publish. I'll message it to you

harrisonmeister:
@ryanrousseau - updated all scripts to follow the same pattern of getting the snapshot by name and setting the `PublishedSnapshotId`

PR #1083, opened 2020-10-02
Fix a couple typos.

pstephenson02:


wordlee (approved):
Thanks for the edits @pstephenson02!

PR #1088, opened 2020-10-13
Removed Octopus Server section

wordlee:
No longer references the octopus-server.include file that added the octopus server content.

I removed the shared octopus-cloud content (it was only used in this one place), simplified and rewrote the introduction.


harrisonmeister (approved):
LGTM

PR #1109, opened 2020-10-27
Fix Spelling mistake in screenshot

danefalvo:
was: evalutedVariables
now: evaluatedVariables


wordlee (dismissed):
Looks good, @danefalvo! However, the build is failing validation because of the capital letter in the new file name (descriptive filenames are much better than numbers, however docsync requires all lowercase titles). The only way I've been able to rename file names in GitHub is to remove the file from the repository locally, commit the deletion, rename the file and add it back into the repo. There's probably an easier but I don't what it is :)

harrisonmeister:
@wordlee - I made the changes to this as was in the repo :)

> Looks good, @danefalvo! However, the build is failing validation because of the capital letter in the new file name (descriptive filenames are much better than numbers, however docsync requires all lowercase titles). The only way I've been able to rename file names in GitHub is to remove the file from the repository locally, commit the deletion, rename the file and add it back into the repo. There's probably an easier but I don't what it is :)

harrisonmeister (approved):
:+1:

PR #1115, opened 2020-11-03
Update path to Calamari to point to Tools

harrisonmeister:
This changed somewhere between 2020.1 and 2020.4 where Calamari used to be installed to `<Tentacle Home>\Calamari`. It's now installed to `<Tentacle Home>\Tools`. Updating docs to reflect this change.


harrisonmeister:
@ryanrousseau / @wordlee - thoughts on also adding a sentence below the current block to mention specifically that `<Tentacle Home>\Calamari` can be safely cleaned up - once I've confirmed which version it changed in?

harrisonmeister:
> @ryanrousseau / @wordlee - thoughts on also adding a sentence below the current block to mention specifically that `<Tentacle Home>\Calamari` can be safely cleaned up - once I've confirmed which version it changed in?

Actually thinking about this, I'm not sure it's worth adding a note about the old folder. I'm leaning the other way and considering if we should remove the 2.6 mentions of the folders?

PR #1125, opened 2020-11-17
Optimum setup projects

octocrock:
Porting this: https://github.com/OctopusDeploy/OptimumSetupBook/blob/master/manuscript/Projects.md to the docs.  Took the first section and added that as project-recommendations.md.

Second section, where the project is configured has gone into octopus-recommendations/configure-project - lots of images and the first part of what, we think, will become the "book".


docs/shared-content/octopus-recommendations/project-recommendations.include.md
```diff
@@ -0,0 +1,58 @@
...
+
+Knowing the underlying concept of Octopus Deploy is consistency; here are our project recommendations.
+
+## Keep projects simple
+
+One of our favorite programming maxims is the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle). It says:
+
+> A class/function/method should do one thing, and it should do it well.  
+
+The same is true for projects.  Projects should be straightforward and should deploy a component of an application.
```
harrisonmeister, line 12: I wonder if this section should be qualified with some pros/cons of a single component project (e.g. microservices architectural style) vs a more combined project approach, even for octofx? 

Splitting it by component makes it easier to deploy the individual components, but makes it more annoying when deploying the whole thing - even with the deploy a release step.

It could even just a call-out to say:

> Whilst creating projects per component is a good practice, it's not the rule. Octopus provides flexibility to allow you to model your projects to meet your requirements.

harrisonmeister (commented):
Looks good, made some minor edits and commented on one of the recommendations.

docs/shared-content/octopus-recommendations/project-recommendations.include.md
octocrock: Good point, tweeked

harrisonmeister (approved):
LGTM

PR #1146, opened 2020-12-14
Linux ARM 32 installer scripts

benPearce1:
I think this would be better served as an additional parameter in the scripts where the user has to set the correct architecture:

```
applicationPath="/home/Octopus/Applications/"
arch="x64" # also supports arm64 or arm

curl -L https://octopus.com/downloads/latest/Linux_x64TarGz/OctopusTentacle --output tentacle-linux_x64.tar.gz
...
```

Then the whole archive section would collapse to a single section. 

Thoughts?


harrisonmeister:
>Then the whole archive section would collapse to a single section.

I think this makes sense. You lose a section so there's less to scroll/read so :+1: from me.

benPearce1:
Refactored script for all platforms.

This PR is waiting on the changes in Octofront to ship to make the download links work

harrisonmeister (approved):
:+1:

PR #1154, opened 2020-12-23
Include tools and templates to auto-update credits.md

uglybugger:
# Background

We recently changed the way we generate licence/credits files and internally test for OSS licence compliance. This has necessitated a change to our `docs` repository with regard to how we generate our `credits.md` file.

# Results

This PR includes
- a .NET tool manifest
- a `dotnet` tool to generate the `credits.md` file
- a markdown template file for the credits file
- a `NuGet.config` file to allow downloading the .NET tool from our unauthenticated/public NuGet feed
- a `.gitignore` entry to ignore temporary directories created during generation of `credits.md` in case the cleanup fails.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "octopus.creditsfilegenerator.tool": {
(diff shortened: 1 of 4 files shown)


harrisonmeister (approved):
LGTM

PR #1162, opened 2021-01-08
Matched "Truthy and Falsy Values" to code

geeeyetee:
Matched the explanation of how conditions are evaluated to the algorithm in what I believe is the operative Octostache method:
https://github.com/OctopusDeploy/Octostache/blob/master/source/Octostache/Templates/TemplateEvaluator.cs#L186-L192
```
        internal static bool IsTruthy(string value)
        {
            return value != "0" &&
                value != "" &&
                string.Compare(value.Trim(), "no", StringComparison.OrdinalIgnoreCase) != 0 &&
                string.Compare(value.Trim(), "false", StringComparison.OrdinalIgnoreCase) != 0;
        }
```


harrisonmeister (approved):
Looks good, thanks for the contribution!

PR #1163, opened 2021-01-08
Updating the documentation to show the worker variable

millerjn21:
https://github.com/OctopusDeploy/Octopus-Docker/blob/master/Tentacle/readme.md

The variable was missing that would allow the containers to be workers. Added it to the list.


Justin-Walsh:
Is it worth going into more detail here, this seems pretty easily missed - maybe a section on setting it up as a worker with an example docker command stack?

harrisonmeister (approved):
LGTM

PR #1177, opened 2021-01-31
Bug fix: remove credits template file from index

uglybugger:
This PR fixes a bug introduced in #1154, which used a file named `credits.template.md` as a template for a generated `credits.md` file.

The indexing picked up the `credits.template.md` file as a navigable page, which was not intended.

This PR renames the file to `credits.md.template` so that it will not be picked up by the indexing code.


harrisonmeister (approved):
:+1: LGTM

PR #1181, opened 2021-02-04
Automatically update credits file with latest licence information

Octobob:
This is an automated update of the credits / license attribution file, based on the latest code in https://github.com/OctopusDeploy/OctopusDeploy.

Please review for

- syntactical correctness (ie, markdown looks correct)
- logical correctness (ie, when reading it, it makes sense)
- license conformance (we agree with the license in question and we are able to redistribute it)

This pull request was created by TeamCity build 1963755


uglybugger:
> This pull request was created by TeamCity build 1963755

I've updated the script to include the build name as well as the ID for next time.

michaelnoonan (approved):
This changes all make sense.

PR #1186, opened 2021-02-15
add anchor for the disable default space

cailyoung:
I was unable to find the section in this page for disabling the default space. Have added an anchor (I think!) to the warning paragraph above the relevant section.

### DIFF
--- a/docs/administration/spaces/index.md
+++ b/docs/administration/spaces/index.md
@@ -92,6 +92,7 @@ The following carries some minor downtime for any automation that relies on the 
 
 For organizations that are new to Octopus, especially those that make heavy use of spaces, a default space is not required, and you can remove the default space entirely. However, this comes with some considerations that should be weighed carefully against the needs of your organization.
 
+{#delete-the-default-space}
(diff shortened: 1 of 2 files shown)


harrisonmeister:
Thanks for the change @cailyoung 

I updated the page a bit to hopefully make each section clearer:
- Added disable default space steps to an include and referenced in the change default space steps at the start
- Replaced the disable default space steps with include
- Changed anchor link in the paragraph at the top talking about disabling the default space

Just as an FYI the anchor you added wouldn't have worked:
- #delete-the-default-space was an anchor that wouldn't have resolved to an existing Heading section (e.g. H1, H2 etc)
- Adding a new anchor would need to be prefixed with a markdown header e.g. `### Delete the default space {#delete-the-default-space}`

Let me know what you think of the change.

cailyoung:
> Let me know what you think of the change.

Thanks a lot for cleaning this up! I think it's a lot clearer overall. I'm not familiar enough with the way the docs get presented to have thought to do something like the `!include` you've done there, but that's much better.

PR #1188, opened 2021-02-15
Consolidate guides/pages on Azure Cloud Services deployments

pstephenson02:
There are 3 different pages here under Deployments -> Azure that all contain very similar instructions and help for deploying to Azure Cloud Services. With this technology being deprecated, I've consolidated these guides and re-organized the links a bit such that there's only 1 guide but includes all the relevant info from the other 3.

image

I'm not sure how strict we've traditionally been with cleaning up orphaned screen shots, but I can certainly do that if needed.


harrisonmeister:
Took a look at the changes, and they look :+1: 

I made a few adjustments:

1. The Learn more + in this section can be condensed into one :smile: 
image
1. Removed a double `:::hint` on the main cloud services page
1. Extracted an include for the warning to allow for displaying on the VIP swap page
1. Amended the wording of the warning to reinforce Azure Cloud Services are considered legacy now.

Thanks!

harrisonmeister (approved):
Nice work!

PR #1200, opened 2021-02-18
Move the Certificate supported formats page content to the Certificates index page.

pstephenson02:
This support formats page is very short and the info on it can just as well belong on the Certificates index page. This gives us one less link in the sidebar 🐙


harrisonmeister (approved):
LGTM

PR #1205, opened 2021-02-24
Remove reference to Windows 2003

pstephenson02:
This is a small thing. I hope no one is still using Windows Server 2003, but since it's not officially supported by Tentacle, I think we shouldn't suggest anything here.


harrisonmeister (approved):
Agreed, we shouldn't really be highlighting an OS which is out of support anymore

PR #1215, opened 2021-02-26
Pl/unused targets handle pre spaces versions

peterlanoie:
Updated the unused targets check script to handle pre-Spaces versions.


OctopusBob (approved):
LGTM

PR #1222, opened 2021-03-11
Changing syntax for iterating through Items.

millerjn21:
We got a ticket saying the syntax was wrong for the loop for variable sets and I tested it and he was right.


harrisonmeister (approved):
LGTM

PR #1230, opened 2021-03-19
Update tentacle containers port

Justin-Walsh:
Update the tentacle port in the example to point to the default port of the server docker container, instead of 8065.


harrisonmeister (approved):
👍

PR #1231, opened 2021-03-22
Update github-actions.md

jbristowe:
* Replaced "install-octocli" with "install-octopus-cli-action" (new name)
* Replaced "Octopus Deploy CLI" with "Octopus CLI" for consistency
* Replaced version "7.4.2" with "latest" (this is a new feature)


harrisonmeister (approved):
LGTM :+1: - FYI I think we can switch out the version to an include file. Will test it out

PR #1257, opened 2021-05-04
Performance support

OctopusBob:
Made the performane troubleshooting section a bit easier to navigate.  Cleaned up / combined the amount of information we ask for when reaching out to support.


harrisonmeister (approved):
LGTM

PR #1270, opened 2021-05-17
Fixes Octo CLI help command example in packaging applications

geofflamrock:
This is a minor fix to the help command example that is given in the "Create packages with the Octopus CLI" page, the current version doesn't run when pasted into powershell as it includes the path "C:\>"


harrisonmeister (approved):
LGTM

PR #1272, opened 2021-05-19
Switch the titles between unique port/address sections on the Polling tentacles with HA page

pstephenson02:


harrisonmeister (approved):
LGTM. Good spot!

PR #1275, opened 2021-05-20
Changes running azure powershell to indicate bundled tools are no longer recommended

geofflamrock:
Changes the documentation on Running Azure Powershell scripts to indicate that using the bundled tools is not longer recommended, instead users should use their own version of the Azure Powershell Modules and Azure CLI


geofflamrock:
[Using Bundled Tools with Azure Steps does not work when using PowerShell Core](https://trello.com/c/DTMxmwiE/29-using-bundled-tools-with-azure-steps-does-not-work-when-using-powershell-core)

PR #1280, opened 2021-05-21
Updates AWS tool guidance for scripts

geofflamrock:
This PR makes changes to the AWS Script Step documentation, mostly to make it explicit that we do not recommend people use the inbuilt AWS tools, and that we will no longer be updating them.


iryan2 (approved):
LGTM

harrisonmeister (approved):
:+1:

PR #1290, opened 2021-06-02
Making some tweaks to the upgrade guide.

BobJWalker:
This PR makes the following changes:
- Breaking out the backup master key and license key to a "prep work" section.  
- Rewriting the automate upgrade page.  Two sections, upgrading a single node instance and a HA instance.  All the scripts have been verified and work out of the box.
- Added a SQL Script a user can run to disable triggers, targets, and workers after creating a cloned instance
- Added in a note about using the new Export/Import project feature to create the test instance


harrisonmeister (approved):
LGTM

PR #1291, opened 2021-06-04
Direct users to Bento

danefalvo:
The warning explicitly stated that Octopus Migrator wasn't to be used for migrating projects between spaces. Now that Bento is released, I've updated the warning to link to the blog.


harrisonmeister (approved):
LGTM

PR #1299, opened 2021-06-14
Change metrics command to deprecated

harrisonmeister:


charles-hague-octopus (approved):
Removed in 2020.5 as per https://octopus.com/downloads/compare?from=2020.1.10&to=2020.6.4855 and linked changes.

matt-richardson (approved):
Seems reasonable.
Do we want to say why?
Do we want to refer to a github issue or similar?
(not sure if the answer is yes to either of them, but thought it useful to ask)

harrisonmeister:
> Seems reasonable.
> Do we want to say why?
> Do we want to refer to a github issue or similar?
> (not sure if the answer is yes to either of them, but thought it useful to ask)

I followed the convention I found with the [external-worker](https://octopus.com/docs/octopus-rest-api/octopus.server.exe-command-line/external-worker) command.

Thinking about this some more, would it be better to have the pages removed, and then link to a deprecated commands section?

matt-richardson:
> > Seems reasonable.
> > Do we want to say why?
> > Do we want to refer to a github issue or similar?
> > (not sure if the answer is yes to either of them, but thought it useful to ask)
> 
> I followed the convention I found with the [external-worker](https://octopus.com/docs/octopus-rest-api/octopus.server.exe-command-line/external-worker) command.
> 
> Thinking about this some more, would it be better to have the pages removed, and then link to a deprecated commands section?

I'm not sure I've got answers here - I think I'll leave the decision with you on how to proceed.

That said, as I think about it, having a deprecated commands section might be useful:
* having it as a menu item might be a bit annoying "oh, yeah, that looks like what I need! oh. deprecated. boo."
* splitting it out to it's own page might cover customers on older versions better?

I'll leave the call with you - you spend much more time thinking about the docs experience than I do :)

harrisonmeister:
Thanks @matt-richardson - I'll have a think on this 😄

harrisonmeister:
For now, I'm going to merge this, and revisit at a later date

PR #1301, opened 2021-06-15
2 minor updates on the Run a runbook CLI page

donnybell:
Fixed --project=VALUE requirement
Fixed  --variable=VALUE from "Label:Value" to "Name:Value"

### DIFF
--- a/docs/octopus-rest-api/octopus-cli/run-runbook.md
+++ b/docs/octopus-rest-api/octopus-cli/run-runbook.md
@@ -15,8 +15,8 @@ Where [<options>] is any of:
 
 Run Runbook:
 
-      --project=VALUE        Name or ID of the project. This is optional if
-                             the runbook argument is an ID
+      --project=VALUE        Name or ID of the project that contains the 
+                             Runbook. This is required.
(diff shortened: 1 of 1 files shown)


harrisonmeister:
@donnybell - the cli pages automatically get updated from the commands (a TC build creates PR) so this change would get undone I think. It might be worth checking with F&M if we can get the actual code changed to reflect your changes?

matt-richardson:
Code in question is at https://github.com/OctopusDeploy/OctopusCLI/blob/master/source/Octopus.Cli/Commands/Runbooks/RunRunbookCommand.cs#L34.

donnybell:
> @donnybell - the cli pages automatically get updated from the commands (a TC build creates PR) so this change would get undone I think. It might be worth checking with F&M if we can get the actual code changed to reflect your changes?

Thanks @harrisonmeister, I'll create a problem ticket in Zendesk about this and send it over. Thanks!

harrisonmeister:
Closing for now as I've raised https://github.com/OctopusDeploy/OctopusCLI/pull/116

PR #1313, opened 2021-07-02
Update Lifecycles with more clarity surrounding auto/manual deployments including tenants

N-lson:
# Background

When working on a fix for lifecycle auto-deployments to tenants, it was unclear what the expected behavior should be. As it turns out, the behavior actually is listed already in the `Tenants and automatic-environments` hint but it's in the middle of the `Create a new lifecycle` section. I think it makes more sense for the behavior to be outlined in the `Phases` section, along with adding some extra context for the difference between manual vs. auto deployments. There's also an issue with the nested list in the hint which has been fixed.

# Result
image


harrisonmeister (approved):
LGTM

PR #1314, opened 2021-07-05
Categorise Octostache Filters

rhysparry:
To improve discovery and make it clearer where new filters might belong
this splits up the filters into a series of categories. The
filter-specific documentation was moved to sit under its appropriate
category.


harrisonmeister (approved):
Made some minor line edits (anchor link format) and added a summary of filter categories

PR #1317, opened 2021-07-13
Updating kill tasks to handle pagination

BobJWalker:
Removed the PS c# example because getting it to work with the client with a list object was proving to be very frustrating


harrisonmeister (approved):
LGTM

PR #1340, opened 2021-08-25
Update usages of GetByPartialName in go examples

harrisonmeister:


twerthi:
I'm wondering if we should refactor functions that use `Name` in the query struct to not be recursive as they should only return one result.  Thoughts?

PR #1362, opened 2021-10-06
Update authentication matrix compatibility for cloud and auth providers

harrisonmeister:


millerjn21 (approved):
LGTM Pending your question for Username/Password with the Cloud team. If it's not usable maybe we should disable it on cloud instances to avoid confusion.

harrisonmeister:
> LGTM Pending your question for Username/Password with the Cloud team. If it's not usable maybe we should disable it on cloud instances to avoid confusion.

Agreed, it could potentially need a further tweak :)

PR #1399, opened 2021-11-24
Fix up images to use Octopus style

ciaracarey:


ciaracarey:
@DevOpsDerek 
I updated the images to include the whole frame, I think I'm getting there!

DevOpsDerek:
The images all look perfect. The table doesn't look right so I'll have a look at that and see what I can do. 

Available on: https://preprod.octopus.com/docs/packaging-applications/package-repositories/guides/cloudsmith-feed?l=1

PR #1416, opened 2021-12-16
Changed Homepage URL in Azure AD Authentication

Clare-Octopus:
Initially changed the 'Sign on URL' to 'Homepage URI' and added a picture to put in the Branding section of Azure as Azure has moved it from the previous docs link.

See ticket for further explanation -

https://octopus.zendesk.com/agent/tickets/80404


harrisonmeister (changes requested):
Can you confirm if Home Page URL is definitely a requirement of Octopus Azure AD auth. As far as I know, it isn't.

Clare-Octopus:
The customer couldn't authenticate with Azure AD until he put that URL in the Homepage URL under Branding. There was another ticket where that also happened (confirming this is the case) but I can't find that ticket now as I was going to link that too.

https://your.octopus.app/api/users/authenticatedToken/AzureAD is required in the Homepage URL (for some reason) even though the Microsoft Documentation suggests as you do (they say it is just the URL of your octopus instance). Both users set their homepages to the above and it worked.

When I created Azure SSO with a new VM I followed the documentation and it did work with the homepage URL in, I then took it out and it still worked so I cannot confirm 100% that this is a requirement. However, with two users saying it did not work, and as soon as they put that URL in the homepage URL it then works I'm betting it's a requirement. It doesn't seem to affect the instance negatively if it is put in.

Logout URL is still under Authentication

I did change Enable ID Token to Homepage URL and re-pushed the git, hopefully, you see that change?

harrisonmeister (approved):
Reverted changes as discussed

PR #1424, opened 2022-01-04
Shaunh/notes on creating custom worker tool images

shaun-od:
Just adding a note to highlight the need to create a custom image if the provided images don't meet the needs of the customer. Also wanted a way to highlight the detailed blog post on creating custom images.


harrisonmeister (approved):
LGTM

PR #1425, opened 2022-01-04
removed the claim about default shell

FinnianDempsey:
(It does not need to be the user’s default shell.)


FinnianDempsey:
See ticket https://octopus.zendesk.com/agent/tickets/80929

harrisonmeister (approved):

PR #1432, opened 2022-01-19
Fix Golang script example for deployment release with prompted variables

a-kuharenko:
Current script does not pass `FormValues` properly. 
API expects UUID of variable as key instead of variable name.
For example: 
```"FormValues": {"8f457792-62f1-d607-2194-cd9846235130": "MyValue"}}```

Currently it is:
```"FormValues": {"MyVariable": "MyValue"}}```

### DIFF
--- a/docs/shared-content/scripts/deploy-release-with-prompted-variables-scripts.include.md
+++ b/docs/shared-content/scripts/deploy-release-with-prompted-variables-scripts.include.md
@@ -385,7 +385,7 @@ func main() {
 
 		// check to see if the element display name matches
 		if (element["Control"].(map[string]interface{}))["Name"] == parsedPromptedVariable[0] {
-			formElements[element["Control"].(map[string]interface{})["Name"].(string)] = parsedPromptedVariable[1]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by chulanovskyi-bs — APPROVED** [review]

File: docs/shared-content/scripts/deploy-release-with-prompted-variables-scripts.include.md [hunk 1]
```diff
@@ -385,7 +385,7 @@ func main() {
 
 		// check to see if the element display name matches
 		if (element["Control"].(map[string]interface{}))["Name"] == parsedPromptedVariable[0] {
```
twerthi, line 387: Would this IF block need to be updated like the variable assignment below?

a-kuharenko: No, here we compare actual variable name, not UUID

twerthi: Ah!  Yes!  You're correct, thank you clarifying.

twerthi (approved):
LGTM

PR #1433, opened 2022-01-19
Removes private link to step-api repo from New-OctopusTarget command page

geofflamrock:
The documentation for dynamically creating deployment targets that are authored with step packages (e.g. an AWS ECS cluster) currently contains a link to learn more about step packages which points to the https://github.com/OctopusDeploy/step-api repo. This repo is currently private and so will result in a 404 for any non-Octopus user of the docs. 

This PR removes the link for the moment. There are plans to make the repo public at which time the link could be restored.


harrisonmeister (approved):
LGTM

PR #1443, opened 2022-01-31
Added include for clarity around using variable filters in conditionals

coryreid:
* Added new include around variable usage in conditional evaluation
* Added heading for the include on variable substitution and run condition pages

Feedback:
* Is there a better way to describe the scenario? Prerelease prefix seemed like an easy example
* Should it be a tiny paragraph, or leave it in a hint block?


harrisonmeister (approved):
LGTM

PR #1471, opened 2022-03-01
API can't handle sending "bytes" data type.

AdamHollow49:
Issue with Create Certificate Python script; adding step to convert BASE-64 certificate data from 'byte' type to utf-8 string.


twerthi (approved):
LGTM

PR #1485, opened 2022-03-10
Remove reference to OCL editor

shipra-octopus:
We have now disabled the option for the user to view/edit their OCL directly from the Octopus UI due to permissioning issues. Updating the docs to remove this.


harrisonmeister (approved):
LGTM (removed the file that was referenced also)

PR #1505, opened 2022-03-31
Remove Fargate requirement for ECS Update Service step

rhysparry:
The Update Amazon ECS Service step works on EC2-backed instances. This removes the documentation directing users that only Fargate clusters are supported.

### DIFF
--- a/docs/deployments/aws/ecs-update-service/index.md
+++ b/docs/deployments/aws/ecs-update-service/index.md
@@ -7,7 +7,7 @@ Octopus supports deploying a new release to an existing ECS Service through the 
 This step provides an opinionated deployment workflow that allows new released to be deployed to an ECS Cluster that is managed externally, for example manually or via Terraform.
 
 :::hint
-The `Update Amazon ECS Step` was added in Octopus **2021.3**. Presently only **Fargate** clusters are supported.
+The `Update Amazon ECS Step` was added in Octopus **2021.3**.
 :::
 
 At a high level, the `Update Amazon ECS Service` step will:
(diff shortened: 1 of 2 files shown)


harrisonmeister:
@rhysparry - I added https://github.com/OctopusDeploy/docs/pull/1505/commits/5f1c582dc7034337d8245a4c8b3e7c9fbfe53fb8 to reflect the fargate requirement changes in the main AWS ECS step too.

Question for you, should [this sentence](https://github.com/OctopusDeploy/docs/pull/1505/files#diff-64802b1435a2f54d427ca07987b6f52a395971f326be29fb8ca629688cd28e7eR6) also be updated:

> This step provides an opinionated deployment workflow that combines a Fargate task definition and service into a single step.

Specifically, is it now an ECS task definition (vs a Fargate task definition) ?

harrisonmeister (approved):
Change LGTM, have left a comment about the wording on the main AWS ECS step. happy for it to be merged as-is, or with that wording changed :)

rhysparry:
We want that wording to remain on the Deploy ECS Step as it does have the restriction, but not for the "Update ECS Service" step which works differently enough that it doesn't have the same restriction. I believe it was originally added to the "Update ECS Service" step as a copy/paste artifact.

harrisonmeister:
> We want that wording to remain on the Deploy ECS Step as it does have the restriction, but not for the "Update ECS Service" step which works differently enough that it doesn't have the same restriction. I believe it was originally added to the "Update ECS Service" step as a copy/paste artifact.

My bad @rhysparry - have reverted the commit in https://github.com/OctopusDeploy/docs/commit/6a37241e38b7b893340fe0d2ed2e51d57e127d00

PR #1517, opened 2022-04-14
Adding instructions for using polling with the same URL

millerjn21:


harrisonmeister:
I don't believe this should be added to this section in the docs. It could lead to a more complicated setup of Octopus (for either a HA or Non-HA setup). An NLB should be able to achieve both polling (using TCP 10943) and web traffic using TCP (Port 80/443) as described [here](https://help.octopus.com/t/how-can-i-configure-my-polling-tentacles-to-hit-my-octopus-deploy-high-availability-instance-to-sitting-behind-an-aws-load-balancer/24890/3).

If you feel it's still worth adding this info, I'd recommend a KB article linking this to the article linked above.

PR #1520, opened 2022-04-19
Use $instance to define file paths

joshgk00:
When moving a tentacle the script wasn't using the $instance variable defined in the config section, so the folder path ended up being hard coded to "InstanceName" instead of the value that was defined in the instance name. This change updates the script so that it will use $instance in the folder paths for $oldHome and $newHome


harrisonmeister (approved):
LGTM, thanks for the contribution @joshgk00 :+1:

PR #1522, opened 2022-04-19
Mention optional phases don't do auto-deploy

N-lson:
Optional phases don't actually execute auto-deployments https://github.com/OctopusDeploy/Issues/issues/7446. Our docs say they follow the standard rules so we should remove that part and add a note on the actual behaviour.


harrisonmeister (approved):
I made minor adjustments to the indentation (not related to your change). I've also upgraded the hint to a warning and added a bold header.

PR #1530, opened 2022-05-04
Updates ECS target authentication

geofflamrock:
This PR updates the ECS cluster target documentation to include details of authentication with worker credentials and assuming IAM roles.

We want to hold off on merging this until we are ready to release the new version of the target with the support.


docs/infrastructure/deployment-targets/amazon-ecs-cluster-target.md
```diff
@@ -17,10 +17,48 @@ Refer to the [AWS documentation](https://docs.aws.amazon.com/AmazonECS/latest/de
 3. Enter a display name for the Amazon ECS Cluster.
 4. Select at least one [environment](/docs/infrastructure/environments/index.md) for the target.
 5. Select at least one [target role](/docs/infrastructure/deployment-targets/index.md#target-roles) for the target.
-6. In the **ECS Cluster** section:
+6. In the **Authentication** section (see the [authentication](#authentication) section for more information):
```
gb-8, line 20: Might want to avoid the repeat of "authentication section" here, especially since it is referring to two different things?

```suggestion
6. In the **Authentication** section (see [Authentication](#authentication) below for more information):
```

docs/infrastructure/deployment-targets/amazon-ecs-cluster-target.md
```diff
@@ -17,10 +17,48 @@ Refer to the [AWS documentation](https://docs.aws.amazon.com/AmazonECS/latest/de
...
+   - Select whether to use an AWS account configured in Octopus or to use credentials from the worker on which your deployment runs.
+     - Select an AWS account if necessary. If you don't have an `AWS Account` defined yet, check our [documentation on how to set one up](/docs/infrastructure/accounts/aws/index.md).
```
gb-8, line 22: The bullets used here seem a bit clunky in that they make the nested bullet item seem higher level than the level above. Probably because they  are nested inside the numbered list so the levelling wraps.

I'm not 100% sure whether the nesting is even deliberate here. I think it works either way.

I think this one is probably outside the scope of the change, and could be left.

image

gb-8 (approved):
Looks good.
I commented a couple of very minor queries, but have pre-approved anyway.

docs/infrastructure/deployment-targets/amazon-ecs-cluster-target.md
geofflamrock: Hmmm yeah that is a bit awkward, I introduced the second level of nesting I think, I'll try and rework it a bit to be a bit nicer.

PR #1537, opened 2022-05-13
Add Private Space doc

iryan2:
#team-user-onboarding has been working on adding a feature called Private Space, which is now ready to ship. We'd like to include a link in the help sidebar to the docs (a "Learn more" link would appear here under "and learn Octopus"):
image

In private-space.md, can you let me know if the relative link to the page on Spaces is correct? Should it be absolute instead (`/docs/administration/spaces/index.md`)?

### DIFF
--- a/docs/administration/spaces/private-space.md
+++ b/docs/administration/spaces/private-space.md
@@ -5,7 +5,7 @@ position: 30
 version: 2022.2
 ---
 
-As the name suggests, a Private Space is just like a normal [Space](./index.md), except it’s just for you. No one else on your team will be able to view or edit any projects in your private space. It allows users a safe environment to learn basic Octopus concepts. It’s also a great place to experiment with workflows that aren’t ready to be shared with the rest of your company.
+As the name suggests, a Private Space is just like a normal [Space](/docs/administration/spaces/index.md), except it’s just for you. No one else on your team will be able to view or edit any projects in your private space. It allows users a safe environment to learn basic Octopus concepts. It’s also a great place to experiment with workflows that aren’t ready to be shared with the rest of your company.
 
 Each Private Space will soon be populated with a sample project to provide an example of a real-world Octopus setup.
 
--- a/docs/administration/spaces/private-space.md
+++ b/docs/administration/spaces/private-space.md
@@ -1,0 +1,15 @@
+---
+title: Private Space
+description: Information on Private Spaces
+position: 30
+version: 2022.2
+---
+
+As the name suggests, a Private Space is just like a normal [Space](./index.md), except it’s just for you. No one else on your team will be able to view or edit any projects in your private space. It allows users a safe environment to learn basic Octopus concepts. It’s also a great place to experiment with workflows that aren’t ready to be shared with the rest of your company.
+
+Each Private Space will soon be populated with a sample project to provide an example of a real-world Octopus setup.
+
+Note:
+
+1. Deployment targets created in your private space will be counted against your company’s license limits.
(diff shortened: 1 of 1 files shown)


docs/administration/spaces/private-space.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Private Space
+description: Information on Private Spaces
+position: 30
+version: 2022.2
+---
+
+As the name suggests, a Private Space is just like a normal [Space](./index.md), except it’s just for you. No one else on your team will be able to view or edit any projects in your private space. It allows users a safe environment to learn basic Octopus concepts. It’s also a great place to experiment with workflows that aren’t ready to be shared with the rest of your company.
```
harrisonmeister, line 8:
```suggestion
As the name suggests, a Private Space is just like a normal [Space](/docs/administration/spaces/index.md), except it’s just for you. No one else on your team will be able to view or edit any projects in your private space. It allows users a safe environment to learn basic Octopus concepts. It’s also a great place to experiment with workflows that aren’t ready to be shared with the rest of your company.
```

docs/administration/spaces/private-space.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Private Space
+description: Information on Private Spaces
+position: 30
+version: 2022.2
```
harrisonmeister, line 5: FYI - No harm in leaving this in, but I don't think it prevents the page from showing or anything like that. This used to be for when we displayed multiple versions of the docs, which we no longer do.

harrisonmeister:
Hey @iryan2 

> In private-space.md, can you let me know if the relative link to the page on Spaces is correct? Should it be absolute instead (/docs/administration/spaces/index.md)?

Correct, we tend to use absolute indexes, mainly for consistency than a specific rule (although I think relative probably would have worked). 

I added a note to the version attribute in the YAML at the top - since Private spaces are only available from Octopus 2022.2, I'd probably suggest a hint at the very top of the page, something like:

```
:::hint
Private spaces were added in **Octopus 2022.2**.
:::
```
Just an example of the wording :smile:

harrisonmeister:
Just a thought, this might be a good page to link to from the [getting started](https://octopus.com/docs/getting-started) page?

It's probably one to discuss with @steve-fenton-octopus and @terence-octo as I believe they manage the getting started page from a content marketing perspective

iryan2:
> Just a thought, this might be a good page to link to from the [getting started](https://octopus.com/docs/getting-started) page?

Good idea but at Private Spaces is currently in "soft launch" mode, because it's pretty empty at the moment. The next thing #team-user-onboarding will be working on is a Sample Project that will live in the private space. Once that's complete marketing will start to publicise the feature, so I'd say it makes sense to hold off on your suggestion until then.

docs/administration/spaces/private-space.md
iryan2: Ah ok thanks for the info. I saw it at the top of spaces/index.md and thought I'd better include one :)

harrisonmeister: Sorry my bad, the title, description and position are required. It was just `version` that can be removed (or kept, it doesn't hurt)

harrisonmeister:
> 

Sounds good :+1:

docs/administration/spaces/private-space.md
iryan2: Oh that makes sense! Thanks

harrisonmeister (approved):
LGTM

PR #1552, opened 2022-05-31
Added hint about common errors when trying to use SSH keys as certificates

FinnianDempsey:
Feel free to adjust as desired! 

[F&m thread ](https://octopusdeploy.slack.com/archives/CNHBHV2BX/p1653872249649439)


harrisonmeister (approved):
Split the errors out, as the hint didn't look consistent with the other errors

PR #1553, opened 2022-05-31
Adds outbound requests for step package public feed

geofflamrock:
A step package public feed was introduced in `2022.1` to support updating existing and creating new steps to be shipped out of band of Server releases. This feed is hosted in Azure and is available on a well-known URL `https://steps-feed.octopus.com`. 

Self-hosted customers may have their Octopus instance behind a firewall and as such may not be able to use the feed without configuring some exceptions. The feed returns some redirects to Azure services as part of servicing requests so multiple exceptions will need to be configured.

This PR adds details to the outbound connections documentation on which URLs Octopus reaches out to. It also moves some existing detail on community step templates into it's own section.


andrewabest (approved):
LGTM

PR #1559, opened 2022-06-15
Update CaC reference for slugs

YuKitsune:
Update existing documentation to use slugs instead of names.

### DIFF
--- a/docs/projects/version-control/config-as-code-reference.md
+++ b/docs/projects/version-control/config-as-code-reference.md
@@ -143,7 +143,8 @@ Currently, Octopus creates the following files:
 The _deployment_process.ocl_ file contains the configuration for your project's steps. Below is an example _deployment_process.ocl_ for a project containing a single _Deploy a Package_ step.
 
 ```hcl
-step "Deploy a Package" {
+step "deploy-a-package" {
+    name = "Deploy a Package"
     properties = {
         Octopus.Action.TargetRoles = "web"
     }
@@ -156,14 +157,14 @@ step "Deploy a Package" {
             Octopus.Action.Package.AutomaticallyRunConfigurationTransformationFiles = "True"
             Octopus.Action.Package.AutomaticallyUpdateAppSettingsAndConnectionStrings = "True"
             Octopus.Action.Package.DownloadOnTentacle = "False"
-            Octopus.Action.Package.FeedId = "Octopus Server (built-in)"
+            Octopus.Action.Package.FeedId = "octopus-server-built-in"
             Octopus.Action.Package.PackageId = "webConfig"
         }
         worker_pool_variable = ""
 
         packages {
             acquisition_location = "Server"
-            feed = "Octopus Server (built-in)"
+            feed = "octopus-server-built-in"
             package_id = "webConfig"
             properties = {
                 SelectionMode = "immediate"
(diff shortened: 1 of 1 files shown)


docs/projects/version-control/config-as-code-reference.md
```diff
@@ -213,17 +214,22 @@ connectivity_policy {
...
     donor_package {
-        step = "Deploy a Package"
+        step = "deploy-a-package"
     }
 }
 ```
 
+## Slugs in OCL
+
+Previously, version controlled projects would reference shared resources using their name. This had a side-effect causing API responses for version controlled projects to contained names instead of IDs.
```
dylanlerch, line 224:
```suggestion
Previously, version controlled projects would reference shared resources using their name. This had a side-effect causing API responses for version controlled projects to contain names instead of IDs.
```

docs/projects/version-control/config-as-code-reference.md
```diff
@@ -213,17 +214,22 @@ connectivity_policy {
...
+This has recently changed. Shared resources are now referenced from OCL by their slug. IDs will be used in API responses instead of names.
+
 ## Items of note
 
 When designing the config-as-code feature, we made several decisions to keep an appropriate balance of usability and functionality. There are a few limitations and items of note you should be aware of with config-as-code.
 
 - The Octopus Terraform Provider and OCL are not a 1:1 match. You cannot copy resources between the two and expect everything to work. We want to narrow the gap as much as possible, but as of right now, a gap exists.
-- Shared resources (environments, external feeds, step templates, etc.) are referenced by name, not ID. Be very careful when changing the names of any shared resources.  
+- Shared resources (environments, external feeds, step templates, etc.) are referenced by their slug, not ID.
```
dylanlerch, line 232: Should we also clarify here that it's by slug in OCL, but ID in the API?

YuKitsune: Good call, updated the wording here 👍

dylanlerch (approved):

shipra-octopus:
Looks good to me @YuKitsune Is there some place we can link to explain what are slugs in the "Slugs in OCL" section? If not, that's ok. We can link it later to your blog post when ready. Thanks!

YuKitsune:
Thanks @shipra-octopus 
We don't have any generic slug documentation at the moment. I think we'll supplement this with the blog post once it's ready 👍

shipra-octopus:
Just thought of another one - is it useful to add the list of "shared resources" that now use slugs to the docs?

YuKitsune:
@shipra-octopus No harm in listing them. I've updated the docs with a list of resources.

docs/projects/version-control/config-as-code-reference.md
```diff
@@ -213,17 +214,36 @@ connectivity_policy {
...
     donor_package {
-        step = "Deploy a Package"
+        step = "deploy-a-package"
     }
 }
 ```
 
+## Slugs in OCL
+
+Previously, Git projects would reference shared resources using their name. This had a side-effect causing API responses for Git projects to contain names instead of IDs.
```
zentron, line 224: How previously? How recently. Add some versions here, even if we need to update it just after it goes out.
`Prior to version 2022.3.x`

zentron (approved):

YuKitsune:
Preprod preview: https://preprod.octopus.com/docs/projects/version-control/config-as-code-reference#slugs-in-ocl

PR #1567, opened 2022-06-22
Explain that Private Space is a cloud-only feature

iryan2:
The Private Space feature has caused a bit of confusion with users that were expecting it to be available for on-premises installations. Adding this note should help them understand that they'd have to switch to Cloud if they want to use the feature.

### DIFF
--- a/docs/administration/spaces/private-space.md
+++ b/docs/administration/spaces/private-space.md
@@ -5,7 +5,7 @@ position: 30
 ---
 
 :::hint
-Private spaces were added in **Octopus 2022.2** as a Cloud-only feature. The feature is not currently available for on-premises instances.
+Private spaces were added in **Octopus 2022.2** as a Cloud-only feature. The feature is not currently available for self-hosted instances.
 :::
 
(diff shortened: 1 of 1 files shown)


docs/administration/spaces/private-space.md
```diff
@@ -5,7 +5,7 @@ position: 30
 ---
 
 :::hint
-Private spaces were added in **Octopus 2022.2**.
+Private spaces were added in **Octopus 2022.2** as a Cloud-only feature. The feature is not currently available for on-premises instances.
```
harrisonmeister, line 8:
```suggestion
Private spaces were added in **Octopus 2022.2** as a Cloud-only feature. The feature is not currently available for self-hosted instances.
```

harrisonmeister (approved):
LGTM (made a minor suggestion, but happy with it as-is)

iryan2:
> LGTM (made a minor suggestion, but happy with it as-is)

Good call, I've accepted your suggestion.

PR #1579, opened 2022-07-07
Specify specific version number

YuKitsune:


docs/projects/version-control/config-as-code-reference.md
```diff
@@ -221,8 +221,8 @@ versioning_strategy {
 
 ## Slugs in OCL
 
-Prior to version 2022.3, Git projects would reference shared resources using their name. This had a side-effect causing API responses for Git projects to contain names instead of IDs.
-From version 2022.3 onwards, a handful of resources are from OCL by their slug. IDs will be used in API responses instead of names.
+Prior to version 2022.3.4517, Git projects would reference shared resources using their name. This had a side-effect causing API responses for Git projects to contain names instead of IDs.
+From version 2022.3.4517 onwards, a handful of resources are from OCL by their slug. IDs will be used in API responses instead of names.
```
dylanlerch, line 225: Not in this change, but while you're here is this missing a word?

```suggestion
From version 2022.3.4517 onwards, a handful of resources are referenced from OCL by their slug. IDs will be used in API responses instead of names.
```

PR #1585, opened 2022-07-13
modified AV folder exclusion list from Calamari to Tools

FinnianDempsey:


FinnianDempsey:
Have confirmed that the Calamari folder doesn't exist on current installations of Server or Tentacle:
Screen Shot 2022-07-13 at 12 32 11

Screen Shot 2022-07-13 at 12 47 02

docs/security/hardening-octopus.md
```diff
@@ -126,8 +126,8 @@ Write-Output "This value allows Windows Defender to download and install definit
 cscript C:\Windows\System32\Scregedit.wsf /AU 3
 
 Write-Output "Excluding the Calamari folder from Windows Defender..."
```
harrisonmeister, line 128:
```suggestion
Write-Output "Excluding the Tools folder (e.g. Calamari) from Windows Defender..."
```

harrisonmeister (commented):
I've made a small suggestion in the write-output line above the changes. 

One item that might be relevant is should Calamari be kept for customers that haven't upgraded?  My 2 cents is no, since we would expect customers to be on a later version, but just sharing as food for thought :)

wizedkyle (approved):
LGTM

PR #1586, opened 2022-07-13
Clarify treatment of ampersands in XML Files processed by Structured Variable Replacement

mjhilton:
* Explicitly add ampersand (`&`) as a character required to be escaped when replacing text elements in XML documents
* Add reference to the spec for clarity that this is required, not just something we chose to do

This fixes OctopusDeploy/Issues#5907


harrisonmeister (approved):
LGTM

PR #1587, opened 2022-07-13
More clarification of XML escaping behaviour in Structured Variables

mjhilton:
PR #1586 added clarification about how special characters are escaped when replacing values into an XML file using the Structured Variables feature.

This PR adds just a bit more "safety net" for customers to help them understand that any XML libraries reading the resulting file will automatically handle un-escpaing the characters.


twerthi (approved):
LGTM, thank you for making that call out.

PR #1606, opened 2022-07-25
Updated container storage description

adamoctoclose:
I've updated the cache description and added a column to highlight whether the volume mount should be local to the container or in durable storage.


adamoctoclose:
The mount field might need a hint to explain what it means

harrisonmeister (approved):
Made some minor tweaks to make the extra column clearer.

PR #1620, opened 2022-08-03
Re-add windows tentacle docker doc

N-lson:
So we do actually want to keep the Windows docker image for Tentacle.. this PR is for re-adding the docs for this. https://github.com/OctopusDeploy/OctopusTentacle/pull/391


harrisonmeister (approved):
LGTM

PR #1626, opened 2022-08-15
Remove docs warning about issue now 6629 is resolved

harrisonmeister:


veochen-octopus:
I've tested the image from our internal feed and the fix seems to be working 🆗. The earliest version with this fix is `2022.3.8529`. `2022.3` is not GA yet but will be soon. It's probably best to wait until then to update the docs. I'll put this on my to do list.

veochen-octopus (approved):
We're GA. Let's go!

PR #1629, opened 2022-08-18
Adding documentation for database maintenance

sburmanoctopus:
Adding documentation for index rebuilding


docs/administration/data/octopus-database/index.md
```diff
@@ -12,8 +12,22 @@ Octopus Server requires access to a SQL Server to use for storing relational dat
...
 ## Routine maintenance {#maintenance}
 
+:::hint
+This section only applies to self-hosted versions of Octopus Server, [Octopus Cloud](/docs/octopus-cloud/index.md) instances perform database maintenance in the background without user interaction.
+:::
+
 You are responsible for the routine maintenance of your Octopus database. Performance problems with your SQL Server will make Octopus run and feel slow and sluggish. You should implement a routine maintenance plan for your Octopus database. Here is a [sure guide](https://oc.to/SQLServerMaintenanceGuide) (free e-book) for maintaining SQL Server.
 
+That being said, Octopus will perform minimal routine maintenance of the database for you. In particular, it will periodically run the following tasks:
+* Reorganize and rebuild heavily fragmented indexes (this operation is done in the background without any downtime when running on Azure SQL or SQL Enterprise).
```
felixi-octopus, line 22: Thought we removed reorganizing indexes?

docs/administration/data/octopus-database/index.md
```diff
@@ -12,8 +12,22 @@ Octopus Server requires access to a SQL Server to use for storing relational dat
...
+* Update query optimization statistics for all tables.
+
+If your database server does not support online index reorganization, Octopus will perform index maintenance every time your Octopus Deploy instance is updated.
```
felixi-octopus, line 25: Nitpick: Is `upgraded` a more ubiquitous piece of terminology?

docs/administration/data/octopus-database/index.md
```diff
@@ -12,8 +12,22 @@ Octopus Server requires access to a SQL Server to use for storing relational dat
...
+
+The schedules for these tasks can be configured via **{{Configuration, Settings, SQL Maintenance}}**. Each schedule is defined as a [cron expression](https://en.wikipedia.org/wiki/Cron#Overview). These settings can also be overridden via the following environment variables: `OCTOPUS__Sql__IsMaintenanceEnabled`, `OCTOPUS__Sql__IndexMaintenanceSchedule`, `OCTOPUS__Sql__StatisticsUpdateSchedule`.
```
felixi-octopus, line 27: Do we generally communicate env vars to customers as an avenue for configuration? We seemed to have a definitive stance not to do that with feature toggle env vars. Does that apply here?

docs/administration/data/octopus-database/index.md
sburmanoctopus: Good catch! You are totally right.

Clearly this was written before reorganizing was removed 🙂

docs/administration/data/octopus-database/index.md
sburmanoctopus: I think I prefer upgraded also. Changed!

docs/administration/data/octopus-database/index.md
sburmanoctopus: 🤔 That is a good question. This was the documentation we received, so I assumed it was what we wanted. 

@MarkSiedle, @harrisonmeister, are these types of configuration-base environment variables something we should include in the docs, or remove?

I couldn't find a precedent in our docs code base. 

Sure, we refer to environment variables all the time, but in different contexts, and none refer to any `OCTOPUS__` style environment variables.

docs/administration/data/octopus-database/index.md
harrisonmeister: I'm confused. We are adding text in to say indexes are reorganized, but you've both agreed it isn't something we do. Does that mean this sentence that is to be added will be changed or removed?

docs/administration/data/octopus-database/index.md
harrisonmeister: This still says updated?
```suggestion
If your database server does not support online index reorganization, Octopus will perform index maintenance every time your Octopus Deploy instance is upgraded.
```

docs/administration/data/octopus-database/index.md
harrisonmeister: There is no existing precedent afaik. When it comes to new things, put yourself in the shoes of the customer and how we intend this change to be used. Things I would ask myself
- Do we want to let customers know about these options?
- Will we provide these options in the future in a different way (e.g. with a web front end settings page)? 
- Are there any risks in providing these options to customers? Could they set it to a silly schedule and kill their DB for example.

FWIW I think they should be added, and ultimately it's up to you :) I would suggest putting them in a bullet point unordered list though e.g.:

```text
- `OCTOPUS__Sql__IsMaintenanceEnabled`
- `OCTOPUS__Sql__IndexMaintenanceSchedule`
- `OCTOPUS__Sql__StatisticsUpdateSchedule`
```

docs/administration/data/octopus-database/index.md
sburmanoctopus: Sorry, these changes were made, I just didn't get a chance to push them up before I finished yesterday...

docs/administration/data/octopus-database/index.md
emilol: I think _in general_ there's nothing wrong with providing configurable environment variables that are available to our customers. Env var based config can be particularly useful for customers who run Octopus as a docker container.

That being said _a specific caveat_ to this would be developer [feature toggles](https://en.wikipedia.org/wiki/Feature_toggle) that can be configured via environment variables. Where we're using env vars specifically as a feature toggle in order to avoid a long-running feature branch, and the feature is incomplete or could have side-effects for some customers, I think it makes sense not to document it publicly.

I don't think that applies in this case though?

felixi-octopus: > I don't think that applies in this case though?

No, I don't think it does in this case. Agreed it can be useful for container scenarios, but this strikes me as a bigger design decision around where we want to expose configuration. We're essentially increasing the configuration surface and exposing a new contract (as @sburmanoctopus pointed out) which I'd suggest is outside the scope of what we're trying to deliver.

Also, to address Mark's points

> Do we want to let customers know about these options?
> Will we provide these options in the future in a different way (e.g. with a web front end settings page)?

We do want these configurations exposed to the customer and we do provide a way through the web front end for them to manipulate these options. So in effect, these env vars provide an alternative form of configuration.

> Are there any risks in providing these options to customers? Could they set it to a silly schedule and kill their DB for example.

Higher configuration complexity for the customer and unintentional coupling to env vars are the main detractors for me which is why I'd suggest we omit these from the docs.

harrisonmeister: FWIW If there is a way that a customer can configure these, then I wouldn't add the env vars in the docs. I would consider (if not already present) documenting how you can change them on the web front end.

docs/administration/data/octopus-database/index.md
harrisonmeister: All good, thanks for confirming!

harrisonmeister (approved):
Aside from the env vars being removed from the section, it LGTM

docs/administration/data/octopus-database/index.md
sburmanoctopus: A good discussion, thanks everyone 🙂 

I've gone with removing the part about the environment variables.

sburmanoctopus:
[sc-11687]

docs/administration/data/octopus-database/index.md
```diff
@@ -12,8 +12,26 @@ Octopus Server requires access to a SQL Server to use for storing relational dat
...
+* Update query optimization statistics for all tables.
+
+If your database server does not support online index rebuilding, Octopus will perform index maintenance every time your Octopus Deploy instance is upgraded.
+
+The schedules for these tasks can be configured via **{{Configuration, Settings, SQL Maintenance}}**. Each schedule is defined as a [cron expression](https://en.wikipedia.org/wiki/Cron#Overview). 
+
+If you already perform your own scheduled maintenance, and would like Octopus to not do this for you, then you can also disable database maintenance via **{{Configuration, Settings, SQL Maintenance}}**.
+
+:::hint
+Please note that clustered indexes are not rebuilt online.
```
felixi-octopus, line 32:
```suggestion
Please note that online rebuilding of clustered indexes is not currently supported
```

sburmanoctopus: As discussed, this will end up being an online only problem, so we will leave the text as is

PR #1633, opened 2022-08-25
Remove mention of private spaces

egorpavlikhin:
Private spaces have been decommissioned from all Octopus Server versions where they existed. I am not sure if we need to add a note about their removal (we never publicly advertised their existence other than in the docs). In any case the documentation page about them is not relevant anymore.

[sc-8366]


harrisonmeister:
When you remove (or move) pages, an entry in the `redirects.txt` file is required, otherwise the build will fail :smile: 
I've added it for you. Agree I don't think we need to add a removal notice of this feature.

harrisonmeister (approved):

PR #1646, opened 2022-09-05
Revert tentacle compatibility with 2022.2

N-lson:
A fix ([private link](https://github.com/OctopusDeploy/OctopusDeploy/pull/13336)) was added to prevent Tentacles from waiting longer than expected with Octopus 2022.2+. This means we don't need to specify any compatibility issues with this version anymore.


harrisonmeister (approved):
LGTM

PR #1647, opened 2022-09-06
Including new audit log retention permissions

felixi-octopus:
Updating docos to include newly introduced permissions into built-in roles

### DIFF
--- a/docs/security/users-and-teams/default-permissions.md
+++ b/docs/security/users-and-teams/default-permissions.md
@@ -562,6 +562,8 @@ description: A listing of the default permissions for each of the built-in user 
 | --------------------------- | ---------------------------------------- |
 | AdministerSystem            | Perform system-level functions like configuring HTTP web hosting, the public URL, server nodes, maintenance mode, and server diagnostics |
 | ConfigureServer             | Configure server settings like Authentication, SMTP, and HTTP Security Headers |
+| EventRetentionDelete        | Delete archived event files |
+| EventRetentionView          | View/list archived event files |
 | EventView                   | View Events, including access to the Audit screen |
 | TeamCreate                  | Create teams |
 | TeamView                    | View teams |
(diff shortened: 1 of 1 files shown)


docs/security/users-and-teams/default-permissions.md
```diff
@@ -562,6 +562,8 @@ description: A listing of the default permissions for each of the built-in user
 | --------------------------- | ---------------------------------------- |
 | AdministerSystem            | Perform system-level functions like configuring HTTP web hosting, the public URL, server nodes, maintenance mode, and server diagnostics |
 | ConfigureServer             | Configure server settings like Authentication, SMTP, and HTTP Security Headers |
+| EventRetentionDelete        | Delete archived event files |
+| EventRetentionView          | View/list archived event files |
```
sburmanoctopus, line 566: Do we want to add `/list` in the `View/list` part?

I only ask because none of the other permissions do it. So this is more about consistency than anything else.

sburmanoctopus (approved):
LGTM. Personally, I'd drop the /list as commented above, but I'll leave that call with you 🙂

docs/security/users-and-teams/default-permissions.md
felixi-octopus: Matching the description that's in code so it's consistent with that

sburmanoctopus: Lol, it then begs the question if the code is right 🤔 

But whatevs 🙂

felixi-octopus: If the `/` is the concern, there's already a `Create/update build information` entry fwiw

sburmanoctopus: No, the (very small) concern is that we don't use the term `list` anywhere else. And I know we do list things elsewhere 🙂 

But like you pointed out, this doc is just reflecting the code. I just think the code should not have added the word `list`, that's all. 

It's a minor point, so nothing to bother with.

PR #1659, opened 2022-09-12
First deployment guide

egorpavlikhin:
This quarter we shipped a completely new user onboarding experience. It has just rolled out to the Cloud version and will roll out to on-prem customers a bit later (likely next week once we verify everything is good to go). This documentation replaces the old screenshots/descriptions with the new ones.

#team-user-onboarding


harrisonmeister:
You've left these pages untouched:

image 

They are used with the "legacy guide".

This makes the menu structure look a little off. I think you'd be better from a navigation perspective from moving them under the legacy guide. 

The issue with that approach, however is that from the variables page onwards both the legacy and new guide re-use those pages forwards, so I'm not sure on the best approach

egorpavlikhin:
How do I hide the pages from the menu? I think the best option will be to remove the old links from the menu and only keep them in the legacy guide document.

harrisonmeister:
> How do I hide the pages from the menu? I think the best option will be to remove the old links from the menu and only keep them in the legacy guide document.

Unfortunately the only way to "remove" the pages from the menu is to delete the pages from the folder. The menu structure is built based on the github folder structure. 

The issue with that is you still want the pages, so you could create a "legacy" folder and make the `/docs/getting-started/first-deployment/legacy-guide.md` page the index.md of that folder e.g.

rename: `/docs/getting-started/first-deployment/legacy-guide.md` to -> /docs/getting-started/first-deployment/legacy/index.md`

And then move the other pages over.

The one issue that remains is that the new and legacy guide still both use the variables pages onwards. This means that if you create a legacy folder, then those pages will be pointing to the non-legacy folder for variables onwards (pages that for the legacy guide are up one level)

Basically, this is why it's less than straight-forward.

It might be easier just to ditch the legacy guide completely and assume the guide is for the latest version only?

egorpavlikhin:
I have moved the old guide to its own folder. It is ok if some links from the old guide go to the new section in the tree. We will remove the legacy guide at some point in the future (early next year). Right now on-prem instances still have the old user flow.

The PR should be ready to merge.

harrisonmeister (approved):

PR #1673, opened 2022-09-21
Updating event retention maximum days for self-hosted customers

sburmanoctopus:
On-prem can now go up to 3650 days for event retention


felixi-octopus (approved):
LGTM

PR #1697, opened 2022-10-20
Add aws cli auth to docs

IsaacCalligeros95:
Were adding support for the aws cli `aws eks get-token` method as a part of authenticating against eks clusters. For context re these changes this will also now be the preferred approach of auth. If both the aws-cli and aws-iam-authenticator are present the aws-cli will be used. The aws-iam-authenticator is still used if the aws cli isn't on path or has any issues validating the clusterUrl or CLI version.

 See [GH PR](https://github.com/OctopusDeploy/Calamari/pull/929) and [SC](https://app.shortcut.com/octopusdeploy/story/27621/for-deploy-a-kubernetes-container-step-targeting-an-eks-instance-deprecate-aws-iam-auth-for-aws-eks-get-token)

This PR is in preparation for merging those changes into master, it's likely to sit for a little while.


harrisonmeister (approved):

harrisonmeister:
@IsaacCalligeros95 - no worries on it being pending. I've added a label to indicate this :+1:

IsaacCalligeros95:
@harrisonmeister I've added a bit more detail that came about during implementation time. I hope to push people to the correct versions of the aws-iam-authenticator based on their kubectl version. We've had some issues since kubectl 1.24.0 which introduced a breaking change with the endpoints and a bug that added a fair bit of confusion around the topic.

PR #1710, opened 2022-11-03
Typos and amendments with regard to auto-deploys

felixi-octopus:
Some typo fixes.

Validated the following behaviour and confirmed the following is no longer true
```
If the initial deployment of a release was successful but an automatic deployment of that release fails, **Octopus will stop automatically deploying that release**.
```
Auto-deploys still get triggered after an automatic deployment failure for a given release.

[sc-29412]


harrisonmeister (approved):
LGTM

sburmanoctopus (approved):
LGTM

PR #1711, opened 2022-11-03
Amending audit log retention rollout dates

felixi-octopus:
[sc-30076]

Reflecting more accurate rollout dates to better inform customers.


sburmanoctopus (approved):
LGTM.

Technically, some Cloud customers already have this. But I think that's detail no one really cares about 🙂

PR #1712, opened 2022-11-03
Fix csv file generation

olsh:
The affected scripts generate CSV files without the first row.
This PR fixes the bug.


harrisonmeister (approved):
I made some minor changes to simplify and fix the variable scripts

PR #1724, opened 2022-11-17
Update delete-targets-by-role-scripts.include.md

ryanrousseau:
Fixed scripts that added /api/ to the URL twice


harrisonmeister (approved):
LGTM

PR #1729, opened 2022-11-22
Add deployment notes template using Octopus.Template.Last variable

harrisonmeister:


coryreid (approved):
Looks good, thanks for adding this case to the docs 👍

PR #1735, opened 2022-12-01
Adds information about Google Cloud accounts for GKE

mjhilton:
* Adds a section for Google Cloud accounts to the existing K8s target auth options, which was missing
* Adds specific information about changes in `kubectl` from v1.26 onward, which may require changes to the customer's execution environment

This is part of the comms for https://github.com/OctopusDeploy/Issues/issues/7621


IsaacCalligeros95 (approved):
LGTM

PR #1736, opened 2022-12-01
Documented new release notes variables

mcasperson:
This PR adds documentation for the new variables included with the issue https://github.com/OctopusDeploy/Issues/issues/7941.


harrisonmeister (approved):
Looks good to me. One question: on each page, should it have a version mentioned to when the variables are available from?

mcasperson:
> Looks good to me. One question: on each page, should it have a version mentioned to when the variables are available from?

I thought about it, but those version notes become redundant and eventually need to be cleaned up (although I don't think we do as a general rule). I made the decision to leave version numbers out to remove that clean up step.

PR #1737, opened 2022-12-02
Update index doc

akirayamamoto:
We do not automatically drop indexes anymore. I updated the doc to reflect that.


harrisonmeister (approved):
LGTM

PR #1748, opened 2022-12-19
Fixed missing latest version on Octopus Server

akirayamamoto:
After the compatibility doc was rolled back the Octopus Server latest version went missing on the compatibility table. This PR brings it back.


N-lson (approved):
Thanks Akira!

PR #1752, opened 2022-12-20
Dynamic Workers Ubuntu 22.04 updates 15 March 2023

goosmurf:
To be merged on 15 March 2023 when the Dynamic Workers Ubuntu (default) image will be switched to Ubuntu 22.04.

Goals:
* indicate that Ubuntu 22.04 is the current Ubuntu (default)
* warn that Ubuntu 18.04 images will be removed on 3 April 2023
* advise customers having issues with 22.04 that they can revert to 18.04 until 3 April 2023


alxmitch (approved):
LGTM

PR #1762, opened 2023-01-09
Update worker-configuration.md

Porter-dm:
I assume this is correct.  The description earlier didn't make any sense -- it cannot be *at least* every hour and *at most* every 72 hours


Porter-dm:
@harrisonmeister -- I re-created the commit from the CLI instead of from the github webui, and now it has my correct email address.

PR #1763, opened 2023-01-12
Add hints for creating deployment targets from scripts

YihaoOct:
This PR adds a hint section for docs on creating new deployment targets to let customer understand the need to add a health check step after creating new deployment targets, if these new targets requires package acquisition.

### DIFF
--- a/docs/shared-content/infrastructure/create-deployment-targets-warning.include.md
+++ b/docs/shared-content/infrastructure/create-deployment-targets-warning.include.md
@@ -1,3 +1,3 @@
 :::warning
-Package acquisition is not performed automatically after a deployment target is dynamically created from scripts. If following steps require packages to be acquired on these new deployment targets, make sure you add a [health check](/docs/projects/built-in-step-templates/health-check.md) step for the corresponding target roles after the target creation step. This will ensure required packages are downloaded for the new targets.
+Package acquisition is not performed automatically after a deployment target is dynamically created from scripts. If subsequent steps require packages to be acquired on these new deployment targets, make sure you add a [health check](/docs/projects/built-in-step-templates/health-check.md) step for the corresponding target roles after the target creation step. This will ensure required packages are downloaded for the new targets.
 :::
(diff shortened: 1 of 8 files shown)


docs/shared-content/infrastructure/create-deployment-targets-warning.include.md
```diff
@@ -0,0 +1,3 @@
+:::warning
+Package acquisition is not performed automatically after a deployment target is dynamically created from scripts. If following steps require packages to be acquired on these new deployment targets, make sure you add a [health check](/docs/projects/built-in-step-templates/health-check.md) step for the corresponding target roles after the target creation step. This will ensure required packages are downloaded for the new targets.
```
IsaacCalligeros95, line 2: Reads well and is to the point 👍.

`If following steps require packages` could change a little, maybe `If subsequent steps require packages` but that's a personal preference and I'm happy either way 🙂

IsaacCalligeros95 (approved):
LGTM, minor comment but I'm happy with it as is. Might be good to get the word smith @mjhilton to take a look as well though

YihaoOct:
I will leave this open until the end of day in case @mjhilton has any suggestions 😃

docs/shared-content/infrastructure/create-deployment-targets-warning.include.md
```diff
@@ -0,0 +1,3 @@
+:::warning
+Package acquisition is not performed automatically after a deployment target is dynamically created from scripts. If subsequent steps require packages to be acquired on these new deployment targets, make sure you add a [health check](/docs/projects/built-in-step-templates/health-check.md) step for the corresponding target roles after the target creation step. This will ensure required packages are downloaded for the new targets.
```
mjhilton, line 2: I think that the grammar and conciseness of this message is quite good.

If I think about how customers might perceive this, the combination of the "Warning" box and the opening statement creates a bit of a negative view, and opens the question "why is package acquisition _not_ performed automatically?".  Instead of positioning this as a "warning/sorry/this-is-a-bug", we could tweak this statement to give off different vibes.

What if we instead made it something like:

:::hint
If your process creates dynamic deployment targets from a script, and then deploys to those targets in a subsequent step, make sure you add a [full health check step](/docs/projects/built-in-step-templates/health-check.md) for the role of the newly created target after the step that creates and registers the targets. 

This allows Octopus ensure the new targets are ready for deployment by staging packages required by steps that perform the deployment.
:::

mjhilton:
> LGTM, minor comment but I'm happy with it as is. Might be good to get the word smith @mjhilton to take a look as well though

Smith'd ;)

docs/shared-content/infrastructure/create-deployment-targets-warning.include.md
YihaoOct: Great advice! That certainly makes it sounds better from a customer's perspective 👍

PR #1768, opened 2023-01-16
Added reference to Tenant Variables

donnybell:


harrisonmeister (approved):
Re-worked the wording to remove double call-out

PR #1790, opened 2023-02-02
Add section on user api key management

markocto:
I've added a User API Key Management section that details what happens when a User account is disabled or deleted.

### DIFF
--- a/docs/security/users-and-teams/index.md
+++ b/docs/security/users-and-teams/index.md
@@ -25,11 +25,13 @@ You should create a different User account for each person that will use Octopus
 :::
 
 ### User API Key Management
+
 There are some things to be aware of when deleting or disabling an Octopus User:
-- If a User account were deleted, any associated API keys would also be deleted and stop functioning.
-- API keys cannot be transferred between Users, so if an in-use key has the associated user account deleted, that API key will no longer function, and a new API key from an active User would need to be created and used.
-- Additionally, any scripts that reference the deleted API key would need to be updated to the new API key.
-- A disabled user's API keys will not function. An attempt to use it will throw a 401 unauthorized error, and would require the User to be re-enabled if you needed to use that API key.
+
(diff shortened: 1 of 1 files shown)


docs/security/users-and-teams/index.md
```diff
@@ -24,6 +24,13 @@ You can manage users and teams in the Octopus Web Portal:
 You should create a different User account for each person that will use Octopus Deploy. You should create a different [Service account](/docs/security/users-and-teams/service-accounts.md) for each service that will integrate with Octopus Deploy.
 :::
 
+### User API Key Management
+There are some things to be aware of when deleting or disabling an Octopus User:
+- If a User account were deleted, any associated API keys would also be deleted and stop functioning.
+- API keys cannot be transferred between Users, so if an in-use key has the associated user account deleted, that API key will no longer function, and a new API key from an active User would need to be created and used.
+- Additionally, any scripts that reference the deleted API key would need to be updated to the new API key.
+- A disabled user's API keys will not function. An attempt to use it will throw a 401 unauthorized error, and would require the User to be re-enabled if you needed to use that API key.
+
```
harrisonmeister, line 33:
```suggestion
### User API Key Management

There are some things to be aware of when deleting or disabling an Octopus User:

- If a User account is deleted, any associated API keys will also be deleted and stop functioning.
- API keys cannot be transferred between Users, so if an in-use key has its associated User account deleted, that API key will no longer function, and a new API key from an active User would need to be created and used.
- Additionally, any scripts that reference a deleted API key need to be updated to a new API key.
- A disabled user's API keys will not function. An attempt to use it will throw a `401 unauthorized error` and require the User to be re-enabled if you intend to use that API key.

```

markocto: that 401 formatting change works for me

markocto: Oh, and the line break too! 👍

harrisonmeister (approved):
LGTM

PR #1795, opened 2023-02-03
Added the Tools folder for Octopus Server into the Malware Protection for Windows Defender Script

Clare-Octopus:
The tools folder for windows server is located in `C:\Octopus\OctopusServer\Tools` and for tentacles it's `C:\Octopus\Tools`. I noticed the other day we needed to add the changed folder structure for Octopus Server as it's no longer located in the same folder as the tentacles.

### DIFF
--- a/docs/security/hardening-octopus.md
+++ b/docs/security/hardening-octopus.md
@@ -125,7 +125,11 @@ Write-Output "Setting Windows Update to 'Download updates but let me choose whet
 Write-Output "This value allows Windows Defender to download and install definition updates automatically, but other updates are not automatically installed."
 cscript C:\Windows\System32\Scregedit.wsf /AU 3
 
-Write-Output "Excluding the Tools folder (e.g. Calamari) from Windows Defender..."
+Write-Output "Excluding the Tools folder for Octopus Server (e.g. Calamari) from Windows Defender..."
+Add-MpPreference -ExclusionPath "C:\Octopus\OctopusServer\Tools"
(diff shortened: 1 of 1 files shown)


docs/security/hardening-octopus.md
```diff
@@ -125,7 +125,11 @@ Write-Output "Setting Windows Update to 'Download updates but let me choose whet
 Write-Output "This value allows Windows Defender to download and install definition updates automatically, but other updates are not automatically installed."
 cscript C:\Windows\System32\Scregedit.wsf /AU 3
 
-Write-Output "Excluding the Tools folder (e.g. Calamari) from Windows Defender..."
+Write-Output "Excluding the Tools folder for Octopus Server (e.g. Calamari) from Windows Defender..."
+Add-MpPreference -ExclusionPath "C:\Octopus\OctopusServer\Tools"
```
harrisonmeister, line 129: Does this location depend on where a customer has chosen for the octopus server location for the specified instance?

Clare-Octopus: Yes, if the customer has chosen to put their Octopus Home folder into a different location other than the C drive it will be different but that whole malware script points to C inc the old tools folder for the tentacles.

In the documentation above that script it states:

`Note: you may need to change the excluded folders/files if you install Octopus Server or Tentacle into a different location.`

So we are covered here.

harrisonmeister (approved):

PR #1796, opened 2023-02-07
Dynamic Workers Ubuntu deprecation clarifications

goosmurf:
Feedback indicates that the existing deprecation documentation is misleading as it said that users did not need to do anything.

Update aims to guide users to test upcoming images (such as Ubuntu 22.04) prior to them being made the OS default.


flin-8 (approved):
:shipit:

PR #1808, opened 2023-02-17
Mark l user caching update

markocto:
Applied Systems are migrating from a single Azure-hosted instance to a 6-node HA cluster with 2 UI nodes, and they asked if they still need to enable session persistence on their F5 load balancer's Octopus HA pool in order to deal with users permissions changing but not being immediately reflected across all nodes.

I experimented with my local 2-node HA cluster, and can confirm that the user credential cache on each of the HA nodes is now automatically invalidated and reloaded.

So this is just me updating the docs to reflect this change.

Happy to reword anything if it doesn't make sense! :-)


docs/shared-content/administration/load-balancer-endpoint-info.include.md
```diff
@@ -14,9 +14,9 @@ The Octopus Server node configuration is also returned as JSON in the HTTP respo
 
 We typically recommend using a round-robin (or similar) approach for sharing traffic between the nodes in your cluster, as the Octopus Web Portal is stateless. 
 
-Each node in the cluster keeps a local cache of data including user permissions. There is a known issue that occurs when a users permissions change. The local cache is only invalidated on the node where the change was made. This will be resolved in a future version of Octopus. 
+Each node in the cluster keeps a local cache of data including user permissions. This cache is reloaded when a user's permissions change.
```
harrisonmeister, line 17: Suggest we remove the sentence. it doesnt make sense to mention this (it was added as a direct result of needing session persistence)
```suggestion
```

harrisonmeister (changes requested):
Suggest to remove the sentences completely.

markocto:
Done!

How's that look?

harrisonmeister (approved):

PR #1811, opened 2023-02-22
Remove UserVoice links

colinbowern:
We're replacing UserVoice with a new roadmap portal. Two documentation pages have had their links replaced.


harrisonmeister (approved):
LGTM

PR #1812, opened 2023-02-23
Repair events link in archived-audit-events

rain-on:
The link on the event archives was `/api/event/archives` - its not been updated to be 'event**s**', which from testing (and swagger) appeared to be correct.


harrisonmeister (approved):
LGTM

PR #1835, opened 2023-03-20
Self service purchasing docs inclusion

jedlehmann-prog:
Update docs page to include self-service purchasing workflow.


NickJosevski (approved):
Nice!

PR #1836, opened 2023-03-22
Updated version number of OSX

stevencl840:
After updating the version of OSX used in our build chain, we need to update the version references in our documentation to match.

See https://octopusdeploy.slack.com/archives/C04UP41MCCW


matt-richardson (approved):
Nice one - thanks for that!

PR #1848, opened 2023-05-03
Clarifying cloud target removal conditions.

twerthi:
Updates the cloud target removal hint with clarifying information regarding what conditions need to be met before removal.


andrewabest (approved):
Looks good, thanks Shawn!

PR #1863, opened 2023-05-17
Add deprecation notice for Create Branch V1

YuKitsune:
Adds a deprecation notice for the Create Branch V1 endpoint.

I've also re-formatted the existing deprecation notices so that each deprecation is under it's own heading, making it easier to link to and add additional details.

[sc-49425]

### DIFF
--- a/src/pages/docs/deprecations/index.md
+++ b/src/pages/docs/deprecations/index.md
@@ -21,6 +21,13 @@ Deprecations have the following lifecycle:
 Deprecations are subject to change in detail or timeframe. If you need help assessing the impact of deprecation of a feature on your particular Octopus Server configuration, please contact our [support team](https://octopus.com/support).
 :::
 
+## Deprecations for 2023.2
+
+### Create Branch V1
+
+The `POST` method on the `/projects/{projectId}/git/branches` endpoint for version controlled projects is being removed in future versions of Octopus. The same functionality is available using the `/projects/{projectId}/git/branches/v2` endpoint, however, a minor change will need to be made to the request payload.
+The `CurrentBranchName` field has been replaced with `BaseGitRef`. The value of this field should be a fully-qualified git ref (E.g: `refs/heads/main` for the `main` branch, `refs/tags/v1.2.3` for the `v1.2.3` tag, or a commit hash).
+
 ## Deprecations for 2023.1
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/deprecations/index.md
```diff
@@ -21,11 +21,25 @@ Deprecations have the following lifecycle:
 Deprecations are subject to change in detail or timeframe. If you need help assessing the impact of deprecation of a feature on your particular Octopus Server configuration, please contact our [support team](https://octopus.com/support).
 :::
 
+## Deprecations for 2023.2
+
+### Project level `/git/branches` API endpoint
+
+The `POST` method on the `/projects/{projectId}/git/branches` endpoint for version controlled projects is being removed in future versions of Octopus. The same functionality is available using the `/projects/{projectId}/git/branches/v2` endpoint, however, a minor change will need to be made to the request payload.
+The `CurrentBranchName` field has been replaced with `BaseGitRef`. The value of this field should be a fully-qualified git ref (E.g: `refs/heads/main` for the `main` branch, `refs/tags/v1.2.3` for the `v1.2.3` tag, or a commit hash).
```
rosslovas, line 29:
```suggestion
The `CurrentBranchName` field has been replaced with `BaseGitRef`. The value of this field should be a fully-qualified git ref (e.g: `refs/heads/main` for the `main` branch, `refs/tags/v1.2.3` for the `v1.2.3` tag, or a commit hash).
```

rosslovas (approved):
Looks good and reads well, I've left one super minor suggestion but otherwise ✅ 

One 💭: depending on the target audience, it _might_ be worth considering reframing things in terms of an "upgrade guide". Real quick unpolished example I came up with:


> The `POST` method on the `/projects/{projectId}/git/branches` endpoint for version controlled projects will be removed in a future version of Octopus, in favour of the `/projects/{projectId}/git/branches/v2` endpoint.
>
> #### Upgrade guide
> * Change the target URL from `/projects/{projectId}/git/branches` to `/projects/{projectId}/git/branches/v2`.
> * Remove `CurrentBranchName` from the payload. This field has been replaced by `BaseGitRef`. Instead of a branch name, the value of this field should be a fully-qualified git ref (e.g: `refs/heads/main` for the `main` branch, `refs/tags/v1.2.3` for the `v1.2.3` tag, or a commit hash).

Not a big deal though, I'm happy with the current wording, just an idea that may/may not be appropriate 🙂

PR #1872, opened 2023-05-22
Add Jira project type bullet point

markocto:
I've added a Jira project type bullet point to the https://octopus.com/docs/approvals/jira-service-management doc based on feedback from Sean Stanway here: https://octopusdeploy.slack.com/archives/C27FNL3QW/p1684744161165749.

If required, I'm happy to edit the wording as I appreciate it could be said differently / better! :-)


steve-fenton-octopus (approved):
LGTM!

PR #1873, opened 2023-05-22
Move last updated date.

steve-fenton-octopus:
Moves the last updated date to the end of the article.


tonykelly-octopus (approved):
Looks good

PR #1878, opened 2023-06-05
Escape table characters

steve-fenton-octopus:
Table cells are separated by pipes `|`, so any pipes in the table need to be escaped - even within code blocks in the table.

## Before

The pipe causes the table cell to end

## After

The pipe is now part of the code block


harrisonmeister (approved):
LGTM

PR #1883, opened 2023-06-06
Table formatting and notes

steve-fenton-octopus:
This makes sure the table icons will remain neatly arranged when inline notes are added.

The inline notes are now linked to the explantion.

New notes layout


harrisonmeister (approved):
LGTM

PR #1885, opened 2023-06-07
Remove the "index" suffix from redirects

steve-fenton-octopus:
The "index" suffix refers to the markdown file, but on the server we serve from the folder.


harrisonmeister (approved):
LGTM

PR #1886, opened 2023-06-07
Fix invalid fragments (remove :)

steve-fenton-octopus:
Fragments can't include a `:` character.

They should also be short and ideally all lower, but these are not in scope for this fix.


tonykelly-octopus (approved):
LGTM

PR #1897, opened 2023-06-14
Resolves tab issue in Firefox

steve-fenton-octopus:
# Technical detail

`element.role` silently failed in Firefox, so `element.setAttribute('role'...` needs to be used instead.


tonykelly-octopus (approved):
LGTM

PR #1898, opened 2023-06-14
Fix the size of the bookmark icon

steve-fenton-octopus:
From: https://octopusdeploy.slack.com/archives/C04PQ8XJ4CB/p1686716227088579

Make the icons all the same size, rather than relative to the heading size.


tonykelly-octopus (approved):
LGTM

PR #1900, opened 2023-06-15
update-mono-docs

zentron:
Added note describing pending Mono deprecation and called out that we do not test against the latest versions of Mono


steve-fenton-octopus (approved):
🚀 LGTM!

PR #1901, opened 2023-06-15
Update index.mdx

EnterTheMatt-trix:
If my edit is accurate/helps clarify*


steve-fenton-octopus (approved):
Thank you for this suggestion!

PR #1903, opened 2023-06-15
Update worker-pools.md

EnterTheMatt-trix:
Removed possible mistaken plural


steve-fenton-octopus (approved):
Well spotted - thank you once again for the contribition(s) :)

PR #1915, opened 2023-06-28
Update Azure DevOps Deploy Release docs

hnrkndrssn:
- Add a warning that we no longer support specifying `latest` as the release version, and the `release_number` output from the create release v6 step should be used instead.
- Update the deploy release v6 step example to show that it's been split into untenanted/tenanted steps

[sc-51793]


steve-fenton-octopus (commented):
Just one suggested change for you to review, as you might disagree.

steve-fenton-octopus (approved):
🚀 LGTM

PR #1916, opened 2023-06-28
Fixes display issue with mis-interpreted colons

steve-fenton-octopus:
This syntax is reserved by the remark / rehype extensions.

The second column should contain "app:port" but this is being converted into "app<port />" due to markdown extensions.

image

image


tonykelly-octopus (approved):
LGTM

PR #1917, opened 2023-06-28
Updated border gradient technique

steve-fenton-octopus:
Because articles don't have a taxonomy, this isn't currently displayed. However, this keeps the CSS in step with the /devops/ site for gradient borders.


tonykelly-octopus (approved):
LGTM

PR #1922, opened 2023-07-03
Updated accelerator and dependencies

steve-fenton-octopus:
Updates dependencies, which improves:

1. Navigation ordering
2. Search experience

## Navigation ordering

There is a bug fix in accelerator to resolve this issue, where items weren't in the specified order.

## Search experience

The search experience has been extended to include automatically extracted keywords from each page.

The idea is to get a list of meaningful words from the page and make it unique, so the index only contains a small extra set of words (not the whole article). This keeps it relatively small, but extends the search to include some of the content.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,16 +8,16 @@
             "name": "@octopus/docs",
             "version": "0.0.1",
             "dependencies": {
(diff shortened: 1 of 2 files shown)


tonykelly-octopus (approved):
LGTM

PR #1923, opened 2023-07-04
Set location for free trial start button

corey-underdown:
Sets the location to "docs" for the free trial start button so that we can record if someone started a free trial after looking at the docs.


steve-fenton-octopus (approved):
Ace - thanks Corey. I'll make sure we set this in the DevOps site, too.

PR #1927, opened 2023-07-05
The # and { are the wrong way around

steve-fenton-octopus:
Bookmarks in markdown should use `{#bookmark-name}`, but understadably we sometimes use `#{bookmark-name}` by accident.


tonykelly-octopus (approved):
LGTM

PR #1932, opened 2023-07-10
[Testing, do not merge] Updating cli docs for octo.exe

borland:
An automated update of the command line docs for octo.exe version 9.0.0.0.

Created on machine 'ORION-XPS' by 'Orion'


steve-fenton-octopus:
It would be amazing if the automation could update the `modDate:` front matter as this is added to our sitemap to signal changes to search engines.

PR #1943, opened 2023-07-11
Updating cli docs for octo.exe

Octobob:
An automated update of the command line docs for octo.exe version 9.1.7.0.

Created by TeamCity project 'Automated Documentation Updates::Update CLI Docs', build 475

### DIFF
--- a/src/pages/docs/octopus-rest-api/octopus-cli/index.mdx
+++ b/src/pages/docs/octopus-rest-api/octopus-cli/index.mdx
@@ -1,7 +1,7 @@
 ---
 layout: src/layouts/Default.astro
 pubDate: 2023-01-01
-modDate: 2023-01-01
+modDate: 2023-07-11
 title: The Octopus Command Line (CLI)
 description: The Octopus CLI (octo) is the Octopus command line tool that builds on top of the Octopus REST API.
 navOrder: 30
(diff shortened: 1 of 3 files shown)


src/pages/docs/octopus-rest-api/octopus-cli/index.mdx
```diff
@@ -12,45 +12,47 @@ import OctoAutocompleteTabCompletion from 'src/shared-content/octopus-cli/octo-a
 
 <OctopusCli />
 
-## Commands \{#octoCommandLine-Commands}
+## Commands {#octoCommandLine-Commands}
```
borland, line 15: @steve-fenton-octopus I notice that the utility has removed the `\` which was in front of the `{` character.

I'm not familiar with our CMS... Is this a bug, or does it not matter?

steve-fenton-octopus: Good spot - this is going to be annoying, so prepare yourself.

In an MDX file, the `{` character get's interpreted as meaningful, so it needs to be escaped `\{`.

In an MD file, it shouldn't be escaped.

This is a really annoying issue with `.mdx`.

If it makes it simpler, we could make all the files MDX so you don't need to handle differences.

steve-fenton-octopus: I've done a commit to show the areas that need an update, hopefully it's a useful reference:

https://github.com/OctopusDeploy/docs/pull/1943/commits/fc5397d01641498aa5ea57c69b1b8f3d0d47225a

steve-fenton-octopus (approved):

src/pages/docs/octopus-rest-api/octopus-cli/index.mdx
borland: Thanks steve. It's pretty simple to make the utility aware of mdx files and escaping the `{` so I'll go with that. I'll also fix up the paths

PR #1946, opened 2023-07-14
Update accelerator

steve-fenton-octopus:
Updates the accelerator (theme and components).

Search updates improves visibility such as "Log Files" for the search "Logs".

Log files is now 2nd result for logs


harrisonmeister (approved):
LGTM!

PR #1950, opened 2023-07-17
Set semver to 6.3.1 due to security issue

steve-fenton-octopus:
The result of `npm audit fix` on the semver issue.


harrisonmeister (approved):
LGTM

PR #1955, opened 2023-07-18
Node memory setting and layout issue

steve-fenton-octopus:
Resolves the node memory issue by allocating more memory.

Resolves the memory over-use by avoiding default views, which impact markdown imports into MDX files.


harrisonmeister (approved):
LGTM! :rocket:

PR #1956, opened 2023-07-18
Auto-ordering for alphabetical section

steve-fenton-octopus:
Rather than keep renumbering these items, which are displayed alphabetically, we can give them the same number and Astro Accelerator will sort it out for us.


harrisonmeister (approved):
LGTM

PR #1957, opened 2023-07-20
updates to retry and timeout

veochen-octopus:
Changes as per [sc-53872]. Updated both screenshots to include the latest updates as well as included a summary for the timeout condition.


public/docs/projects/steps/conditions/images/conditions.png
veochen-octopus, line 1: Added the `Time out` setting in case not immediately obvious.

hnrkndrssn (approved):
👍

PR #1959, opened 2023-07-23
Change halibut logging documentation

raynhamdev:
# Background

[sc-52877]

As part of https://github.com/OctopusDeploy/OctopusDeploy/pull/19180, we are introducing a new feature to filter Halibut logs in Octopus Server. The server will hide any Halibut logs whose level is lower than the value configured in the new environment variable `OCTOPUS__Logging__Context__Halibut__LogEventLevel`. By default, if there is no value assigned to this environment variable, all logs from Halibut will not appear in Octopus Server.
eg. The environment variable is set to Information, all Halibut logs whose levels are lower than Information level, such as Verbose and Debug, will be filtered out from Octopus Server.

This PR is to update the public doc to reflect this change.

### DIFF
--- a/src/pages/docs/support/log-files.md
+++ b/src/pages/docs/support/log-files.md
@@ -1,7 +1,7 @@
 ---
 layout: src/layouts/Default.astro
 pubDate: 2023-01-01
-modDate: 2023-07-04
+modDate: 2023-07-25
 title: Log files
 description: Octopus Server and Tentacle logs, log file locations, with information about retention and logging levels.
 navOrder: 1
--- a/src/pages/docs/support/log-files.md
+++ b/src/pages/docs/support/log-files.md
@@ -47,7 +47,7 @@ The verbosity of file logging is controlled in the `octopus-log-file` section:
 
 The `minlevel` attribute is most useful for configuring the logging level. Change this value to `Trace` and set `OCTOPUS__Logging__File__LogEventLevel` environment variable to `Verbose` to gather more information.
 
-The Octopus process will automatically switch to the new logging level as soon as the file is saved.
+This process requires a server restart in order to apply the changes to the environment variables.
 
 :::div{.warning}
 **Don&#39;t forget to reset your changes**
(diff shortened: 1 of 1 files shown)


src/pages/docs/support/log-files.md
```diff
@@ -58,17 +58,21 @@ Leaving your `minlevel` too low will impact the performance of Octopus Server. W
 
 To change the logging level for Halibut as logged in the Octopus Server, we follow a similar process as described above with a few changes.
 
-First, find the `octopus.server.exe.nlog` file associated with the application. This is usually in a subfolder of the Octopus "Program Files" folder. **Take a backup** of the file before making changes.
+First, ensure the environment variable `OCTOPUS__Logging__File__LogEventLevel` is set to `Verbose` or an other log level that you want Octopus Server to write to Octopus log file.
```
gian-lorenzetto, line 61: Small typo "an other" should be "another"

src/pages/docs/support/log-files.md
```diff
@@ -47,7 +47,7 @@ The verbosity of file logging is controlled in the `octopus-log-file` section:
 
 The `minlevel` attribute is most useful for configuring the logging level. Change this value to `Trace` and set `OCTOPUS__Logging__File__LogEventLevel` environment variable to `Verbose` to gather more information.
 
-The Octopus process will automatically switch to the new logging level as soon as the file is saved.
+This process requires a server restart in order to apply the changes to the environment variables.
```
michaelongso, line 50: The original statement,
> The Octopus process will automatically switch to the new logging level as soon as the file is saved.

, is referring to the changes for `octopus.server.exe.nlog`, which is accurate. ([ref](https://octopus.com/docs/support/log-files))

raynhamdev: The change to .nlog file is applied as soon as the file is saved. However, the config alone in nlog would not help if `OCTOPUS__Logging__File__LogEventLevel` and the new environment variable are not set.

raynhamdev: Leaving that statement will cause confusion to users as they will think the changes to the environment variables are also automatically applied.

raynhamdev: Discussion: https://octopusdeploy.slack.com/archives/C056PF0S876/p1690179264126229

michaelongso (approved):
Changes LGTM. 👍

PR #1965, opened 2023-08-08
Fonts and font preloading

steve-fenton-octopus:
Sets and preloads fonts.

No visual difference, except where the user's computer didn't have the font, in which case this fixes the rendering of the intended font.


tonykelly-octopus (approved):
LGTM

PR #1971, opened 2023-08-16
Removing stack from this help tip, as this can be named.

briggs-octo:
See ticket here: https://octopus.zendesk.com/agent/tickets/146328


steve-fenton-octopus (approved):
LGTM

PR #1976, opened 2023-08-22
Fix image urls on new pages

steve-fenton-octopus:
The image URLs came up in the test suite. They had a `/images/` part of the path that isn't needed.

I also tested the pages and previous / next buttons. All looks good.


ryanrousseau (approved):
LGTM - I copied that from another guide so we may want to update those eventually.

steve-fenton-octopus:
> LGTM - I copied that from another guide so we may want to update those eventually.

I think there is some historic inconsistency with the image locations. Some get put in /images subfolders and others don't. The tests should pick up on broken internal images and links, so we should be good if they pass.

PR #1982, opened 2023-08-29
Fix preload urls.

steve-fenton-octopus:
Resolves the 404 caused by the incorrect path to the font preload URLs.


tonykelly-octopus (approved):
LGTM

PR #1983, opened 2023-08-29
Missing spaces and invalid double id markdup fixed

steve-fenton-octopus:
See discussion at: https://octopusdeploy.slack.com/archives/C03UDS16R6G/p1693205676000199

When adding a custom ID to a heading, it must have a space to be recognised.

Invalid: `### Heading{#heading-id}`

Valid: `### Heading {#heading-id}`

I looked for other issues and found cases where people were trying to add two ids to the same heading, which isn't possible.

Invalid: `### Heading {#heading-id}{#another-id}`

Fixes included in this PR.


tonykelly-octopus (approved):
LGTM

PR #1985, opened 2023-08-31
Migrate to PNPM and update Astro

steve-fenton-octopus:
Upgrade to Astro 3.

There were some old images with sizes that were ignored in older versions of Astro:

    

These widths needed to be removed in Astro 3.


tonykelly-octopus (approved):
LGTM

PR #1986, opened 2023-08-31
Improve on page search experience

steve-fenton-octopus:
When using on-page search, for example `CTRL` + `F` the page flow means the whole navigation must be searched before the main content is reached. This changes the page flow to improve the exprience by putting the content first.

Examples use: https://octopus.com/docs/getting-started

## Before

All matches in the extensive navigation must be passed before the first useful result is found

## After

The search takes you to the content first, before the navigation

## Credits

Kudos to Lauren and Finnian for pointing out this issue.

cc @laurencagneyoctopus  @FinnianDempsey

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "dependencies": {
         "@astrojs/mdx": "^1.0.0",
         "astro": "^3.0.3",
-        "astro-accelerator": "^0.3.0",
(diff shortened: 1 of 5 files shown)


tonykelly-octopus (approved):
LGTM

PR #1987, opened 2023-09-01
Add docs for Project Tenants page and bulk tenant connections

mjhilton:
In 2023.3 we added a new page to Projects, which allows customers to see and manage the tenants connected to a given project. We also added an ability to bulk-connect many tenants to a project. Previously, to connect tenants to a project, customers would have to go through each tenant one-by-one in a very time-consuming operation. 

This PR adds some basic documentation about the new Project Tenants page and the bulk connection wizard.


N-lson (approved):
LGTM

PR #1990, opened 2023-09-01
If a scrollbar is needed for the menu, just show it

steve-fenton-octopus:
When you hover/focus the navigation, a scrollbar appears if needed.

This looks janky as moving your mouse causes the scrollbar to appear and disappear.

Instead, just show it if it's needed.


tonykelly-octopus (approved):
LGTM

PR #1991, opened 2023-09-01
Astro 3 auto-collapses spaces that we depended on

steve-fenton-octopus:
Tokens separated by spaces used to be preserved:

    { item.A } { item.B }
              ^

But since Astro 3 these spaces are stripped, so you need to force it:

    { item.A + ' ' + item.B }


tonykelly-octopus (approved):
LGTM

PR #1994, opened 2023-09-06
Fixes edge case for navigation on mobile

steve-fenton-octopus:
By setting the source navigation's display property, it overrides the CSS. This prevents that.


tonykelly-octopus (approved):
LGTM

PR #1996, opened 2023-09-07
Move content out of tables.

steve-fenton-octopus:
Moves the content out of a giant table into a structured document flow.

image


tonykelly-octopus (approved):
LGTM

PR #2002, opened 2023-09-20
Fix markdown in MDX file for nested list

steve-fenton-octopus:
The syntax for lists doesn't include an option to us `a. `, `b. `, etc. This may have been a custom extension written for the old docs site.

This list has been updated to use the standard syntax. We can add CSS rules about the numbering system deployed at different levels of nesting if required.


tonykelly-octopus (approved):
LGTM

PR #2006, opened 2023-09-25
Update index.md

RobMcCarther:
Ticket came in regarding 'year' after selecting value outside the "allowed values" limit. The range according to here:https://en.wikipedia.org/wiki/Cron#CRON_expression is 1970-2099


tonykelly-octopus (approved):
LGTM

PR #2009, opened 2023-09-27
Add docs for using OpenID Connect with the Octopus API

geofflamrock:
This PR adds docs on using OpenID Connect (OIDC) to connect to the Octopus API, with specifics for GitHub Actions vs other issuers.


src/pages/docs/octopus-rest-api/openid-connect/index.md
```diff
@@ -0,0 +1,111 @@
...
+modDate: 2023-09-27
+title: Using OpenID Connect with the Octopus API
+description: External systems can use OpenID Connect with service accounts to access the Octopus API without needing to provision API keys
+navOrder: 30
+hideInThisSection: true
+---
+
+Octopus supports using [OpenID Connect (OIDC)](https://openid.net/) to access the Octopus API without needing to provision API keys.
+
+Some of the benefits of using OIDC include:
```
geofflamrock, line 13: I'm wondering if we want some sort of preamble to "what is OIDC" including these benefits here.

src/pages/docs/octopus-rest-api/openid-connect/github-actions.md
```diff
@@ -0,0 +1,29 @@
...
+hideInThisSection: true
+---
+
+Octopus has first-class support for using OpenID Connect (OIDC) within GitHub Actions when using the [`OctopusDeploy/login`](https://github.com/OctopusDeploy/login) action.
+
+:::div{.hint}
+Using OIDC to access the Octopus API is only supported for service accounts, to access the API for a user account please use [an API key](/docs/octopus-rest-api/how-to-create-an-api-key).
+:::
+
+## Configuring an OIDC identity for GitHub Actions
```
geofflamrock, line 17: Prob also need a section on how to filter for different workflow runs e.g. PR vs branch vs environment etc. This seems to be what trips most people up, so need to consider how best to present it.

zentron:
The extension for file `src/pages/docs/octopus-rest-api/index.mdx` is wrong :)

zentron:
Just realised you didnt add that file. Didnt even know that was a thing, ignore my comment 🙉

geofflamrock:
> Just realised you didnt add that file. Didnt even know that was a thing, ignore my comment 🙉

Haha yep I didn't know it was a thing until I saw it and had to look it up. Apparently you can use jsx inside markdown, who knew 🤷.

src/pages/docs/octopus-rest-api/openid-connect/index.md
zentron: Yeah I agree, It would also be nice to link out to an authoritative site that deals with the topic so that you dont need to go into deep detail here and can defer to the expertise elsewhere.

src/pages/docs/octopus-rest-api/openid-connect/index.md
```diff
@@ -0,0 +1,111 @@
...
+
+- API keys do not need to be provisioned and stored in external systems, reducing the risk of unauthorized access to the Octopus API from exposed keys.
+- Access tokens granted by Octopus are short-lived, reducing the risk of unauthorized access to the Octopus API.
+- API keys do not need to be rotated manually by administrators, reducing the risk of disruption when updating to newer keys in external systems.
```
zentron, line 17: Its worth adding a comment/section about how the permissions provided by the token generated are based on the permissions provided by the service account.  This allows the user to configure fine-grained controls for their needs, supporting our recommended principle of least access. Something that helps provide a background of why this architecture was selected

src/pages/docs/octopus-rest-api/openid-connect/index.md
```diff
@@ -0,0 +1,111 @@
...
+After the OIDC identity for GitHub Actions has been created a snippet of the `OctopusDeploy/login` step will be provided which you can use in your workflow to configure the workflow run job to use OIDC authentication.
+
+:::figure
+!['OctopusDeploy/login' snippet](/docs/octopus-rest-api/images/oidc-github-actions-details.png "width=500")
+:::
+
+1. Click the Copy to clipboard to copy the `OctopusDeploy/login` step.
+2. Paste the `OctopusDeploy/login` step into your workflow job.
+3. Add `id-token: write` to the `permissions` on the workflow job. This is required to allow the `OctopusDeploy/login` action to request an OIDC token from GitHub to use.
+   1. Depending on the other steps you have in your workflow, you may need to add additional permissions such as `contents: read`.
```
zentron, line 55: What are those other steps that would require me to set this permission? Is there somewhere we can just link to at GitHub?

src/pages/docs/octopus-rest-api/openid-connect/index.md
```diff
@@ -0,0 +1,111 @@
...
+
+When the workflow runs the `OctopusDeploy/login` action will authenticate with Octopus using OIDC and configure the remainder of the workflow job to work without needing to provide the `server` or `api_key` values.
+
+:::figure
+![Using 'OctopusDeploy/login' in a workflow](/docs/octopus-rest-api/images/oidc-github-actions-example.png "width=500")
+:::
+
+## Getting started with other issuers
+
+To get started using OIDC with other issuers use the below instructions. For more information see [Using OpenID Connect with Other Issuers](./other-issuers.md).
```
zentron, line 66: Why do we split the details for this on this page, then on the other issuers page?
Are we going to repeat the steps here on that page?

zentron (changes requested):

src/pages/docs/octopus-rest-api/openid-connect/index.md
geofflamrock: The idea was that this was supposed to be a minimal "getting started" kind of guide, with further explanation on the specific "other issuers" page. You are right we would most likely repeat the steps in some fashion, but maybe as separate sections with more detail rather than just dot points. I do wonder whether it makes sense to have a getting started for other issuers given how manual it is likely to be for users and different per issuer. It might make more sense for GitHub Actions given its first class.

src/pages/docs/octopus-rest-api/openid-connect/index.md
geofflamrock: An example of other steps that would now require additional specific permissions would be `actions/checkout` which will require `contents: read`. I remember getting tripped up when I first started playing with oidc because of the way it works, by default you don't need to supply any permissions to checkout code, but as soon as you say you want to get the id token you do need to ask for permission to read contents because using the `permissions` property resets it back to nothing. It's a bit of a footgun. 

I'd really like to get to a place with the getting started where someone can follow it and have their workflow work first time, so it felt important to note this in some way, I wonder if having it as a callout with more detail and a link like you suggested might be helpful.

src/pages/docs/octopus-rest-api/openid-connect/index.md
geofflamrock: That's true, the same thing applies when using API keys though right (the permissions come from the service account)? So it's not necessarily a benefit of oidc as such, but yeah could still be worth mentioning, perhaps more in terms of the OIDC identity itself is granular via the specific subject string, so it promotes principle of least access from the external in (if that makes sense).

zentron (approved):

steve-fenton-octopus:
@geofflamrock & @zentron are you happy for this to be merged?

geofflamrock:
> @geofflamrock & @zentron are you happy for this to be merged?

@steve-fenton-octopus yep I think so, I'm planning to merge them in today, if we get any more feedback we can iterate further to improve.

PR #2011, opened 2023-09-28
Simplifies the telemetry request doc page

michaelnoonan:
Previously, this page was brittle because it tried to describe every data element, making it harder to change telemetry, resulting in this page becoming out of date.

The detailed list isn't really needed any more because:

1. Customers can download a preview of the request to see what is being sent.
2. We have our GDPR page which describes our stance on data privacy and our compliance.

### DIFF
--- a/src/pages/docs/security/outbound-requests/telemetry.md
+++ b/src/pages/docs/security/outbound-requests/telemetry.md
@@ -3,7 +3,7 @@ layout: src/layouts/Default.astro
 pubDate: 2023-01-01
 modDate: 2023-09-28
 title: Telemetry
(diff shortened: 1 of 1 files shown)


steve-fenton-octopus:
I updated the meta description to improve it and to trigger the new test action that runs on pull requests.

steve-fenton-octopus:
Hey @jraddon-octopus - I just wanted to check in on this PR. If you need help with the review, let me know.

jraddon-octopus (approved):
Looks good, added an extra commit to remove the frequency, as some of the older versions that still exist currently send 3 times a day.

PR #2012, opened 2023-09-28
Interim search icon

steve-fenton-octopus:
Adds an interim search icon so a search can be easily triggered from any page.

image

Also fixes stacking contexts on desktop and mobile (i.e. skiplinks, navigation, fixed header).


tonykelly-octopus (approved):
LGTM

PR #2021, opened 2023-10-03
Conditionally run deploy steps only if an Octopus URL is present

### LINKED ISSUE #2018: Workflows shouldn't fail when secrets aren't available
I have a fork of this repository with actions enabled (this is a common state for forks and should be encouraged).

When I refreshed `main`, the [`build-astro`](https://github.com/jsoref/OctopusDeploy-docs/actions/runs/6381706010/workflow) workflow [failed](https://github.com/jsoref/OctopusDeploy-docs/actions/runs/6381706010).

It's possible to avoid failing by either checking for secrets -- this requires a bit of:

```yaml
env:
  has_secret1: ${{ secrets.secret1 != '' }}
  if: ${{ env.has_secret1 }}
```

_or_ using an `if: github.repository_owner` -- I don't recommend this practice, but it does work if you don't want to let forks set up their own secrets to run workflows.

steve-fenton-octopus:
Resolves #2018 by conditionally running Octopus steps based on the presence of the URL secret.

### DIFF
--- a/.github/workflows/build-astro.yml
+++ b/.github/workflows/build-astro.yml
@@ -17,6 +17,7 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
+      SHOULD_DEPLOY: ${{ secrets.OCTOPUSSERVERURL != '' }}
       OCTOPUS_URL: ${{ secrets.OCTOPUSSERVERURL }}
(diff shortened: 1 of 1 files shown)


tonykelly-octopus (approved):
LGTM

PR #2022, opened 2023-10-03
Adjust condition for SHOULD_DEPLOY

steve-fenton-octopus:
See #2018 - adjusted condition to handle whether deployment steps should run.


tonykelly-octopus (approved):
LGTM

PR #2023, opened 2023-10-03
Allow people to download microsite

jsoref:
There's a step `Push a package to Octopus Deploy 🐙` which is now optional.
For forks that can't do that step, it's nice to be able to access the zips: https://github.com/jsoref/OctopusDeploy-docs/actions/runs/6394199002


steve-fenton-octopus:
Hey @jsoref 

I was going to ask for this to be conditional, but I notice you've already added the condition - nice one.

Thanks for the submission!

steve-fenton-octopus (approved):
LGTM

PR #2024, opened 2023-10-04
Change Dynamic Worker defaults to Linux

samjetski:
[sc-59991]

We are changing the default worker pool on Octopus Cloud from `Windows (default)` to `Ubuntu (default)`. This PR updates the implied default by changing the examples in the docs to emphasise `Ubuntu`.

This actual change will only affect new customers who sign up from the cut-over date (TBC). We will add a note about the cut-over date in a [follow up task](https://app.shortcut.com/octopusdeploy/story/61174/add-cutover-date-to-public-documentation).

Changes required before merging:

- [x] Update modified date on all pages

### DIFF
--- a/src/pages/docs/infrastructure/workers/dynamic-worker-pools.md
+++ b/src/pages/docs/infrastructure/workers/dynamic-worker-pools.md
@@ -1,7 +1,7 @@
 ---
 layout: src/layouts/Default.astro
 pubDate: 2023-01-01
-modDate: 2023-01-01
+modDate: 2023-10-05
 title: Dynamic Worker pools
 description: Dynamic Worker pools are used in our cloud product to dynamically create and assign workers to running tasks.  This page describes how dynamic worker pools work.
 navOrder: 50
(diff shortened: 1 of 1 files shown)


src/pages/docs/infrastructure/workers/dynamic-worker-pools.md
```diff
@@ -28,26 +28,26 @@ Each worker is provisioned exclusively to a specific customer, and is completely
 
 ## Dynamic Worker Images
 
-Each dynamic worker pool can specify the worker image used. Windows Server Core 2019 is the default. Ubuntu Server 22.04 worker images are also available.
+Each dynamic worker pool can specify the worker image used. Ubuntu Linux 22.04 is the default. Windows Server Core 2019 worker images are also available.
```
samjetski, line 31: In this PR we aim to simply swap the OS emphasis.

For context, this sentence is a little misleading since there are two things going on here: the **default image** version for each OS (Ubuntu -> 22.04), and then which Worker Pool is the initial default (and by extension, which image is used by the default pool). However, we're hoping to address this as part of a broader overhaul in #2025 .

steve-fenton-octopus (approved):
LGTM

PR #2026, opened 2023-10-04
Patch for zod

steve-fenton-octopus:
Fixes a published CVE: https://fluidattacks.com/advisories/swift/

We don't use this feature of zod, but let's update it to remove the risk.


tonykelly-octopus (approved):
lgtm

PR #2030, opened 2023-10-06
Update calc description to describe wrapping variables in { }

rain-on:
The calc variable substitution has been updated such that if a variable appears on the lhs of a calc divide or subtraction operation - they can be wrapped in braces such that the parse can determine if the "/" or "-" is part of the variable, or the operation being performed.

This has necessitated an update to the docs to explain how to best to make use of this solution.

Don't _love_ the code block `{...}` in the warning section, but couldn't _quickly_ work out how to escape the braces!

image

### DIFF
--- a/src/pages/docs/projects/variables/variable-substitutions.mdx
+++ b/src/pages/docs/projects/variables/variable-substitutions.mdx
@@ -93,8 +93,8 @@ Basic mathematical calculations are supported in Octopus using the `calc` statem
 - Division - `/`
 
 :::div{.warning}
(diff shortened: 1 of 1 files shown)


steve-fenton-octopus (approved):
LGTM

steve-fenton-octopus:
@rain-on I've left the PR open as you've also requested a review from @domenicsim1 

Once you're ready, you can merge it to trigger the build and deployment to staging. I'll make sure it makes it along to production.

PR #2034, opened 2023-10-10
Fix stacking for headers.

### LINKED ISSUE #2032: Docs tablist content overlays the header
On the [output variables docs page](https://octopus.com/docs/projects/variables/output-variables#:~:text=PowerShellBookmark&text=From%20a%20PowerShell%20script%2C%20you,conventions%20used%20for%20input%20variables.), the tab lists for different scripting types overlay the page headers. Adding a z-index looks to fix this. I've only tested it in browser on that single page.

image

image

steve-fenton-octopus:
Fixes #2032

## Before

Content is stacked over the header

## After

Content flows beneath the header

### DIFF
--- a/public/docs/css/main.css
+++ b/public/docs/css/main.css
@@ -375,7 +375,7 @@ nav.skip-links a:focus {
     min-height: 55px;
     position: sticky;
     top: 0;
-    z-index: 1;
+    z-index: 10;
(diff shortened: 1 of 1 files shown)


tonykelly-octopus (approved):
LGTM

PR #2035, opened 2023-10-11
Updating C# scripts and references from ScriptCS to dotnet script

IsaacCalligeros95:
This updates references to ScriptCS and example usage to bring the docs in line with the dotnet-script changes. As this is a 2023.4 change it may make more sense to call out each of the individual scripting libraries rather than just update in place as I've done here. The migration is highlighted in the blog post with recent changes here https://github.com/OctopusDeploy/blog/pull/1077


steve-fenton-octopus:
@IsaacCalligeros95 I've merged in changes to resolve conflicts with our new spelling checks, which occur for each PR. Please merge when you're ready (I guess this is waiting for a release?)

steve-fenton-octopus (approved):
LGTM

IsaacCalligeros95:
@steve-fenton-octopus  Thanks Steve, this one will probably be sitting for a while longer, there's a couple of changes to go out and there will be a transition period between the old C# scripting and the new. I'm planning on merging this when most customers have migrated to the newer version.

PR #2041, opened 2023-10-13
Make font path relative to root.

steve-fenton-octopus:
Responding to a warning in Edge, which may be caused by path confusion.


tonykelly-octopus (approved):
LGTM

PR #2044, opened 2023-10-19
De-duplicates the extension and CI word lists.

steve-fenton-octopus:
The word list has been moved into a simple text file - one entry per line.

Removing the settings seems to work - the VS Code extension responds to changes in the custom Octopus dictionary.

I've started with the "almost empty" dictionary so we can have the appropriate discussion around each term. I'd like to make sure we pick the right entry, for example MySql or MySQL, not both.


tonykelly-octopus (approved):
lgtm

PR #2049, opened 2023-10-20
Update dictionary-octopus.txt

tonykelly-octopus:
Added keywords to dictionary list as per Bob's Slack message - https://octopusdeploy.slack.com/archives/C03UDS16R6G/p1697806413861879

### DIFF
--- a/dictionary-octopus.txt
+++ b/dictionary-octopus.txt
@@ -1,1 +1,5 @@
 astro
+runbook
+Runbook
+reprovisioned
+reprovisioning


BobJWalker (approved):
LGTM, it's a shame we have to worry about capitalization.

steve-fenton-octopus:
Hey folks.

We only need to worry about adding the capitalized version if we're pretending something is a proper noun. It understands that all words can be capitalized at the start of a sentence, but with things like "Runbooks", if we want to say "this is a proper noun" and use it like "You can add a Runbook", then we do need to make that clear.

Otherwise, it would allow any word to be capitalized, which would be much worse :)

PR #2053, opened 2023-10-24
Reduce noise in search from legacy upgrades

steve-fenton-octopus:
The legacy upgrade pages are making it hard for folks to find the more relevant newer upgrade pages.

We should reclassify 3x and 4x into the legacy folder and hide all but the legacy overview from the search.

## Before

Results for upgrade

Results for upgrading

## After

Results for upgrade

Results for upgrading


tonykelly-octopus (approved):
LGTM

PR #2054, opened 2023-10-25
Removes the credits file (lots of names in there) from spellcheck.

steve-fenton-octopus:
The credits file is auto-generated and includes many author names. It's better to exclude it from spellcheck.


tonykelly-octopus (approved):
LGTM

PR #2056, opened 2023-10-25
Allows zero-file checks to pass.

steve-fenton-octopus:
If all the changes found in the branch are excluded from checking, it raises an error.

This change supresses errors caused by zero files in the list to check.

## Before

image

## After

image


tonykelly-octopus (approved):
LGTM

PR #2058, opened 2023-10-26
Follow up from the docs call

steve-fenton-octopus:
The first couple of examples added to the non-visible index page.

CODEOWNERS is now owned.

Astro tips added to README to help public contributors.


tonykelly-octopus (approved):
LGTM

PR #2063, opened 2023-10-31
Remove sticky sessions advice

adam-mccoy:
# Background

In previous versions of Octopus Server, permissions changes wouldn't appear on all nodes in a HA setup instantly, due to caches not being invalidated. This resulted in customers experiencing permission errors depending on which node answered their request. To avoid this, it was recommended to use load balancer "sticky sessions", which would ensure a user always talked to the same node.

The cache invalidation issue no longer exists, and has not for at least a year. Thus, the use of sticky sessions to avoid it is not required.

Using sticky sessions also carries other risks. If the node a user is "stuck" to experiences heavy load they may not be able to talk to the API at all.

# Results

Remove the advice to use sticky sessions.

### DIFF
--- a/dictionary-octopus.txt
+++ b/dictionary-octopus.txt
@@ -18,6 +18,7 @@ lifecycles
 maxage
 msiexec
 MTTR
+Netscaler
 nologs
 NTLM
 Octo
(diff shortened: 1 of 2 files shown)


tekerson (approved):
✂️

PR #2064, opened 2023-10-31
Updates accelerator.

steve-fenton-octopus:
This resolves the issue with YT videos that aren't on their own line being expanded.

```markdown
[This video will auto-embed.](https://www.youtube.com/watch?v=FjPO1k0Y-xA)

This video won't [as it is accompanied by text nodes](https://www.youtube.com/watch?v=FjPO1k0Y-xA)

[The full stop is outside the link, so this won't embed](https://www.youtube.com/watch?v=FjPO1k0Y-xA).
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,9 +22,9 @@
     "dependencies": {
         "@astrojs/mdx": "^1.0.0",
         "astro": "^3.0.3",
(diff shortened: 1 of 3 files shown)


tonykelly-octopus (approved):
LGTM

PR #2066, opened 2023-11-02
Frontmatter CMS Support

steve-fenton-octopus:
See https://frontmatter.codes/

Frontmatter might provide useful assistance to people editing docs and it can be extended with snippets.

One of the key things it can help with is inserting images, which is a common problem (incorrect paths), but it will also help with other custom items we commonly use in docs.

It persists media information that could help ensure alternate text is added with images.


tonykelly-octopus (approved):
LGTM

PR #2073, opened 2023-11-03
Fix async issue in synonyms

steve-fenton-octopus:
Async function wasn't awaited. This fixes the issue.


tonykelly-octopus (approved):
LGTM

PR #2074, opened 2023-11-06
Add link to K8s use case page

tonykelly-octopus:
Adding a link to the K8s use case page. I've added it to what is essentially the K8s overview page in Docs which I think gives us license to add a link to a 'marketing' page. 

This docs page is also the first Octopus page that appears in search when someone searches for deployments to kubernetes, so I want to try and speed up the new page being recognised by search engines and also pass some of the authority from the docs page to the use case page.


steve-fenton-octopus (approved):
LGTM

PR #2076, opened 2023-11-07
Fix linux tentacle pre-req link and systemd header

harrisonmeister:
I found a broken link and this PR fixes that and what looks like a erroneous update to the systemd header for linux tentacle


steve-fenton-octopus (approved):
LGTM!

PR #2087, opened 2023-11-20
Bonus score for shallow results

steve-fenton-octopus:
A small search scoring change to boost shallow results vs deeply nested results.

This may help to boost the landing pages for a search like "packages", where the Packages landing page is currently in position 4.

## Before

image


tonykelly-octopus (approved):
LGTM

PR #2092, opened 2023-11-23
Updated search

steve-fenton-octopus:
Changes the scoring system to surface less nested pages, as these are often the better results - especially for single-word searches, like "runbooks".

## Before

image

image

## After

image

image


tonykelly-octopus (approved):
LGTM

PR #2102, opened 2023-11-30
Version-controlled projects clone is not supported

abderrahmaneizri:


steve-fenton-octopus:
Hey @abderrahmaneizri 

Similar message as this branch will need an update, too.

You may want to update from `main` on this branch before we run the PR checks. Otherwise it will have the same issue as https://github.com/OctopusDeploy/docs/pull/2100

I've submitted a PR to your first patch to update from main, which should unblock that PR.

Thanks,
Steve

PR #2103, opened 2023-12-05
Remove server extensibility

tothegills:


steve-fenton-octopus:
Hey @tothegills 

Instead of deleting the files, they should be updated to redirect users to an appropriate location.

More info on this in the README here: https://github.com/OctopusDeploy/docs#redirects

- Tests: It looks like there are internal links to the deleted pages, so the tests caught the 404s caused by the deletions.
- Spellcheck: I can help resolve these once we've got passing tests and redirects.

Thanks,
Steve

steve-fenton-octopus (changes requested):
See the notes here: https://github.com/OctopusDeploy/docs/pull/2103#issuecomment-1840289471

Let me know if you need help with this.

PR #2108, opened 2023-12-06
Adding supporting video for forking repos

mcasperson:
This PR includes a few small line edits and links to a supporting video for forking CaC enabled projects.


robpearson (approved):
Looks good. I assume the variable references are correct as they're consistent and they appear to be fixes. 

Test the video links and they both work.

PR #2113, opened 2023-12-07
Add "Bento" as in "Project Bento" to the dictionary.

steve-fenton-octopus:
Allows the use of "Bento" as in "Project Bento" for project export/import.


tonykelly-octopus (approved):
LGTM

PR #2118, opened 2023-12-11
Remove deprecated server extensibility content

steve-fenton-octopus:
Replaces #2103 

Redirects the server extensibility pages to the deprecation notice from 2022.

Replicates the content removals from the original PR, but uses redirects not deletions.


steve-fenton-octopus:
@tothegills this implements the deprecations in your PR using redirects instead of deletions.

Please can you review to make sure it has been faithfully reproduced.

Thanks

tothegills (approved):
Thanks for putting this together @steve-fenton-octopus

PR #2120, opened 2023-12-12
Fix incorrect url

tmphealthtrack:


steve-fenton-octopus (approved):
Thanks for spotting and fixing this. LGTM!

PR #2123, opened 2023-12-13
Resolves spelling issue in API key placeholder

steve-fenton-octopus:
In files we update that flag the spelling for API-YOURAPIKEY we are using the less redundant API-YOUR-KEY


tonykelly-octopus (approved):
LGTM

PR #2124, opened 2023-12-14
Test short link bookmarks

steve-fenton-octopus:
This test runs through the list of short link bookmarks to ensure the fragments doesn't break.

For example, there is a shortlink for `/docs/insights/metrics#deployment-failure-rate`.

If the heading is changed or removed, there is a chance this fragment will be broken. The test will catch this.

More entries can be added to `bookmark.spec.ts` if more fragments become "crucial" (i.e. an in-product link points to them).


tonykelly-octopus (commented):
LGTM

tonykelly-octopus (approved):
LGTM

PR #2125, opened 2023-12-15
edited format of PATH line

FinnianDempsey:
Use of `>` adjusted to be code block instead 

Screenshot 2023-12-15 at 11 20 12 am


steve-fenton-octopus (approved):
LGTM!

PR #2127, opened 2023-12-18
Add warning message about known Helm upgrade limitations

kevjt:
[sc-66477]

This PR adds a warning message about a known limitation with the Helm `--atomic` flag and its interaction with Octopus's step cancellation mechanism. Octopus sends a SIGKILL to terminate processes when they are cancelled, and this results in the Helm process not being able to perform an automatic rollback. This can happen when either:

1. The user clicks on the cancel button during the deployment
2. The deployment times out (before the Helm `--timeout` limit is reached)

### DIFF
--- a/src/pages/docs/deployments/kubernetes/helm-update/index.md
+++ b/src/pages/docs/deployments/kubernetes/helm-update/index.md
@@ -18,7 +18,7 @@ A Helm Feed in Octopus refers to a [Helm Chart repository](https://helm.sh/docs/
 
 :::div{.info}
 
-The built-in repository is [capable of storing Helm Charts](/docs/packaging-applications/#supported-formats). However, the mechanism for determining the **PackageID** and **Version** may differ depending on the contents of the `.tgz` file.  If the `.tgz` file contains a `chart.yaml` file, the PackageID is determined by the `name`, and the version is determined by the `version` sections of the YAML.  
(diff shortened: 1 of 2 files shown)


src/pages/docs/deployments/kubernetes/helm-update/index.md
```diff
@@ -115,6 +115,10 @@ Helm deployments using Tar.gz packages can fail if the path is 100+ characters,
 See [https://github.com/OctopusDeploy/Issues/issues/8132](https://github.com/OctopusDeploy/Issues/issues/8132) for more info.  
 :::
 
+:::div{.warning}
+Due to how deployment cancellation currently works, the Helm `--atomic` argument does not result in automatic rollbacks when a deployment is cancelled. Furthermore, if the Octopus deployment timeout is set lower than the Helm timeout, a similar issue may arise. To ensure a smooth deployment experience, we recommend setting a larger Octopus timeout than the Helm timeout.
```
rain-on, line 119: nit: If `--atomic` doesn't result in automatic rollback - what does it do? (or is that obvious and I'm just not well versed?)

Also - if the deployment timeout is lower than helm-timeout, what _does_ happen? Or again, is that obvious to people in the know?

kevjt: Added in some more details

rain-on (approved):
Great update - love that it explains what can go wrong :+1:

PR #2130, opened 2023-12-20
Small improvements to OIDC docs based on recent feedback

geofflamrock:
This PR makes some small improvements to the docs for using OpenID Connect to access the Octopus API based on some recent feedback.
- Some users have been tripped up when setting the `id-token` permissions on a job: other existing implicit permissions now need to be explicitly set. We have a callout for this but have found that this can be skimmed over and users will reach for the example which didn't explain this well. The examples now have comments highlighting the explicitly required permissions.
- A callout has been added that OIDC is used for machine-to-machine connections, not for user authentication, with a link to the authentication providers docs. 
- A callout has been added that the subject on an OIDC identity doesn't support wildcards at the moment.


stevencl840 (approved):
Looks clear and concise

PR #2132, opened 2024-01-03
Handles chained terms in search

steve-fenton-octopus:
If you search for chained words, commonly found in code blocks, you'll expect to find results.

For example, `Octopus.Release.Channel.Name` is a chunk of code that appears on [Release versioning](https://octopus.com/docs/releases/release-versioning) page. You'd expect to find that page if you searched for "Octopus.Release.Channel.Name".

Pages found using chained search


tonykelly-octopus (approved):
LGTM

PR #2133, opened 2024-01-09
Resolves issue with tabs:

steve-fenton-octopus:
- Add closing tags
- Each group should have it's own id

Working tabs


tonykelly-octopus (approved):
LGTM

PR #2139, opened 2024-01-17
Removed some console logs

KMajkrzakOctopus:
@steve-fenton-octopus  correct me if I'm wrong but I don't think we need these console logs. I figured maybe it's better to get rid of those from production.

Before:
Screenshot 2024-01-17 at 15 13 56

After:
Screenshot 2024-01-17 at 15 14 19


steve-fenton-octopus:
Hey @KMajkrzakOctopus - I would make these changes in the Accelerator project, as they will then flow out to both Docs and DevOps.

PR #2141, opened 2024-01-25
updated description of where examples for package push live

benPearce1:
[sc-69469](https://app.shortcut.com/octopusdeploy/story/69469)

The **Show Examples** button has been removed and replaced with content in the help sidebar

image

### DIFF
--- a/dictionary-octopus.txt
+++ b/dictionary-octopus.txt
@@ -1,6 +1,5 @@
 ALLUSERSPROFILE
 apikey
-YOURAPIKEY
 approot
 appsettings
 ASPNET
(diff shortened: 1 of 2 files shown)


dictionary-octopus.txt
```diff
@@ -1,5 +1,6 @@
 ALLUSERSPROFILE
 apikey
+YOURAPIKEY
```
benPearce1, line 3: This is the sample api key used in the page that I changed, it is used in a few other places, but there are a number of other examples of placeholder api keys used

steve-fenton-octopus: We're trying to standardize on the also-used `YOUR-API-KEY` which is a little more readable.

benPearce1: I agree that it is more readable but it suggests a different format. Api Keys are always `API-XXXXX` with no additional hyphens.

IsaacCalligeros95 (approved):
LGTM

PR #2143, opened 2024-01-25
Update AWS OIDC docs

hnrkndrssn:
Try and make it easier for users to fall in the pit of success.

Updates based on feedback from a customer trying to get AWS OIDC setup

[sc-69507]

### DIFF
--- a/src/pages/docs/infrastructure/accounts/aws/index.md
+++ b/src/pages/docs/infrastructure/accounts/aws/index.md
@@ -60,7 +60,7 @@ When setting up the identity provider you need to use the host domain name of yo
     {
       "Effect": "Allow",
       "Principal": {
-          "Federated": "arn:aws:iam::{awsaccount}:oidc-provider/{youridentityprovider}"
+          "Federated": "arn:aws:iam::{aws-account}:oidc-provider/{your-identity-provider}"
       },
       "Action": "sts:AssumeRoleWithWebIdentity",
       "Condition": {
--- a/src/pages/docs/infrastructure/accounts/aws/index.md
+++ b/src/pages/docs/infrastructure/accounts/aws/index.md
@@ -33,18 +33,35 @@ See the [AWS documentation](https://oc.to/aws-access-keys) for instructions to c
 Support for OpenID Connect authentication to AWS requires Octopus Server version 2024.1
 :::
 
+To use OpenID Connect authentication you have to follow the [required minimum configuration](/docs/infrastructure/accounts/openid-connect#configuration).
+
 See the [AWS documentation](https://oc.to/aws-oidc) for instructions to configure an OpenID Connect identity provider.
+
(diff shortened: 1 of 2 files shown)


src/pages/docs/infrastructure/accounts/aws/index.md
```diff
@@ -33,18 +33,35 @@ See the [AWS documentation](https://oc.to/aws-access-keys) for instructions to c
...
-1. Provide a description for the account.
-1. Set the Role ARN to the ARN from the identity provider associated role.
-1. Set the Session Duration to the Maximum session duration from the role, in seconds.
-1. Click the **SAVE AND TEST** to save the account and verify the credentials are valid.
+2. Add a memorable name for the account.
+3. Provide a description for the account.
+4. Set the **Role ARN** to the ARN from the identity provider associated role.
+5. Set the **Session Duration** to the Maximum session duration from the role, in seconds.
+6. Click **SAVE** to save the account.
+7. Before you can test the account you need to add a condition to the identity provider in AWS:
```
IsaacCalligeros95, line 55: This is in the _IAM -> Roles -> Trust Relationships_ right? I think spelling that out here is worthwhile, people might just be looking under the Identityprovider

IsaacCalligeros95: Should we also include the Principal to better understand the whole trust relationship JSON? 

I'm also not sure if it's worth mentioning that they can avoid the subject requirement, from a security perspective probably better to leave that out but during testing I configured mine on just the audience and that was pretty helpful in my case.

`
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "Federated": "arn:aws:iam::720766170633:oidc-provider/a333-2001-8003-ac03-7500-6dfc-aa25-5c69-abec.ngrok-free.app"
            },
            "Action": "sts:AssumeRoleWithWebIdentity",
            "Condition": {
                "StringEquals": {
                    "a333-2001-8003-ac03-7500-6dfc-aa25-5c69-abec.ngrok-free.app:aud": "a333-2001-8003-ac03-7500-6dfc-aa25-5c69-abec.ngrok-free.app"
                }
            }
        }
    ]
}
`

IsaacCalligeros95 (commented):
A couple of minor comments, I'm happy to approve it as is if you don't think they're worth changing

src/pages/docs/infrastructure/accounts/aws/index.md
hnrkndrssn: > not sure if it's worth mentioning that they can avoid the subject requirement
I'd rather not, the chance that someone then forgets to update it to after testing and leaving their identity provider wide open is higher than I'd like compared to the convenience.

IsaacCalligeros95 (approved):
LGTM

PR #2144, opened 2024-01-29
Make docs a bit clearer on scheduled trigger logic

N-lson:
Relates to https://github.com/OctopusDeploy/Issues/issues/8525
[sc-66947]

### DIFF
--- a/src/pages/docs/projects/project-triggers/scheduled-deployment-trigger.md
+++ b/src/pages/docs/projects/project-triggers/scheduled-deployment-trigger.md
@@ -34,6 +34,10 @@ If you are using [tenants](/docs/tenants) you can select the tenants that will r
 
 5. Save the trigger.
 
+:::div{.hint}
+All schedule options run based on CRON expressions. The other options provide a convenient way of setting up the schedule without worrying about the syntax. A custom CRON expression provides you with more fine-grained control over the exact schedule.
+:::
+
 ### Using CRON expressions {#cron-expression}
 
 CRON expressions allow you to configure a trigger that will run according to the specific CRON expression.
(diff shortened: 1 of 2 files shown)


src/pages/docs/projects/project-triggers/scheduled-deployment-trigger.md
```diff
@@ -57,3 +57,7 @@ The CRON expression must consist of all 6 fields, there is an optional 7th field
 | Month         | 1-12 or JAN-DEC      | * , - /                     | Y        |
 | Day of week   | 0-6 or SUN-SAT       | * , - / ? L #               | Y        |
 | Year          | 0001–9999            | * , - /                     | N        |
+
+:::div{.hint}
```
shaun-od, line 61: I think this hint should be moved to the bottom of the Add a scheduled trigger section. It feels like it could be missed at the bottom of the cron section, and I think we want to make sure users are aware that this is the case even if they don't read the cron section.

src/pages/docs/runbooks/scheduled-runbook-trigger/index.md
```diff
@@ -60,3 +59,7 @@ The CRON expression must consist of all 6 fields, there is an optional 7th field
 | Month         | 1-12 or JAN-DEC      | * , - /                     | Y        |
 | Day of week   | 0-6 or SUN-SAT       | * , - / ? L #               | Y        |
 | Year          | 1970–2099            | * , - /                     | N        |
+
+:::div{.hint}
```
shaun-od, line 63: I would move this one as well

shaun-od (approved):
LGTM

PR #2147, opened 2024-01-31
Code block update for light/dark modes

steve-fenton-octopus:
## Before

Code blocks were always dark mode

### Light mode (default)

Light mode is lighter and on brand

### Dark mode

Not yet actively used, but in the pipeline. Hence the page itself is still in light mode.

Dark mode is available


tonykelly-octopus (approved):
LGTM

PR #2150, opened 2024-02-02
Fix grammatical error

KennethBates:
`in to obtain` -> `in order to obtain` :D


geofflamrock (approved):
Good catch, thanks

geofflamrock (approved):
Good catch, thanks

PR #2152, opened 2024-02-06
Update accelerator (resolves VITE issue)

steve-fenton-octopus:
Dependabot has [highlighted an issue in VITE](https://github.com/OctopusDeploy/docs/security/dependabot/10) that is resolved by updating the Accelerator dependency.


tonykelly-octopus (approved):
LGTM

PR #2163, opened 2024-02-14
Makes skip links visible on focus

steve-fenton-octopus:
When you use a keyboard, the skiplinks should be visible on focus.

1. Navigate to the page
2. Hit tab
3. The skip to... should appear

Skip links


KMajkrzakOctopus (approved):
👌

PR #2164, opened 2024-02-14
Fixed enter submit issue and some styling

KMajkrzakOctopus:
Fixed:
- z-index conflict on mobile
- No results smaller and with padding
- Tabbing into search caused losing focus on input (blur was called)
- Return btn cleared input (submit issue)


steve-fenton-octopus (approved):
LGTM

PR #2166, opened 2024-02-15
Search bug fixes

KMajkrzakOctopus:
Fixed reported issues:

- If you scroll to the bottom of the page, then start typing in search, the page scrolls up a bit on each keypress
- When typing in search and backspace everything (because you made a mistake) it won't let you type again unless you click inside the search bar again with your mouse
- When searching for e.g. Tenants and then clicking on a search result and then trying to search again won't populate new search results (no dropdown with results when searching a second time)

### DIFF
--- a/public/docs/js/search.js
+++ b/public/docs/js/search.js
@@ -1,22 +1,21 @@
 // @ts-check
 
-import { qs } from './modules/query.js';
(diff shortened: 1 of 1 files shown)


steve-fenton-octopus (approved):
LGTM

PR #2167, opened 2024-02-15
Resolving site_features issue

KMajkrzakOctopus:
If you select a page from search dropdown there is a chance this error will occur. Sometimes right away sometimes on a 3rd or 4th page selection. Can't reproduce this issue locally but happens on staging. Anyway I'm hoping this change will resolve it.

Screenshot 2024-02-15 at 17 20 26


steve-fenton-octopus (approved):
Let's try these out on staging...

PR #2170, opened 2024-02-19
Replace "outage window" with "maintenance window" for Octopus Cloud

x-cubed:
We want to move away from the term "outage window" as it implies that the instance is offline for the entire duration of the window every time.

Instead, we're using the term "maintenance window" to make it more obvious that it's an opportunity for us to perform maintenance on the instance.

[sc-59702]


steve-fenton-octopus:
Hi @x-cubed - is this change still required? If so, is there anything stopping this moving from draft to "proper PR"?

x-cubed:
> Hi @x-cubed - is this change still required? If so, is there anything stopping this moving from draft to "proper PR"?

Sorry @steve-fenton-octopus, just trying to coordinate a few different PRs across different repos in the same ticket.
Currently waiting on a review from the Product team on a larger document, then we'll try to land all the PRs around the same time.

PR #2171, opened 2024-02-20
Update RHEL supported version

zentron:
As noted in [internal slack thread](https://octopusdeploy.slack.com/archives/C033W4273/p1708385768287979) we have


steve-fenton-octopus (approved):
LGTM. I've updated the modified date.

PR #2174, opened 2024-02-22
Image component

KMajkrzakOctopus:
This PR is about a new image component. Although it's not being used anywhere for now - each image has to be manually replaced with this new component. Due to that fact for now I added some styles for current images so they'll look more or less as the component. This is how they'll look like now:
Screenshot 2024-02-22 at 10 20 05

This is the image component:
1. Default

https://github.com/OctopusDeploy/docs/assets/148771144/b605b5ae-ec02-4da2-b2cf-e5410a4b2300

2. With caption

https://github.com/OctopusDeploy/docs/assets/148771144/d9e82d36-a40a-4339-9733-7f39b97124aa

### DIFF
--- a/public/docs/css/main.css
+++ b/public/docs/css/main.css
@@ -177,6 +177,56 @@ img {
   background-color: var(--white);
 }
 
+/* Image component */
+.image {
+  border-radius: 0.9375rem;
+  border: 0.0625 solid var(--navy-200);
+  background: #F4F6F8;
+  display: flex;
+  flex-direction: column;
+  gap: 1.56rem;
+  padding: 2rem;
+  position: relative;
+}
+
+@media (max-width: 930px) {
+  .image {
+    padding: 1rem;
+  }
+}
+
+.image .magnify-container {
+  position: absolute;
+  right: 0;
+}
+
+.image__img {
+  border-radius: 0.6875rem;
+  box-shadow: 0px 4px 20px 0px rgba(0, 0, 0, 0.10);
+}
+
+.image__caption {
+  color: var(--blue-grey-dark);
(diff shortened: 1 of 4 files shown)


public/docs/css/main.css
```diff
@@ -177,6 +177,56 @@ img {
...
+}
+
+.image__caption {
+  color: var(--blue-grey-dark);
+  font-size: 1.125rem;
+  font-weight: 400;
+}
+
+/* TODO: Once all images are used with Image component these styles should be removed */
+figure > p > img {
```
KMajkrzakOctopus, line 215: Styling by tags like this isn't good idea but based on `figures.js` it should work for now

public/docs/js/modules/figures.js
```diff
@@ -1,28 +1,28 @@
...
 
-/** 
+/**
  * Enables opening image in new tab
-*/
+ */
 function enhanceFigures() {
-    qsa(`figure > p > img`).forEach(node => {
-        const src = node.src;
+  qsa(`figure > p > img, [data-image] > .image__img`).forEach((node) => {
```
KMajkrzakOctopus, line 9: I'll update it in the accelerator 🙇

src/components/Image.astro
```diff
@@ -0,0 +1,14 @@
+---
+// Define the component's properties
+type Props = {
+  imgSrc: string;
```
steve-fenton-octopus, line 4: For the props, do we need to prefix with `img` as the component is called `Image`.

i.e. `<Image src="" alt="" caption="" />` makes sense instead of `<Image imgSrc="" imgAlt="" caption=""" />`

PR #2175, opened 2024-02-22
Card component

KMajkrzakOctopus:
Card component created with two variants:
1. Default

https://github.com/OctopusDeploy/docs/assets/148771144/480bde2b-f544-4864-8126-c9b0eefc56d8

2. Padded

https://github.com/OctopusDeploy/docs/assets/148771144/8d0ec854-1265-4c61-84e5-f066094109ce

### DIFF
--- a/src/components/Card.astro
+++ b/src/components/Card.astro
@@ -6,21 +6,15 @@ type Props = {
   imgSrc: string;
   link: string;
   title: string;
-  variant?: "" | "padded";
+  variant?: "padded";
 };
-const {
-  description,
-  imgAlt,
-  imgSrc,
-  link,
-  title,
-  variant = "",
-} = Astro.props satisfies Props;
+const { description, imgAlt, imgSrc, link, title, variant } =
+  Astro.props satisfies Props;
 
 const classList = `card ${variant === "padded" ? "card--padded" : ""}`;
 ---
 
-<article class={ classList }>
+<article class={classList}>
   <a href={link} class="card__link">
     <div class="card__image-wrapper">
       <img src={imgSrc} alt={imgAlt} class="card__image" />
--- a/src/components/Card.astro
+++ b/src/components/Card.astro
@@ -1,0 +1,31 @@
+---
+// Properties
+type Props = {
+  description: string;
+  imgAlt: string;
+  imgSrc: string;
+  link: string;
+  title: string;
+  variant?: "default" | "padded";
+};
+const {
+  description,
+  imgAlt,
+  imgSrc,
+  link,
+  title,
+  variant = "default",
+} = Astro.props satisfies Props;
+---
+
+<article class={`card ${variant === "padded" ? "card--padded" : ""}`}>
+  <a href={link} class="card__link">
+    <div class="card__image-wrapper">
+      <img src={imgSrc} alt={imgAlt} class="card__image" />
+    </div>
+    <div class="card__copy">
+      <p class="card__title">{title}</p>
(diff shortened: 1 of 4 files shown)


src/components/Card.astro
```diff
@@ -0,0 +1,31 @@
...
+  title: string;
+  variant?: "default" | "padded";
+};
+const {
+  description,
+  imgAlt,
+  imgSrc,
+  link,
+  title,
+  variant = "default",
```
steve-fenton-octopus, line 17: Make the default "", rather than a magic word.

Process what they pass in the code, rather than inline in the HTML.

With Astro, it's best to keep the HTML part as plain as possible, so try to do any logic before you get there if you can.

src/components/Card.astro
```diff
@@ -0,0 +1,31 @@
...
+} = Astro.props satisfies Props;
+---
+
+<article class={`card ${variant === "padded" ? "card--padded" : ""}`}>
```
steve-fenton-octopus, line 21: Then on this line, you could just have `class={cssClass}`

steve-fenton-octopus (changes requested):
I think we can make this a little cleaner with a couple of changes. See the comments.

src/components/Card.astro
KMajkrzakOctopus: I agree with default value as an empty string but I'm not sure if I follow this one.
We don't need a prop `class` or `cssClass` that might not be used - if that's going to be needed we can extend the component for that case (YAGNI). I'd much rather have a predefined variant that will control the look/functionality. Therefore even with default changed to empty string this logic would remain for now.
`<article class={`card ${variant === "padded" ? "card--padded" : ""}`}>`

In case of having more variants in the future I'd go for doing something like this:
```
type Props = {
  description: string;
  imgAlt: string;
  imgSrc: string;
  link: string;
  title: string;
  variant?: "" | "padded" | "outlined";
};

const variantClasses = {
  default: "",
  padded: "card--padded",
  outlined: "card--outlined",
};

const variantClass = variantClasses[variant];
---

<article class={`card ${variantClass}`}>
```

steve-fenton-octopus: Let me send a commit to help explain what I mean...

KMajkrzakOctopus: Btw come to think of it we don't even need a default or empty string for it. It can be simply 
`variant?: "padded";`
for now since variant isn't a required prop. When it's used it'll be only for `padded` or any other varian in the future

steve-fenton-octopus:
I've added a commit to pull the logic from the HTML section into the code block.

steve-fenton-octopus (approved):

KMajkrzakOctopus:
> I've added a commit to pull the logic from the HTML section into the code block.
> 
> [d7e0425](https://github.com/OctopusDeploy/docs/commit/d7e042583a69234515b95299d5c817f79811727a)

Oh yeah sure it can be like that 😅  I thought you wanted another prop for custom css classes.
Also @steve-fenton-octopus on top of that what do you think about changing this like that:
```
  variant?: "padded";
};
const {
  description,
  imgAlt,
  imgSrc,
  link,
  title,
  variant,
} = Astro.props satisfies Props;
```

We don't need it really. Default state will be without even passing a variant prop.

steve-fenton-octopus:
Yes - as the prop is optional, it makes sense to simplify the union type by removing "".It gets the default value assigned in the next block if it's null.

👍

PR #2176, opened 2024-02-22
Image typo fix

KMajkrzakOctopus:
Border was without unit `rem` I missed that somehow so that's a quick fix to that.


steve-fenton-octopus (approved):
LGTM

PR #2180, opened 2024-02-23
Update CODEOWNERS

tonykelly-octopus:
Adding final round of codeowners based upon Ownership Map. I'll update any if / when we get pushback.


steve-fenton-octopus (approved):
LGTM

PR #2186, opened 2024-02-29
Disconnect accelerator

steve-fenton-octopus:
Moves files into docs for local use.


KMajkrzakOctopus (approved):
Great work 🙇  thanks for taking care of that!

PR #2192, opened 2024-03-01
Prettier implementation

KMajkrzakOctopus:


steve-fenton-octopus:
Hey @KMajkrzakOctopus 

I don't think you started from an up to date `main`...

image

You'll need to be careful not to add back in files that we moved and deleted in the "detach accelerator" change last week.

Whenever you make a branch, it's always worth making sure you've done a `sync` on `main` first - the less merging you have to do, the better :)

steve-fenton-octopus (approved):
LGTM!

PR #2199, opened 2024-03-05
Adding 'Deprecated' tag to title of Octo CLI

tonykelly-octopus:
Adding deprecation notice to appear in page title and side navigation.


steve-fenton-octopus (approved):
LGTM

PR #2212, opened 2024-03-18
Update config-as-code-reference.md

zentron:
Include note indicating missing slugs will result in an error Relates to https://app.shortcut.com/octopusdeploy/story/61036/cac-target-deletion-does-not-cascade-to-variable-scope


steve-fenton-octopus (approved):
LGTM!

PR #2249, opened 2024-04-22
Resolve spelling issues

steve-fenton-octopus:
Updates legacy "super long fragments" and updates references.


APErebus:
@steve-fenton-octopus do you need me to check short.io and the product to see if this breaks anything?

This will obviously hold up the k8s agent documentation that we need to get by CoB on Monday 30th, so just want to know what the path forward is.

steve-fenton-octopus:
> @steve-fenton-octopus do you need me to check short.io and the product to see if this breaks anything?
> 
> This will obviously hold up the k8s agent documentation that we need to get by CoB on Monday 30th, so just want to know what the path forward is.

@APErebus I check the fragments in short.io when I update them. I was told the product uses short.io links for everything. If it doesn't, it would be worth adding them to the fragment tests to make sure they don't get broken.

PR #2254, opened 2024-04-23
Remove section on exporting data from Octopus Cloud

goosmurf:
We don't currently have a good way to comprehensively export data from a Cloud instance.

For Octonauts, this is being discussed internally [here](https://octopusdeploy.slack.com/archives/C01EG659GHK/p1713855979631789).

[sc-76512]


steve-fenton-octopus (approved):
LGTM!

PR #2256, opened 2024-04-24
Explicitly list the steps that support Server-Side Apply

kevjt:
We want to be more specific than just saying "Octopus supports Server-Side Apply for all of our Kubernetes steps" because some people might expect to see it in the Helm step or kubectl script step which don't have it.

Screenshot 2024-04-24 at 4 31 53 pm


kevjt:
I considered making each item a hyperlink to the respective doc page, but decided against it because not every step has a page in the docs. e.g. there's no page for _Configure and apply a Kubernetes Secret_ 🤷

steve-fenton-octopus (approved):
LGTM!

PR #2258, opened 2024-04-26
Dark mode

KMajkrzakOctopus:
https://github.com/OctopusDeploy/docs/assets/148771144/991f1a80-824b-421a-aa65-51e401569e71


src/layouts/Default.astro
```diff
@@ -65,7 +65,7 @@ const showSearch = !isSearchPage;
 
 <html dir={textDirection} lang={lang} class="initial">
   <Head frontmatter={frontmatter} headings={headings} lang={lang} />
-  <body>
+  <body data-theme="dark">
```
steve-fenton-octopus, line 68: Are we intentionally going for dark theme by default?

src/layouts/Default.astro
```diff
@@ -74,7 +74,7 @@ const showSearch = !isSearchPage;
       showSearch={showSearch}
     />
     <div class="content-group">
-      <main id="site-main">
+      <main id="site-main test">
```
steve-fenton-octopus, line 77: Is the "test" class supposed to still be here or was it just for a test?

PR #2259, opened 2024-04-26
Update link on new K8s page

harrisonmeister:


steve-fenton-octopus (approved):
Thank you - good spot!

PR #2261, opened 2024-04-26
Updating list of database products and providing links

twerthi:
Updated the list of products and reordered them, placing the our Redgate partners at the top.


steve-fenton-octopus (approved):
LGTM!

PR #2268, opened 2024-04-30
MaxParallelism takes precedence over Window size

tothegills:
Closes https://github.com/OctopusDeploy/Issues/issues/8170


steve-fenton-octopus (approved):
LGTM!

PR #2271, opened 2024-04-30
Removing import instruction

tonykelly-octopus:
See screenshot - code snippet is displaying on prod. The content from the imported segment has already been added to the page directly so it is no longer needed anyway.

image


steve-fenton-octopus:
@tonykelly-octopus just FYI imports work with `.mdx` files, so if an import is needed, the file would need a rename. The tips on [working with mdx files](https://github.com/OctopusDeploy/docs?tab=readme-ov-file#mdx-file-differences) would need to be followed when a file is changed from `.md` to `.mdx`.

And also... LGTM.

steve-fenton-octopus (approved):

PR #2272, opened 2024-05-01
page not found edits

TeganAli:


TeganAli:
Just want to remove some repetition, remove the assumption people have already visited the Octopus home page, and make the difference between the docs home page and Octopus home page clearer.

steve-fenton-octopus (approved):
LGTM!

PR #2281, opened 2024-05-07
Add Kubernetes Agent OpenShift storage warning

APErebus:
OpenShift does not support privileged containers by default and when enabled, there are issues with file access. Adding a warning to the storage page indicating people should use a custom storage class.

Shortcut story: [sc-77491]

image

### DIFF
--- a/src/pages/docs/infrastructure/deployment-targets/kubernetes/kubernetes-agent/storage.md
+++ b/src/pages/docs/infrastructure/deployment-targets/kubernetes/kubernetes-agent/storage.md
@@ -32,7 +32,7 @@ This default implementation is made to let you try the Kubernetes agent without 
 The NFS server requires `privileged` access when running as a container, which may not be permitted depending on the cluster configuration. Access to the NFS pod should be kept to a minimum since it enables access to the host. 
 
 :::div{.warning}
-Red Hat OpenShift does not enable `privileged` access by default. When enabled, we have also encountered inconsistent file access issues using the NFS server. We highly recommend the use of a [custom storage class](#custom-storage-class).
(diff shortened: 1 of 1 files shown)


src/pages/docs/infrastructure/deployment-targets/kubernetes/kubernetes-agent/storage.md
```diff
@@ -31,12 +31,16 @@ This default implementation is made to let you try the Kubernetes agent without
...
+Red Hat OpenShift does not enable `privileged` access by default. When enabled, we have also encountered inconsistent file access issues using the NFS server. We highly recommend the use of a [custom storage class](#custom-storage-class).
+:::
+
 ### Reliability
 Since the NFS server runs inside your Kubernetes cluster, upgrades and other cluster operations can cause the NFS server to restart. Due to how NFS stores and allows access to shared data, script pods will not be able to recover cleanly from an NFS server restart. This causes running deployments to fail when the NFS server is restarted.
 
 If you have a use case that can’t tolerate occasional deployment failures, it’s recommended to provide your own `StorageClass` instead of using the default NFS implementation.
 
-## Custom StorageClass
+## Custom StorageClass \{#custom-storage-class}
```
steve-fenton-octopus, line 43: If this bookmark is important (for example, you use it in a blog or in-product help) we can add it to the tests. We set up a list based on URLs in our short-link tools, but we can add more.

The test makes sure the bookmark doesn't get broken by future edits:

https://github.com/OctopusDeploy/docs/blob/main/tests/bookmark.spec.ts

APErebus: @steve-fenton-octopus Thanks for the info. It's not an important bookmark outside of the docs. I only introduced it so I could link in this page

steve-fenton-octopus (approved):
LGTM!

flin-8 (approved):
LGTM

PR #2286, opened 2024-05-08
Hide the page so only people with the url can read the docs

mcasperson:
These settings essentially make a docs page accessible by URL only:

```
navSearch: false
navSitemap: false
navMenu: false
robots: noindex, follow
```


harrisonmeister (approved):
LGTM

PR #2287, opened 2024-05-08
Add extensions configuration to help onboarding

steve-fenton-octopus:
Suggests the right list of extensions to users of VS Code / GitHub codespaces.


tonykelly-octopus (approved):
LGTM

PR #2291, opened 2024-05-10
Update the Copilot App name

mcasperson:
Based on feedback from the Github Team, the app name of the Copilot agent needs to be updated.


robpearson (approved):
👌

PR #2297, opened 2024-05-13
Daniel tentacle requirements update

danielfischeroctopus:


zentron (approved):
Thanks @danielfischeroctopus

zentron (changes requested):
@danielfischeroctopus are we able to be more specific about which LTSC we support?

zentron (approved):
Thanks for the update @danielfischeroctopus

steve-fenton-octopus (approved):
LGTM!

PR #2301, opened 2024-05-14
Document Kubernetes agent automatic updating

APErebus:
Update the updating section of the Kubernetes agent with Automatic updating information. ALso updated the machine policies page with kubernetes agent information

image

image


steve-fenton-octopus (approved):
Nice! Great feature.

PR #2302, opened 2024-05-14
updated step conditions to better reflect max parallelism

FinnianDempsey:
@steve-fenton-octopus I'm not too sure what's going on with the package.json file being updated with pnpm 9.1.0, I didn't touch that file

Changes:
Screenshot 2024-05-14 at 13 50 27

More than 10 steps can execute in parallel by default:
Screenshot 2024-05-14 at 13 30 11

### DIFF
--- a/src/pages/docs/projects/steps/conditions/index.mdx
+++ b/src/pages/docs/projects/steps/conditions/index.mdx
@@ -1,7 +1,7 @@
 ---
 layout: src/layouts/Default.astro
 pubDate: 2023-01-01
-modDate: 2023-01-01
+modDate: 2023-05-14
 title: Conditions
 description: Adding conditions to steps to define your project's deployment process.
 navOrder: 20
--- a/src/pages/docs/projects/steps/conditions/index.mdx
+++ b/src/pages/docs/projects/steps/conditions/index.mdx
@@ -108,16 +108,10 @@ When you review a process with two steps that run in parallel, you'll notice two
 
 ### Maximum parallelism
 
-To help your Octopus Server remain stable, we have defaulted the maximum number of steps you can run concurrently in any given deployment to `10`. You can configure any number of steps to run in parallel, and Octopus will roll through them as quickly as possible, but it will only run `10` of them at any one time by default.
+There is no limitation on the number of steps that can run in parallel. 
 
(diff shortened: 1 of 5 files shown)


src/pages/docs/projects/steps/conditions/index.mdx
```diff
@@ -108,16 +108,10 @@ When you review a process with two steps that run in parallel, you'll notice two
 
 ### Maximum parallelism
 
-To help your Octopus Server remain stable, we have defaulted the maximum number of steps you can run concurrently in any given deployment to `10`. You can configure any number of steps to run in parallel, and Octopus will roll through them as quickly as possible, but it will only run `10` of them at any one time by default.
+There is no limitation on the number of steps that can run in parallel. 
 
-#### Overriding maximum parallelism
+The `Octopus.Action.MaxParallelism` variable allows for specifying the number of machines on which the action will concurrently execute which is the same variable used for [Rolling Deployments](https://octopus.com/docs/deployments/patterns/rolling-deployments).
```
gb-8, line 113: Is it worth making it clear that the default parallelism of 10 still applies for the number of targets a given step will execute on?

Eg.:


```
When a step executes on multiple deployment targets, it will execute on up to 10 targets in parallel by default.
You can change this number by setting the `Octopus.Action.MaxParallelism` variable. This variable also controls [Rolling Deployments](https://octopus.com/docs/deployments/patterns/rolling-deployments)
```

gb-8 (approved):
Approved.

The new description matches the current behaviour from my reading of the code.

I left a comment about whether we still want to explain the default parallelism across targets.

steve-fenton-octopus:
@FinnianDempsey I fixed the lock file and fired it into the branch - the actions are running okay now.

src/pages/docs/projects/steps/conditions/index.mdx
FinnianDempsey: this is called out other places in the docs but happy to mention it here too but was trying to avoid including a number there for the skim readers

https://octopus.com/docs/projects/variables/system-variables#user-modifiable-settings

gb-8: Cool. It was just a drive-by opinion.

PR #2303, opened 2024-05-14
Fix invalid project casing in "Octopus Server Container with Docker Compose" guide

MissedTheMark:
The example project name in the guide is `Octopus` (used as `docker compose --project-name Octopus up -d`), except docker throws an error, because, quoting the Docker docs, "Project names must contain only lowercase letters, decimal digits, dashes, and underscores, and must begin with a lowercase letter or decimal digit."

See docs here: https://docs.docker.com/compose/project-name/#set-a-project-name 

Before:
image

After:


steve-fenton-octopus (approved):
Good catch! Thanks.

PR #2306, opened 2024-05-17
Update getting-started-with-cloud.mdx

irynamelnyk-octopus:
Update the Set the maintenance window section.


steve-fenton-octopus (approved):
I checked the solution and couldn't find any backlinks to this bookmark.

It's also not in the tests for backlinks in short.io

That means we're okay to change it - in cases where someone has this linked from some other source, they'll still end up on the page and can find this information in the table of contents.

PR #2310, opened 2024-05-21
UI batch

KMajkrzakOctopus:
Before:

https://github.com/OctopusDeploy/docs/assets/148771144/a76b6f4d-eaa5-4ec2-92f9-183e715f918e

After:

https://github.com/OctopusDeploy/docs/assets/148771144/454a7b61-007e-44ab-8c0d-05a75c417603

New listing page for article available components:
`/components`

https://github.com/OctopusDeploy/docs/assets/148771144/156799ee-e352-4f3a-95f4-c9672aabfcd1


public/docs/css/main.css
```diff
@@ -29,6 +29,9 @@ body {
   line-height: 1.6;
   color: var(--color-text);
   background-color: var(--color-base-primary);
+  transition-property: background-color, color;
```
steve-fenton-octopus, line 32: Users sometimes switch of animations for various perceptual conditions. We have a block elsewhere that uses the "prefers reduce motion" flag - it would be great to make sure we use it wherever we animate, so users who have issues with animation can avoid it.

steve-fenton-octopus (approved):
LGTM!

PR #2316, opened 2024-05-24
Faster CI feedback

steve-fenton-octopus:
Short circuits in-flight actions when a new commit arrives. Fails fast. Caches pnpm.


tonykelly-octopus (approved):
LGTM

PR #2317, opened 2024-05-24
Fixing a typo in the modification date field

steve-fenton-octopus:
The modification date was before the publishing date.


tonykelly-octopus (approved):
LGTM

PR #2322, opened 2024-05-30
Add troubleshooting info for when the Agent install command times out

scme0:
# Background

One main way the agent installation can fail is with a timeout. This can be caused by a number of different things but there are a few things that are likely to be common causes.

# Results

I've updated the documentation to list several common causes to help users identify the problem in their case. I've also added another command that gets the logs for the tentacle pod as this will most likely help them to identify the issue.


steve-fenton-octopus (approved):
LGTM!

PR #2326, opened 2024-06-03
Updating Variables page and supporting links

tonykelly-octopus:
- Updating Variables overview page
- Adding Getting started page and moving content from Variables overview page to here
- Updating bookmark links to point at new content on the Getting started page


steve-fenton-octopus (approved):
LGTM!

PR #2327, opened 2024-06-04
Moving the variable into the AWS library variable set.

steve-fenton-octopus:
Moving this into the library set, so we can use it from all projects.


tonykelly-octopus (approved):
LGTM

PR #2333, opened 2024-06-10
Just warn for PNPM mismatch

steve-fenton-octopus:
Instead of faulting out, just warn users when their PNPM version is incorrect. This is especially useful if you're ahead of the specified version.


tonykelly-octopus (approved):
LGTM

PR #2337, opened 2024-06-11
Docs update based on Handbook and Integrations changes

KMajkrzakOctopus:


src/components/Card.astro
```diff
@@ -36,7 +36,13 @@ const labelText = isYouTubeLink ? 'Watch Video' : 'Learn More';
 <article class={classList}>
   <a href={link} class="card__link" role="link" aria-label={`Open ${title}`}>
     <div class="card__image-wrapper">
-      <img src={imgSrc} alt={imgAlt} class="card__image" />
+      <img
+        src={imgSrc}
+        alt={imgAlt}
+        class="card__image"
+        onerror="this.src='https://i.octopus.com/library/step-templates/other.png';"
```
steve-fenton-octopus, line 43: Do we definitely want this fallback in docs?

KMajkrzakOctopus: Yeah I was thinking the same - most likely we're not going to use that, but on the other hand it's not going to cause any harm and it's easier to maintain shared components once they are exact copies 🤔

steve-fenton-octopus (approved):
LGTM!

PR #2344, opened 2024-06-14
Added hidden docs for migrating spaces with octoterra

mcasperson:
This PR adds documentation that describes how to migrate a space and projects, including CaC projects, with Octoterra.

### DIFF
--- a/src/pages/docs/administration/migrate-spaces-with-octoterra/index.md
+++ b/src/pages/docs/administration/migrate-spaces-with-octoterra/index.md
@@ -12,20 +12,22 @@ navMenu: false
 robots: noindex, follow
 ---
 
-Octoterra is a tool that exports Octopus projects, runbooks, and spaces to a Terraform module. Octoterra can be used to migrate resources between spaces and instances. However, there are limitations that must be accounted for as part of a migration. This documentation outlines a process for migrating projects from one space to another, highlighting the limitations and workaround of the octoterra tool.
+[Octoterra](github.com/OctopusSolutionsEngineering/OctopusTerraformExport/actions) exports Octopus projects, runbooks, and spaces to a Terraform module. Octoterra can be used to migrate resources between spaces and instances.
 
 ## Limitations of octoterra and migrating projects between instances
 
+There are limitations that must be accounted for as part of a migration.
+
 ### Sensitive values
 
-Octoterra inspects the state of a space via the Octopus API and the API does not expose sensitive values. This means that octoterra can not export:
+Octoterra reads the state of a space via the Octopus API and the API does not expose sensitive values. This means that octoterra can not export:
 
 * The value of sensitive variables associated with a project, tenant, or library variable set 
 * Credentials defined in feeds, accounts, or git credentials
 * The contents of a certificate
 * Sensitive values defined in steps
 
-Octoterra defines Terraform variables to allow the value of these fields to be defined when the module is applied.
+Octoterra creates Terraform variables to allow the value of these fields to be defined when the module is applied.
 
 ### Step templates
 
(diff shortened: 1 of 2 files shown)


src/pages/docs/administration/migrate-spaces-with-octoterra/index.md
```diff
@@ -0,0 +1,309 @@
...
+
+### Step templates
+
+The Octopus Terraform provider does not yet support defining step templates. Because step templates are assigned a unique ID in each space, it is not possible to export a project or runbook that references a step template across spaces.
+
+As a workaround, octoterra can detach step templates while exporting a project or runbook. Projects or runbooks with detached step templates can be recreated in a new space.
+
+### New step framework
+
+Some steps, such as the ECS deployment steps, rely on a new framework. Steps that use the new framework are not currently supported by the Terraform provider. These steps can not be exported by octoterra.
```
robpearson, line 40: I think it's worth mentioning which steps this includes as our customers wouldn't know and our team might not know. AFAIK ... 
- Amazon ECS Steps
- Bicep step 
- S3 step

robpearson: Note: I've confirmed the list in Slack.

src/pages/docs/administration/migrate-spaces-with-octoterra/index.md
```diff
@@ -0,0 +1,309 @@
...
+  -dummySecretVariableValues \
+  -dest ~/Desktop/project
+```
+
+The values of sensitive variables must be manually updated after the module is applied.
+:::
+
+## Recreating Config-as-Code projects
+
+CaC projects can not share a Git repository. When a CaC project is recreated, it must point to a new Git repository. This is done by overriding the default value of a Terraform variable in the exported module.
```
robpearson, line 246: It might be handy to mention that if they're migrating from a self-hosted instance to Octopus Cloud that this doesn't apply.

src/pages/docs/administration/migrate-spaces-with-octoterra/index.md
```diff
@@ -0,0 +1,309 @@
...
+variable "project_my_project_git_url" {
+  type        = string
+  nullable    = false
+  sensitive   = false
+  description = "The git url for \"My Project\""
+  default     = "https://github.com/organization/my_project.git"
+}
+```
+
+The default value of this variable is the URL of the Git repo configured for the source project. To override this value when applying the module, pass the argument `-var=project_my_project_git_url=https://github.com/organization/new_repo_name.git`, for example:
```
robpearson, line 291: Following up on my last comment, is this still required? I have hit problems in the past when I added an existing demo repo to a new Octopus instance. i.e., it complains that the repo is already initialised. My workaround is to remove the .octopus files, commit, initialise the Octopus CaC project, and then copy the original files back. 

Could this workaround work here? I know it sucks but would it still work?

robpearson: This might be confusing so please ping me if it doesn't make sense.

robpearson (approved):
I think this is great Matt. Added a few minor comments. 

I was speaking to Adam about the issues he was having yesterday so if we can get this to work, it's a much better option than any of the alternatives. i.e. It can be consistent and repeatable. We could add docs on how to migrate the exceptions as well. i.e. Use the Space Cloner for the additional resources (Teams etc) and draft a manual process to move projects that use new "step framework" steps.

mcasperson:
@robpearson This is good feedback. The automerge went ahead and applied these changes, but I'll look at doing a second PR with your suggestions.

PR #2351, opened 2024-06-19
Use new templated deployment process

steve-fenton-octopus:
The templated deployment will be shared by all microsites.


tonykelly-octopus (approved):
LGTM

PR #2370, opened 2024-07-01
Audit and fix jFrog links

steve-fenton-octopus:
I noticed the jFrog links were a mixture of www and non-www, so audited them and found they all redirected to new pages. Links updated to follow the redirect.

Two links were broken and have no alternative, but the content is covered in the other link in the same section.


tonykelly-octopus (approved):
LGTM

PR #2377, opened 2024-07-09
Add gitlab auth to know docker limitatons

tothegills:
See https://github.com/OctopusDeploy/Issues/issues/8156

I've collapsed the "problem" dialogue into the known limitations sections. It doesn't appear to require a callout, I don't know what customers can do about it.


steve-fenton-octopus (approved):
This looks like a good way to organize the information.

domenicsim1 (approved):

PR #2388, opened 2024-07-22
fix: Replacing using the appropriate value

dix:
The doc says that if we replace something with the same thing, the output is different than the output?

To obtain the given output, the replacement value should be "\\\\" (_replace with two escaped slashes_).

Tested and validated.


steve-fenton-octopus:
Hey @dix thank you for spotting this and submitting a fix.

steve-fenton-octopus (approved):
LGTM!

PR #2389, opened 2024-07-25
Update troubleshooting-tentacles.md

garrettdass:
Added note in 'Other error messages' section about SHA1 Octopus server certificates causing communication issues with modern Linux distros.


APErebus (approved):
Looks good to me! You'll just need to add `Wireshark` to the approved words list as it's causing the spell check to fail

steve-fenton-octopus (approved):
LGTM!

PR #2391, opened 2024-07-26
Docs overview pages updated to use new product UI

tonykelly-octopus:
- Updated all overview pages with screenshots to use the UI with the side navigation


steve-fenton-octopus (approved):
LGTM!

PR #2395, opened 2024-07-30
Shift around Kubernetes content to dedicated section

thekatstevens:
**TLDR:** we aimed to unify the disparate Kubernetes documentation so it can more easily be found, and give more prominent signal of our capabilities here to users browsing the docs. Mostly these are structural changes, not page content changes.

It's not perfect, but merging this WIP will prevent it drifting further :P and give the chance for feedback from staff and users about the direction.


steve-fenton-octopus:
Pulling in the latest from main to fix the "Wireshark" spelling error.

steve-fenton-octopus (approved):
I think we're good to go.

Feel free to review the changes on staging before merging if you want to run a final check.

I've added content to the steps overview page and updated modified dates.

http://octopus-docs-staging.s3-website-us-east-1.amazonaws.com/docs/kubernetes/

PR #2400, opened 2024-08-02
Update Linux versions tested against

rhysparry:
Updates the list of Linux distributions we run Tentacle integration tests against.

The MacOS version was left unchanged as the version could not be determined. However, we also only run unit tests in this environment.

[SC-85714]


steve-fenton-octopus (approved):
LGTM!

PR #2403, opened 2024-08-05
Correct information about DW IP addresses.

Justin-Walsh:
Update our Static IP docs to use the correct terminology. We don't use the built-in worker in Octopus Cloud at all.


steve-fenton-octopus (approved):
LGTM!

PR #2406, opened 2024-08-08
Add Kubernetes agent diagrams

nikita-dergilev:
Added Rob's diagrams to the Kubernetes target overview and Kubernetes agent overview pages


APErebus:
Do we need to manually update the `modDate` on the docs files?

nikita-dergilev:
> Do we need to manually update the `modDate` on the docs files?

@APErebus , to be honest, I don't know, maybe. @thekatstevens, do you know the answer?

zentron:
I think this looks pretty good. I'm not sure if we will need more context around the components diagram, but what you have might be sufficient for now.

zentron (approved):
I think this looks pretty good. I'm not sure if we will need more context around the components diagram, but what you have might be sufficient for now.

thekatstevens (approved):
Looks fine to me, if the images load properly when you run it locally then 👍 

I'll tweak a couple of links while we're editing too

steve-fenton-octopus:
@APErebus @nikita-dergilev If you make a solid change to the content, it's worth updating the `modDate`.

We use this to signal to Google that there's a change worth re-indexing, which is why we don't update it automatically. If you signal low-change or no-change updates, they stop listening to your signal.

nikita-dergilev:
@steve-fenton-octopus , @APErebus , @thekatstevens , @zentron , thanks a lot! Merged :-)

PR #2411, opened 2024-08-13
Octoterra Wizard migration docs

mcasperson:
This PR rewrites the octoterra migration docs to document the Octoterra Wizard.


steve-fenton-octopus:
Are we ready to put in real dates here?

```
pubDate: 2099-01-01
modDate: 2099-01-01
```

steve-fenton-octopus (approved):
I've left a note about the publish/modified dates. Otherwise looks good to me.

mcasperson:
> Are we ready to put in real dates here?
> 
> ```
> pubDate: 2099-01-01
> modDate: 2099-01-01
> ```

These docs will be hidden for some time yet. We're working with a customer now to validate this process and needed to be able to give them a process to follow, but it is still not ready for public consumption.

PR #2414, opened 2024-08-13
Update favicon

tonykelly-octopus:
Adding the latest version of the icon this time around


steve-fenton-octopus (approved):
Nice - this will work better in both light and dark mode.

PR #2420, opened 2024-08-16
Updating dacpac instructions

twerthi:
Updating instructions as they were referencing a template that has been deprecated and removed.


steve-fenton-octopus (approved):
LGTM!

PR #2427, opened 2024-08-22
Updating Administration area screenshots

MarkSiedle:
As owners of the `Administration area`, and as part of [this request for R&D in Q3](https://octopusdeploy.slack.com/archives/C025P7WQ3G9/p1718359925727959), the Backend Foundations Team are updating the screenshots to match the latest Vertical Navigation design.

This PR covers:
* Diagnostics
* High-Availability (nodes)
* Script Console
* Maintenance Mode
* Auditing
* Subscriptions
* Spaces

Reviewer notes:
* I don't recommend going to the trouble of running this up locally, as docs and server have different `pnpm` versions and you will have a bad time 🙋 . Instead, if you could just eye-ball the image diffs and confirm they are effectively same-same with with the newer design, that would be much appreciated 🙇 
* There are no width/height conditions on our doc images, and CSS styling will apply rounded corners to cover up any slight discrepancies on the edges
* In some cases (like HA), I've opted to crop the existing image content, rather than go through the effort of setting up HA locally. This still gives customers enough of a screenshot hint, without the old design confusion


desruc (approved):
I've eyeballed everything using the swipe functionality and it all looks correct. There are some small differences, like the script saying `testing` now, instead of `Hello World`, but I don't think that matters.

PR #2429, opened 2024-08-23
Switch the test logic to allow full redirect addresses.

steve-fenton-octopus:
This resolves the issue with the redirects appearing in the nav.


steve-fenton-octopus:
Deploying to staging for a double-check.

steve-fenton-octopus:
Staging looks good - solves the issue.

KMajkrzakOctopus (approved):
👌

PR #2445, opened 2024-08-28
Update updating step templates page

N-lson:
Reduced to a single screenshot of the dialog and lifted it to the top of the sub-section. I think it makes more sense this way.

## Old
image

## New
image


steve-fenton-octopus (approved):
Nice! LGTM!

PR #2446, opened 2024-08-28
Rework multi-tenant SaaS guide to only one page and reduce screenshots

N-lson:
This guide has way too many screenshots. A lot of them are already shown on other pages, and the samples instance can be used as a more up-to-date reference.

The guide itself isn't that long so it feels more reasonable to keep it all on a single page. Open to thoughts.

Live version of the pages: https://octopus.com/docs/tenants/guides/multi-tenant-saas-application


steve-fenton-octopus (approved):
I agree this is better as one page.

PR #2448, opened 2024-08-29
Refresh tenant types and tags pages

denys-octopus:
- Update screenshots in tenant-tags page
- Update tenant types page


src/pages/docs/tenants/tenant-types.md
```diff
@@ -3,6 +3,7 @@ layout: src/layouts/Default.astro
 pubDate: 2023-01-01
 modDate: 2023-12-07
```
denys-octopus, line 4: I assume we don't need to update `modDate` when only icon is added...

N-lson (approved):

PR #2449, opened 2024-08-29
Consolidate multi-tenant region guides to single page

N-lson:
Changes include:

- Reducing number of screenshots
- Aligning some sections to match the changes made in https://github.com/OctopusDeploy/docs/pull/2446 for the SaaS guides
- Changed the step in "Configure steps to use a worker pool variable" since the old step didn't actually use a worker pool variable on the samples instance

This one is a bit longer than the SaaS one, so I think there's an argument that the "Create infrastructure" (because of the example script) and "Region-specific workers" sections could be split out to their own pages.


steve-fenton-octopus (approved):
LGTM!

PR #2450, opened 2024-08-29
Update tenants sharing guide (not completely)

N-lson:
I didn't update many of the screenshots so that still needs to be done. Hoping the solutions team can take the screenshots since they should be able to edit the demo instance.


steve-fenton-octopus (approved):
LGTM!

PR #2466, opened 2024-09-09
added version warning for bento migrations

benPearce1:
Added warning around version support for bento migrations.
Due to the nature of GA being 1 version behind Cloud, migration from cloud to self hosted is not practically supported.


hnrkndrssn (approved):
LGTM

PR #2476, opened 2024-09-10
Minor language cleanup for custom Octopus Server certificates with K8s Agent

APErebus:
Make it clearer that it's the public key of the root organization CA certificate


kevjt (approved):
lgtm

PR #375, opened 2019-04-08
Adding advice on possible high latency or network congestion in Octopus Cloud

lawrencewilson:

PR #2484, opened 2024-09-15
Update Octopus Cloud Getting Started

goosmurf:
Structure/content changes are individually noted.

Other changes are primarily to match instructions to our current UI.

[sc-82146]

### DIFF
--- a/src/pages/docs/octopus-cloud/getting-started-with-cloud.mdx
+++ b/src/pages/docs/octopus-cloud/getting-started-with-cloud.mdx
@@ -1,7 +1,7 @@
 ---
 layout: src/layouts/Default.astro
 pubDate: 2023-01-01
-modDate: 2024-05-07
+modDate: 2024-09-12
 title: Getting started with Cloud
 navOrder: 5
 description: Learn how to setup your Cloud instance, add users to your instance, and set the maintenance window
--- a/src/pages/docs/octopus-cloud/getting-started-with-cloud.mdx
+++ b/src/pages/docs/octopus-cloud/getting-started-with-cloud.mdx
@@ -110,9 +110,3 @@ We keep your instance of Octopus Cloud up-to-date and running the latest version
 2. In the bottom left corner, look for the Octopus logo, or your Gravatar if you have one setup. 
 3. Click the Octopus logo (or your Gravatar) to open the menu. 
 2. The version is shown under your name.
-
-## Uploading packages to Octopus Cloud servers \{#uploading-packages-to-octopus-cloud-servers}
-
-Factors like geographical distance, network bandwidth, and network congestion may cause package uploads to your Octopus Cloud Server to take longer than expected.
-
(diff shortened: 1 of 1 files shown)


src/pages/docs/octopus-cloud/getting-started-with-cloud.mdx
```diff
@@ -23,78 +23,90 @@ After you create an Octopus account, you can create a new instance of Octopus Cl
...
 5. Review the terms of our customer agreement, privacy policy, and acceptable usage policy.
-6. Click **Agree, deploy my Octopus**!
+6. Click **Start your trial**.
 
 You’ll be taken to the account provisioning screen. Your Octopus Cloud instance should be ready in a minute or so. You’ll receive an email when the instance is ready to use.
 
-When the instance is ready, you’ll see it (and any other instances you have access to) the next time you log in to your Octopus account at [octopus.com/signin](https://octopus.com/signin).
+## Managing Octopus Cloud subscriptions \{#managing-octopus-cloud-subscriptions}
 
-## Uploading packages to Octopus Cloud servers \{#uploading-packages-to-octopus-cloud-servers}
```
goosmurf, line 37: This section was introduced in #375 over 5 years ago. I've taken it out because I don't think it belongs on a "Getting Started" page.

I think in general people understand how networks work, and this note is probably only useful to a customer migrating from self-hosted to Cloud. The context there would be if they were deploying from self-hosted to network-close targets (e.g. on-prem to on-prem, or self-hosted in Cloud to targets within the same Cloud region).

A better place for this might be `Migrating to Cloud`, I might add it back there when I get to that doc.

src/pages/docs/octopus-cloud/getting-started-with-cloud.mdx
```diff
@@ -23,78 +23,90 @@ After you create an Octopus account, you can create a new instance of Octopus Cl
...
+:::div{.hint}
+We are in transition between two systems for managing Octopus Cloud subscriptions. We refer to the new system as Control Center V2, and it replaces the original [Control Center V1](https://octopus.com/control-center). **Accounts** are created in both systems but all new **subscriptions** are created in [Control Center V2](https://billing.octopus.com/).
+:::
 
-Factors like geographical distance, network bandwidth, and network congestion may cause package uploads to your Octopus Cloud Server to take longer than expected.
+You can find all of your Octopus Cloud subscriptions in [Control Center V2](https://billing.octopus.com/).
 
-If you’re having difficulty uploading packages to your Octopus built-in package feed in the default timeout threshold, typically within 5 minutes, you might be affected by one or more of the factors mentioned above. In this case, you can try increasing the timeout threshold of the Octopus CLI or nuget.exe to a value that suits you. We also recommend using the Octopus CLI, **Octopus.Client**, or the **TeamCity Plugin (v4.41.0+)** because they use delta compression.
-
-## Change your password \{#change-your-password}
```
goosmurf, line 43: Moved this down the page as I think it is more important to guide users how to find their instance before worrying about how they should change their password.

steve-fenton-octopus (approved):

PR #2491, opened 2024-09-16
Add Custom certificates support scope message

samdanaei:


dictionary-octopus.txt
```diff
@@ -142,6 +142,7 @@ HKLM
 hostnames
 hotfix
 hotfixes
+Howtousecustomcertificateswith
```
steve-fenton-octopus, line 145: This definitely isn't a word :) I think we need a different fix for this.

samdanaei: Thanks for the help @steve-fenton-octopus

PR #2498, opened 2024-09-16
Change cli link to github releases page.

geoffrichoctopus:
There are old links to the downloads page on octopus.com which is an out of date page and a source of confusion for customers.

Fix these links to direct users to the github releases page for that same cli tool (the deprecated one).


wlthomson (approved):
Thanks for this!

PR #2505, opened 2024-09-17
Secrets example

susanpann:


tothegills:
I think the URL should also be a pipeline variable

susanpann:
> I think the URL should also be a pipeline variable

Oki, have added a separate section for variables

tothegills (approved):
Looks great!

PR #2360, opened 2024-06-25
Reviewing naming of Variable Sets

MarkSiedle:
In preparation for Vertical Navigation's release next week, we've been reviewing docs and discovered various combinations of the wording around "Variable Sets".

This PR reviews the various wording and standardises where appropriate, based on both the current UI and the new UI coming soon. We no longer refer to "Variable Sets" as "Library Variable Sets" or "Library Sets". They are just Variable Sets.

PR #2516, opened 2024-09-17
Updating Security area screenshots

MarkSiedle:
As owners of the `Security` area and as part of [this request for R&D in Q3](https://octopusdeploy.slack.com/archives/C025P7WQ3G9/p1718359925727959), the Backend Foundations Team are updating the screenshots to match the latest Vertical Navigation design (or remove layout concerns from screenshots where appropriate).

Reviewer notes:
* I don't recommend going to the trouble of running this up locally, as docs and server have different `pnpm` versions and you will have a bad time 🙋 . Instead, if you could just eye-ball the image diffs and confirm they are effectively same-same with with the newer design, that would be much appreciated 🙇 
* There are no width/height conditions on our doc images, and CSS styling will apply rounded corners to cover up any slight discrepancies on the edges
* In many cases, we've opted to crop the existing image content, rather than go through the effort of setting up the same scenario locally for a new screenshot. This still gives customers enough of a screenshot hint, without the old design confusion


src/pages/docs/security/users-and-teams/security-and-un-scoped-variables.md
```diff
@@ -9,7 +9,7 @@ navOrder: 1
 
 Sometimes, a team will be granted contributor access to a project, but be restricted in the environments that it can access. By default, Octopus's security system will then prevent members of the team editing [variables](/docs/projects/variables) that apply outside of their allowed environments. During development this can be inconvenient, as variables frequently need to be added in support of new application features.
 
-## Why restrict editing of unscoped variables? {#Securityandun-scopedvariables-Whyrestricteditingofun-scopedvariables?}
+## Why restrict editing of unscoped variables? {#why-restrict-editing-of-unscoped-variables}
```
MarkSiedle, line 12: I followed the same approach Steve Fenton used in this previous PR where we encountered this: https://github.com/OctopusDeploy/docs/pull/2360/commits/2bd1155bb6a346f4ff1819565f39c3bc515f9a70#diff-1e8a102c0408e76c53dd207562893eff2122c73073bd5e5f1da9dbb7f399b467

And confirmed there's no anchor link references to these that needed updating.

desruc (approved):
I didn't spin this up locally, but I went through each image and confirmed that they are like for like.

PR #2518, opened 2024-09-18
Update to pnpm 9.10.0

APErebus:
Updates to pnpm `9.10.0`, which will be consistent with the version used in the frontend (being updated in https://github.com/OctopusDeploy/OctopusDeploy/pull/27659)


APErebus:
> I adjusted the actions to sync with this change.
> 
> 
> 
> Should be ready to go.

Ah yep, thanks! Forgot about those.

PR #2521, opened 2024-09-19
Update @pnpm/exe to 9.10.0

APErebus:
Missed this one when doing #2518


geofflamrock (approved):
Looks good, I've tested this locally and it works now, thanks.

PR #2530, opened 2024-09-24
Dynamic Workers refresh

goosmurf:
Minor touchups as part of #project-docs-refresh.

I have removed the mentions of Ubuntu1804 and Windows2019 EOL as the images have been unavailable for 17 and 8 months respectively. It's hard to imagine anyone still using those old images as their deployments would have been failing all this while, or they are very old projects that have not been deployed since the worker images went EOL.

However, I have left the EOL pages. Users will get there either via the navbar, or direct links (e.g. from the notification emails we  sent during the EOL process).
* https://octopus.com/docs/infrastructure/workers/dynamic-worker-pools/ubuntu-1804-end-of-life
* http://localhost:3000/docs/infrastructure/workers/dynamic-worker-pools/windows-2019-end-of-life

I am unable to update the screenshots in the EOL pages so these pages will reflect old versions of Octopus.

An alternative is to remove the contents of the EOL pages and replace them with redirects to the top level Dynamic Worker Pools page. The con of this approach is that anyone actually looking for EOL information won't find what they're looking for. If we went this route, we might want to leave the deprecation notices within the top level page.

[sc-82146]


src/pages/docs/infrastructure/workers/dynamic-worker-pools.md
```diff
@@ -71,12 +71,11 @@ You should validate that your deployments and runbooks work with the new version
 
 ### Modifying the worker pool 
 
-If the Worker Pool has been configured to specifically use a deprecated worker type, you will need to update the Worker Image on the Worker Pool.
+If the worker pool has been configured to specifically use a deprecated worker type, you will need to update the worker image on the worker pool.
 
-The Worker Pool with a deprecated Worker Type will show a `Deprecated` label next to the worker pool, available by navigating to **Infrastructure ➜ Worker Pools**:
-![Worker Pool list with deprecated worker](/docs/infrastructure/workers/images/deprecated-worker-pool-overview.png)
```
goosmurf, line 77: I've removed this image altogether as there is no `Deprecated` image to take an example screenshot of. It's also of low utility, since there would be a vanishingly small number of customers who would not already have migrated off Ubuntu1804 or Windows2019 worker images.

steve-fenton-octopus (approved):

PR #2531, opened 2024-09-25
chore(docs): Add troubleshooting details for users with 200+ security groups

hnrkndrssn:
Add a new troubleshooting section for EntraID users with 200+ security groups as this requires specific configuration of the app registration in Azure that may be missing.

image

[sc-93371]


IsaacCalligeros95 (approved):
LGTM

Issue #2399, opened 2024-08-01
Upload artifact to existing deployment includes headers in file content


szogun1987:
I have converted [this article](https://octopus.com/docs/octopus-rest-api/examples/artifacts/create-and-upload-artifacts) to use curl: 

```
# Upload file content
curl -v -X PUT -F fileData=@report.html -H "X-Octopus-ApiKey: ${OCTOPUS_API_KEY}" "${uploadUrl}"
```

Unfortunatelly it includes http part data:
```
--------------------------246875bc4e54cce7
Content-Disposition: form-data; name="fileData"; filename="report.html"
Content-Type: text/html

<!DOCTYPE html>
<html lang="en">
```
This is big issue for binary files, I can live with it for text ones.

Probably it is issue with octopus itself not with documentation so please move this issue to the right place.

steve-fenton-octopus:
Hi @szogun1987 

Can you try using the `--data-binary` flag to see if it resolves your issue?

```
curl -v \
  -X PUT \
  --data-binary @report.html \
  -H "X-Octopus-ApiKey: ${OCTOPUS_API_KEY}" \
  "${uploadUrl}"
```

szogun1987:
Thank you @steve-fenton-octopus  now there is no extra initial bytes. But content is truncated.
HTML was only example I upload tar.gz file with screenshots

steve-fenton-octopus:
Hi @szogun1987 

I've asked our R&D team to take a look at this one. I'll let you know the outcome as soon I hear.

PR #2537, opened 2024-10-18
fix: update upload artifact to deployment script to use byte stream

grace-rehn:
The script to upload an artifact to an existing deployment prepends extra headers in the file content when uploading to Octopus.

This PR updates the script to use a byte stream instead of streaming as multipart form data. This ensures that the header content is not added to the artifact.

Fixes https://github.com/OctopusDeploy/docs/issues/2399

Tested using [artifact.csv](https://github.com/user-attachments/files/17427805/artifact.csv)

### Before

Running the original script and downloading the artifact from Octopus:

Screenshot 2024-10-18 at 2 05 59 PM

### After

Running the new script and downloading the artifact from Octopus:

Screenshot 2024-10-18 at 2 06 29 PM


IsaacCalligeros95 (approved):
LGTM

benPearce1 (approved):

Issue #2129, opened 2023-12-19
Dark Mode?


antmeehan:
Any chance dark mode support is on the radar for the documentation site? So bright 😎

steve-fenton-octopus:
Hello @antmeehan 

Yes, this is on our plan. We have a set of design updates coming that includes a dark mode based on your preference.

Hopefully you won't need the sunglasses too much longer.

Thanks,
Steve

antmeehan:
Thanks @steve-fenton-octopus. Looking forward to it.

steve-fenton-octopus:
Additional note for code blocks.

We introduced a light/dark mode for code blocks. When we implement dark mode it would good to update this to use a separate Shiki theme for each one using [dual themes](https://shiki.style/guide/dual-themes).

The interim version uses calculated values: https://github.com/OctopusDeploy/docs/pull/2147

steve-fenton-octopus:
Also see #2258 

Dark mode is on its way, including user-preference detection and a toggle.

steve-fenton-octopus:
@antmeehan we now have light and dark modes available. Let me know what you think.

![Docs site in light mode](https://github.com/OctopusDeploy/docs/assets/99181436/fdab8b60-032a-4c86-a3c9-4584c622de52)

![Docs site in dark mode](https://github.com/OctopusDeploy/docs/assets/99181436/61c0509c-9966-4e59-901a-d2d70f640e96)

antmeehan:
Great work! Looks good. One comment so far is that the font on the code examples is a little hard to read:

![2024_05_07_21_09_09_Clipboard](https://github.com/OctopusDeploy/docs/assets/9822503/15df4d80-c40a-4ead-9dce-d6414743f8c3)

steve-fenton-octopus:
Thanks @antmeehan - We'll take a look at this and tune it up.

What browser are you using?

antmeehan:
Google Chrome on Windows 11. Dev Tools says its Consolas:

![image](https://github.com/OctopusDeploy/docs/assets/9822503/3fe781f3-006d-4f42-8855-b5b602de292f)

steve-fenton-octopus:
@antmeehan the code font should be better now! Let me know if you still see any issues.

PR #2540, opened 2024-10-25
Code font is too small for readability

steve-fenton-octopus:
Fixes the follow up issue on #2129

# Before

image

# After

image


KMajkrzakOctopus (approved):
👌

PR #2541, opened 2024-10-28
Adding troubleshooting tip

twerthi:
Adding an annotation wasn't as straight-forward as I thought, this is to help others who encounter similar issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     },
     "dependencies": {
         "@astrojs/mdx": "^2.3.1",
+        "@octopus/docs": "file:",
         "@pnpm/exe": "^9.7.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -24,6 +24,7 @@
     },
     "dependencies": {
         "@astrojs/mdx": "^2.3.1",
+        "@octopus/docs": "file:",
```
steve-fenton-octopus, line 27: @twerthi do you know what this is for? I thought it was maybe an IDE auto addition that we don't need?

steve-fenton-octopus:
This branch seems super-out-of-date with main, so I've cut a new branch and migrated the change.

See #2542

PR #2543, opened 2024-10-29
Update docs to include codefresh login step

bec-callow-oct:
This PR is to update the existing Codefresh integration documentation to include a new Octopus login step. This step authenticates to Octopus via OIDC. All other Octopus Codefresh steps can now accept either an Octopus Access Token or API key as authentication.

SC-96167


tothegills (approved):
Nice set of changes. I like that you've provided `OCTOPUS_ACCESS_TOKEN` as the default and `OCTOPUS_API_KEY` as the alternative.

steve-fenton-octopus (approved):

PR #2545, opened 2024-10-29
Image changes for DFS page

tonykelly-octopus:
Updating images on DFS page so they work with a dark background.

image


KMajkrzakOctopus (approved):
👌

PR #2546, opened 2024-10-30
Image border fix on DFS page

tonykelly-octopus:
Image border isn't applied correctly on last image on the page

image


KMajkrzakOctopus (approved):
👌

PR #2548, opened 2024-11-04
Updates references to AES to 256 bit where appropriate

andrewabest:
Documentation changes to accompany https://github.com/OctopusDeploy/OctopusDeploy/pull/28689, which changed our default master key length to 256 bits.

### DIFF
--- a/src/pages/docs/administration/data/backup-and-restore.md
+++ b/src/pages/docs/administration/data/backup-and-restore.md
@@ -18,7 +18,10 @@ A successful disaster recovery plan for Octopus Deploy requires the ability to r
 :::div{.problem}
 
 **Without your Master Key, backups are useless**
-Sensitive information is encrypted using AES-256 with the Master Key as the encryption key. Without this Master Key you will lose your sensitive variables, passwords, and other encrypted data. Make sure you've taken a copy of the key! [Learn more about backing up the Master Key](/docs/security/data-encryption).
+
+Sensitive information is encrypted using AES with the Master Key as the encryption key. Without this Master Key you will lose your sensitive variables, passwords, and other encrypted data. Make sure you've taken a copy of the key! [Learn more about backing up the Master Key](/docs/security/data-encryption).
+
+Octopus Server 2024.4 and newer use AES-256 by default but support AES-128 for compatibility. Previous versions use AES-128.
 
 :::
 
--- a/src/pages/docs/administration/data/backup-and-restore.md
+++ b/src/pages/docs/administration/data/backup-and-restore.md
@@ -1,7 +1,7 @@
 ---
 layout: src/layouts/Default.astro
 pubDate: 2023-01-01
-modDate: 2024-07-15
+modDate: 2024-11-04
 title: Backup and restore
 description: Information on where Octopus data is stored and how to backup and restore an Octopus instance.
 navOrder: 10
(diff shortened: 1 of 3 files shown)


src/pages/docs/administration/data/backup-and-restore.md
```diff
@@ -18,7 +18,7 @@ A successful disaster recovery plan for Octopus Deploy requires the ability to r
 :::div{.problem}
 
 **Without your Master Key, backups are useless**
-Sensitive information is encrypted using AES128 with the Master Key as the encryption key. Without this Master Key you will lose your sensitive variables, passwords, and other encrypted data. Make sure you've taken a copy of the key! [Learn more about backing up the Master Key](/docs/security/data-encryption).
+Sensitive information is encrypted using AES-256 with the Master Key as the encryption key. Without this Master Key you will lose your sensitive variables, passwords, and other encrypted data. Make sure you've taken a copy of the key! [Learn more about backing up the Master Key](/docs/security/data-encryption).
```
borland, line 21: I think this would be fine wording if we'd used AES-256 for years, but given that customers will start reading this as soon as tomorrow, whilst the actual AES change won't reach them for _months_ or even _years_ I feel like we'd be better to qualify it. Perhaps something like

> Sensitive information is encrypted using AES* with the Master Key as the encryption key...
>
> * Octopus Server 2024.4 and newer use AES256 by default but support AES128 for compatibility. Older versions use AES128.

borland: Or just don't specify the key length at all; anyone who's interested would likely then follow the link to the data-encryption doc

src/pages/docs/administration/data/backup-and-restore.md
```diff
@@ -58,6 +58,6 @@ When connecting to an existing database, you will be prompted for this key durin
 :::div{.problem}
 
 **Without your Master Key, backups are useless**
-Sensitive information is encrypted using AES128 with the Master Key as the encryption key. Without this Master Key you will lose sensitive variables, passwords, and other encrypted data. Make sure you've taken a copy of the key! [Learn more about backing up the Master Key](/docs/security/data-encryption).
+Sensitive information is encrypted using AES-256 with the Master Key as the encryption key. Without this Master Key you will lose sensitive variables, passwords, and other encrypted data. Make sure you've taken a copy of the key! [Learn more about backing up the Master Key](/docs/security/data-encryption).
```
borland, line 61: As above

src/pages/docs/projects/variables/sensitive-variables.md
```diff
@@ -13,7 +13,7 @@ As you work with [variables](/docs/projects/variables) in Octopus, there will be
 Sensitive variables can be sourced from either:
 
 - A Secret Manager/Key Vault using one of our Community step templates.
-- Octopus itself, with values stored securely using **AES128 encryption**.
+- Octopus itself, with values stored securely using **AES-256 encryption**.
```
borland, line 16: As above

src/pages/docs/administration/data/backup-and-restore.md
andrewabest: Addressed.

src/pages/docs/projects/variables/sensitive-variables.md
andrewabest: Addressed.

src/pages/docs/administration/data/backup-and-restore.md
andrewabest: Addressed.

borland (approved):
Thanks for making those suggested changes, LGTM

PR #2564, opened 2024-11-19
Adds examples documentation for enabling/disabling tenants.

IsaacCalligeros95:
Adds examples docs for enabling and disabling tenants. At this stage we're not calling it out in the Tenants section, if you think we should let me know.
image


src/shared-content/scripts/enable-disable-tenant-scripts.include.md
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/tenants"
+)
+
+func main() {
+	apiURL, err := url.Parse("http://localhost:8066")
+	if err != nil {
+		log.Fatalf("Error parsing API URL: %v", err)
+	}
+
+	APIKey := "API-K6JR7YR2UTPH2BQZH1INRZUICEXHDF"
```
benPearce1, line 173: should be updated to `API-YOUR-KEY`

src/shared-content/scripts/enable-disable-tenant-scripts.include.md
```diff
@@ -0,0 +1,247 @@
...
+	"log"
+	"net/url"
+
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/spaces"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/tenants"
+)
+
+func main() {
+	apiURL, err := url.Parse("http://localhost:8066")
```
benPearce1, line 168:
```suggestion
	apiURL, err := url.Parse("https://your-octopus-url")
```

src/shared-content/scripts/enable-disable-tenant-scripts.include.md
benPearce1, line 173:
```suggestion
	APIKey := "API-YOUR-KEY"
```

benPearce1 (approved):
couple of small tweaks

steve-fenton-octopus:
One thing we ought to consider is that we're using words here from our [non-inclusive word list](https://www.octopus.design/932c0f1a9/p/457bc4-grammar-rules/t/page-457bc4-58679287-21).

If there's a way to switch to active/inactive tenants here, that would read far better. I'm not a fan of us ranking for "disabled tenants" or making folks uncomfortable with this term.

steve-fenton-octopus (approved):

PR #2568, opened 2024-11-20
Logo updated

KMajkrzakOctopus:
After:
Screenshot 2024-11-20 at 18 45 29
Screenshot 2024-11-20 at 18 45 34


KMajkrzakOctopus:
@steve-fenton-octopus  could you take a look what is wrong in the deploy to channel pipeline? I don't want to mess up anything but seems like something is wrong in here:
`tests/bookmark.spec.ts`

KMajkrzakOctopus:
> @steve-fenton-octopus could you take a look what is wrong in the deploy to channel pipeline? I don't want to mess up anything but seems like something is wrong in here: `tests/bookmark.spec.ts`

Nevermind turns out someone fixed that with a latest pr

steve-fenton-octopus:
@KMajkrzakOctopus as the SVG is growing, I wonder if it might be an idea to move it into an `.svg` file so browsers can cache it, rather than inlining it? What are your thoughts?

steve-fenton-octopus (approved):

PR #2396, opened 2024-07-31
Tweak agent tooling to be clearer

APErebus:
The `Agent tooling` section wasn't clear that the agent does not use execution containers for all k8s steps _past_ a certain version

PR #2569, opened 2024-11-20
Update Kubernetes agent tooling

APErebus:
This is a re-application of #2396 as it appears to have been lost during the `kubedocs` re-organization


rain-on (approved):
LGTM

PR #2570, opened 2024-11-21
Small improvements to GH action docs

caitlynstocker:
## What's this? 🐙 

While dogfooding my own GH action docs, I realised they could be a bit clearer (and more correct). 

I will comment an explanation for each change on the code. Feel free to suggest some more changes and I'll add them, if you've been using the docs yourself too!

## How to review 🔍 
✅ Correctness
☑️ Typos etc. 

Fixes[sc-93404]

### DIFF
--- a/src/pages/docs/packaging-applications/build-servers/github-actions.md
+++ b/src/pages/docs/packaging-applications/build-servers/github-actions.md
@@ -29,7 +29,7 @@ Octopus Deploy provides GitHub Actions which enable you to:
 
 ## Getting started
 
-Octopus Deploy GitHub Actions can be easily incorporated into your own GitHub Action workflows by including them as steps in your workflow YAML. Here is a simple GitHub Action workflow YAML to get you started.
+Octopus Deploy GitHub Actions can be easily incorporated into your own GitHub Action workflows by including them as steps in your workflow YAML. Here is a simple workflow YAML to get you started.
 
 ### Example workflow - Create and deploy a release
 ```yaml
@@ -97,7 +97,6 @@ jobs:
 | `OCTOPUS_SPACE`   | The Name of the Space where this command will be executed.                                                                                               |
 | `OCTOPUS_URL`     | The base URL hosting Octopus Deploy (i.e. `https://octopus.example.app`). It is strongly recommended that this value retrieved from a [GitHub secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions). |
 | `OCTOPUS_API_KEY` | The API key used to access Octopus Deploy. It is strongly recommended that this value retrieved from a [GitHub secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions). |
-| `SERVER`          | The base URL hosting Octopus Deploy (i.e. `https://octopus.example.app`). It is strongly recommended that this value retrieved from a [GitHub secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions). |
 
 
 ### 📥 Inputs
@@ -108,8 +107,9 @@ jobs:
 | `release_number`      | The number for the new Release. If omitted, Octopus Deploy will generate a Release number.                                                           |
 | `environments`        | A list of Environments in Octopus Deploy in which to run (i.e. Dev, Test, Prod). Add each environment on a new line.                                 |
 | `variables`           | A list of Variables to use in the Deployment in `key: value` format. Add each variable on a new line.                                                |
-| `git_ref`             | The Git branch from which to source the project code. Required for Projects using version control in Octopus. The example above sources this value from the main action's [contextual information.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) |
-| `git_commit`          | The Git commit from which to source the project code. Required for Projects using version control in Octopus. The example above sources this value from the main action's [contextual information.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) |
+| `git_ref`             | The Git branch from which to source the project code. Required for Projects using version control in Octopus. The example above sources this value from the workflow's [contextual information.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) |
+| `git_commit`          | The Git commit from which to source the project code. Required for Projects using version control in Octopus. The example above sources this value from the workflow's [contextual information.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) |
+| `server`              | The base URL hosting Octopus Deploy (i.e. `https://octopus.example.app`). It is strongly recommended that this value retrieved from a [GitHub secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions). |
 | `service_account_id`  | The id of the OIDC service account you wish to login as. Service accounts can be viewed and created on the Octopus app under 'Users' on the configuration menu. |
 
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/packaging-applications/build-servers/github-actions.md
```diff
@@ -88,8 +88,6 @@ jobs:
         variables: |
             Flip: Bling
             Fizz: Buzz
-        git_ref: ${{ github.ref }}
-        git_commit: ${{ github.sha }}
```
caitlynstocker, line 92: Turns out these values are not actually required for this action! They don't break it but they're just being ignored and causing a warning message.

src/pages/docs/packaging-applications/build-servers/github-actions.md
```diff
@@ -109,10 +108,9 @@ jobs:
...
-| `service_account_id`  | The id of the service account you wish to login as.                                                                                                  |
+| `git_ref`             | The Git branch from which to source the project code. Required for Projects using version control in Octopus. The example above sources this value from the GitHub Action's [contextual information.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) |
+| `git_commit`          | The Git commit from which to source the project code. Required for Projects using version control in Octopus. The example above sources this value from the GitHub Action's [contextual information.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) |
```
caitlynstocker, line 112: When I came back to use this example code a month later, it wasn't clear to me where the example values were being sourced.

src/pages/docs/packaging-applications/build-servers/github-actions.md
```diff
@@ -99,6 +97,7 @@ jobs:
 | `OCTOPUS_SPACE`   | The Name of the Space where this command will be executed.                                                                                               |
 | `OCTOPUS_URL`     | The base URL hosting Octopus Deploy (i.e. `https://octopus.example.app`). It is strongly recommended that this value retrieved from a [GitHub secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions). |
 | `OCTOPUS_API_KEY` | The API key used to access Octopus Deploy. It is strongly recommended that this value retrieved from a [GitHub secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions). |
+| `SERVER`          | The base URL hosting Octopus Deploy (i.e. `https://octopus.example.app`). It is strongly recommended that this value retrieved from a [GitHub secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions). |
```
caitlynstocker, line 100: I've just moved this one up from the Inputs list as it's actually an Environment Variable in the example.

src/pages/docs/packaging-applications/build-servers/github-actions.md
```diff
@@ -109,10 +108,9 @@ jobs:
...
-| `service_account_id`  | The id of the service account you wish to login as.                                                                                                  |
+| `git_ref`             | The Git branch from which to source the project code. Required for Projects using version control in Octopus. The example above sources this value from the GitHub Action's [contextual information.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) |
+| `git_commit`          | The Git commit from which to source the project code. Required for Projects using version control in Octopus. The example above sources this value from the GitHub Action's [contextual information.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) |
+| `service_account_id`  | The id of the OIDC service account you wish to login as. Service accounts can be viewed and created on the Octopus app under 'Users' on the configuration menu. |
```
caitlynstocker, line 113: This one just needed more detail to make the explanation useful.

caitlynstocker:
Last week we discussed changing the example workflow to reflect the ability to trigger deployments from releases. 

Perhaps for the sake of providing a simple hello-world example here, it might be better to leave this one the way it is, as long as we're still thinking of providing more comprehensive examples elsewhere?

src/pages/docs/packaging-applications/build-servers/github-actions.md
geofflamrock: Just want to confirm: Is `SERVER` the right environment variable here? It doesn't sound familiar to me, I _thought_ it was `OCTOPUS_URL` or maybe `OCTOPUS_SERVER`???

src/pages/docs/packaging-applications/build-servers/github-actions.md
```diff
@@ -109,10 +108,9 @@ jobs:
...
-| `git_commit`          | The Git commit from which to source the project code. Required for Projects using version control in Octopus.                                        |
-| `server`              | The base URL hosting Octopus Deploy (i.e. `https://octopus.example.app`). It is strongly recommended that this value retrieved from a [GitHub secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions). |
-| `service_account_id`  | The id of the service account you wish to login as.                                                                                                  |
+| `git_ref`             | The Git branch from which to source the project code. Required for Projects using version control in Octopus. The example above sources this value from the main action's [contextual information.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) |
```
geofflamrock, line 111:
```suggestion
| `git_ref`             | The Git branch from which to source the project code. Required for Projects using version control in Octopus. The example above sources this value from the workflow run [contextual information.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) |
```

A small suggestion here: The naming for GitHub Actions is a bit weird, the overall "thing" that runs is called a workflow, but the overall product feature is called "GitHub Actions" and each individual step runs an "action". I think in this case the context is related to the workflow run.

src/pages/docs/packaging-applications/build-servers/github-actions.md
```diff
@@ -109,10 +108,9 @@ jobs:
 | `release_number`      | The number for the new Release. If omitted, Octopus Deploy will generate a Release number.                                                           |
 | `environments`        | A list of Environments in Octopus Deploy in which to run (i.e. Dev, Test, Prod). Add each environment on a new line.                                 |
 | `variables`           | A list of Variables to use in the Deployment in `key: value` format. Add each variable on a new line.                                                |
-| `git_ref`             | The Git branch from which to source the project code. Required for Projects using version control in Octopus.                                        |
-| `git_commit`          | The Git commit from which to source the project code. Required for Projects using version control in Octopus.                                        |
-| `server`              | The base URL hosting Octopus Deploy (i.e. `https://octopus.example.app`). It is strongly recommended that this value retrieved from a [GitHub secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions). |
```
geofflamrock, line 114: The `server` input looks like it is still actually used in the example, in the `OctopusDeploy/login` step, I wonder if it might be worth keeping it in here in addition to the env var section.

geofflamrock (commented):
Generally this looks like a good improvement, thanks for tackling it 👍 I've left a couple of small thoughts and questions.

src/pages/docs/packaging-applications/build-servers/github-actions.md
caitlynstocker: Righhhht! Thank you! I've always heard 'workflow' and 'action' used pretty much interchangeably, but ChatGPT says you are right 😅

src/pages/docs/packaging-applications/build-servers/github-actions.md
caitlynstocker: Ahhh! You're right. It is. I got confused between the 'server' variable and the 'OCTOPUS_URL' env var.

src/pages/docs/packaging-applications/build-servers/github-actions.md
geofflamrock:
```suggestion
| `service_account_id`  | The id of the service account you wish to login with using OpenID Connect (OIDC). Service accounts can be viewed and created on the Octopus app under 'Users' on the configuration menu. |
```

A small suggestion here: A service account can use API keys or OIDC to login, and it feels like calling it an "OIDC service account" might lead users to think they are a different concept. Having something mentioned about OIDC makes sense as this is the use case for that property, does moving it to the end sound ok?

geofflamrock (approved):
Looks good, I've left one more small suggestion for you to consider 🙂

PR #2572, opened 2024-11-24
Update actually supported Kubernetes versions for the agent

APErebus:
It turns out we had upgraded the SDK used by the agent a few versions ago (by @OctopusDeploy/team-server-at-scale as part of the .NET 8 work).

As a result, the supported versions of the agent change to 1.28 -> 1.31


MissedTheMark (approved):
🚀

PR #2576, opened 2024-11-27
Add Cloud uptime track record; update SLO; update references.

adrianparker:
Improves, defines and clarifies content about Octopus Cloud uptime SLO. 

Adds a table of track record data.

Add recent & accurate data points to maintenance windows page.

Link FAQ to uptime track record.


steve-fenton-octopus (approved):
Ace! Thanks for the update.

PR #2577, opened 2024-11-28
Remove Fedora 39 from our supported platforms list

sburmanoctopus:
We no longer support Fedora 39. So it should be removed from our docs.


gb-8 (approved):
This was the change I expected.

evolutionise (approved):

PR #2585, opened 2024-12-09
Updated dependencies

steve-fenton-octopus:
Updates NPM packages


steve-fenton-octopus:
Short.io links updated to reduce redirects and preserve the bookmark links.

PR #2590, opened 2024-12-11
Kubernetes agent supported versions policy

eddymoulton:
Adds a short explainer to describe when we make updates to the supported versions of Kubernetes.

Unsure if we want to direct people to talk to support for something that we'll get to eventually - but I think it's a nice touch.

image


steve-fenton-octopus (approved):

APErebus (approved):

eddymoulton:
[[sc-98669]]

PR #2591, opened 2024-12-11
Update outbound request documents related to in app messaging

mandymichael:
Updating the outbound requests documentation relating to In App Messaging.


steve-fenton-octopus (approved):
You'll need to sign the contribution license agreement, but otherwise it's good to go!

PR #2479, opened 2024-09-12
Update octopus-cli create release with new parameter

APErebus:
The `--git-resource` parameter is new. Can only be merged _after_ https://github.com/OctopusDeploy/cli/pull/392 has been merged and released

Shortcut story: [sc-69100]

PR #2596, opened 2024-12-16
Remove --git-resource from deprecated CLI create release command

APErebus:
Was erroneously added in #2479.

This was not added to the deprecated CLI, just the new CLI which has it's docs auto-updated


flin-8 (approved):
LGTM

PR #2608, opened 2025-01-10
Update Astro to resolve redirect bug

steve-fenton-octopus:
Astro 5.0.3 had a bug treating frontmatter as content.


tonykelly-octopus (approved):
LGTM

PR #2613, opened 2025-01-22
Cater for multi-line link labels in right sidenav

iryan2:
# Before
Screenshot 2025-01-22 at 4 32 51 pm

# After 
Screenshot 2025-01-22 at 4 27 17 pm


david-wu-octopus (approved):
Looks much better now!

PR #2617, opened 2025-01-29
Update build info docs to callout maven feed groupId & packageId requ…

IsaacCalligeros95:
A couple of customers have run into issues trying to get build information for maven feeds to display. This is usually because the groupId is not included. This is a Maven feed specific requirement that calling out should help customers diagnose and resolve this problem. 

[Internal thread](https://octopusdeploy.slack.com/archives/CNHBHV2BX/p1738079840110819)


hnrkndrssn (approved):
LGTM

benPearce1 (approved):

PR #2618, opened 2025-01-30
Add doc for system variable OctopusShouldFailDeploymentOnSubstitutionFails

HuyPhanNguyen:
Add doc for system variable OctopusShouldFailDeploymentOnSubstitutionFails which allow user control if the deployment should fail when variable substitution fail.

### DIFF
--- a/.github/workflows/branch-validation.yml
+++ b/.github/workflows/branch-validation.yml
@@ -87,7 +87,7 @@ jobs:
           packages: ${{ steps.package.outputs.package_file_path }}
 
       - name: Let people download package
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


src/pages/docs/projects/variables/system-variables.md
```diff
@@ -1418,6 +1418,13 @@ The amount (in megabytes) of available free disk space we should check for (over
 
 Example: 100
 
+
+`OctopusShouldFailDeploymentOnSubstitutionFails`
+
+If set to "True", the deployment will fail if any variable substitution fails.
```
benPearce1, line 1424:
```suggestion
If set to "True", the deployment will fail if any variable substitution fails. This variable was added in Octopus 2025.1.0.
```

benPearce1 (approved):

PR #2627, opened 2025-02-14
Project Triggers documentation updates

ellen-octopus:
PGT will be running a test ([What’s after activation?](https://whimsical.com/pitch-what-s-after-activation-Uqg4MSx7KHiAhgMDAHhWZR)) where we link users to docs for features they should try after completing their first successful deployment. One of the features we might link to is triggers, so we'd like to give the docs a small update. 

- Updated the doc to align with our in-product description, which better describes what triggers do. 
- Updated and simplified the image example.
- Added doc links to each trigger.
- Removed the 'comprehensive guide' link because it's not actually a guide, but a contents page for scenarios. In its current format, it's more likely to cause confusion than provide value.

## Before

CleanShot 2025-02-14 at 14 21 26@2x

## After

CleanShot 2025-02-14 at 14 31 26@2x


steve-fenton-octopus (approved):
LGTM!

PR #2628, opened 2025-02-14
Document need for care when deleting resources used by Git Runbooks

caitlynstocker:
## Background 🌵 

When a user tries to delete an environment used by a DB Runbook, there are guardrails in place to prevent this. However with CaC Runbooks we can't maintain these guardrails as it's not practical to check every commit in a customer's repo to be sure there are no versions of their runbooks referencing a given environment.

There is a warning about this in the Unsupported Config as Code Scenarios page of the docs, but it's quite hidden away and we would like this info to be more readily available to CaC Runbooks users. 

## What's this?

This change just adds a couple of sentences to the CaC Runbooks doc warning against deleting required resources.  And repeats a similar warning in the docs for channels, environments, lifecycles, tenant tags, worker pools and step templates.

## What's it look like?
image

Fixes[sc-103829]


src/pages/docs/runbooks/config-as-code-runbooks.md
```diff
@@ -87,3 +87,9 @@ The information that was previously found on the **Snapshot** page is still avai
 If you use automated scripts that run runbooks via the Octopus Server API and you convert your runbooks to Config As Code the URL for the runbook will change to include a branch reference (e.g. `refs/heads/main`) as a result you need to update your scripts to include the branch reference where the runbook is stored. 
 
 - [PowerShell example](https://github.com/OctopusDeploy/OctopusDeploy-Api/blob/master/REST/PowerShell/Runbooks/RunConfigAsCodeRunbook.ps1)
+
+## Deleting required resources
+
+Once your Runbooks are version controlled, it's up to you to take care to avoid deleting any Octopus resources required by your Runbooks. See our [core design decisions](docs\projects\version-control\unsupported-config-as-code-scenarios#core-design-decisions) for more information.
```
steve-fenton-octopus, line 93:
```suggestion
Once your Runbooks are version controlled, it's up to you to take care to avoid deleting any Octopus resources required by your Runbooks. See our [core design decisions](/docs/projects/version-control/unsupported-config-as-code-scenarios#core-design-decision) for more information. 
```

PR #2636, opened 2025-02-21
Scoped Tenant Variables

susanpann:
[sc-93473]


tothegills (approved):
Looks good for now. We'll need to update the screenshots when we go GA. Can you keep track of that somewhere, maybe in the rollout plan?

PR #2649, opened 2025-03-20
Style for horizontal rule

steve-fenton-octopus:
You can insert a horizontal rule in markdown using the syntax:

```
A horizontal rule is added below, using three dashes:

---
```

This PR styles it.


tonykelly-octopus (approved):
LGTM

PR #2653, opened 2025-03-24
Add note about different vsc providers using different repo URL formats

hnrkndrssn:
[sc-104369]

### DIFF
--- a/src/shared-content/version-control-repository-url-format.include.md
+++ b/src/shared-content/version-control-repository-url-format.include.md
@@ -1,3 +1,5 @@
 :::div{.hint}
-Different VCS providers require different URL formats for the Git repository, some (e.g. GitLab) require the URL include `.git` at the end while others (e.g. Azure DevOps) does not support this. The best option to get the correct URL is to go to the repository in your provider and copy the URL used for cloning the repository from there.
+Different VCS providers require different URL formats for the Git repository, some (e.g. GitLab) require the URL include `.git` at the end while others (e.g. Azure DevOps) does not support this. GitHub supports either format.
+
+The best option to get the correct URL is to go to the repository in your provider and copy the URL used for cloning the repository from there.
 :::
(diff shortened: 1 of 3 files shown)


src/shared-content/version-control-repository-url-format.include.md
```diff
@@ -0,0 +1,3 @@
+:::div{.hint}
+Different VCS providers require different URL formats for the Git repository, some (e.g. GitLab) require the URL include `.git` at the end while others (e.g. Azure DevOps) does not support this. The best option to get the correct URL is to go to the repository in your provider and copy the URL used for cloning the repository from there.
+:::
```
hnrkndrssn, line 3: Should we include links to docs on how to get the URL for cloning for our most used VCS providers, i.e. GitHub, GitLab, and ADOps...?

IsaacCalligeros95: I'm on the fence, calling out the .git postfix and that Gitlab requires it and AzureDevops does not is the main piece of information and I'm not sure if the docs would add anything relevant that a reader would be looking for. It might be good to explicitly call out where GitHub falls (I think both are valid?)

IsaacCalligeros95 (approved):
LGTM, minor comment which is just my personal preference on being explicit about saying where GitHub falls.

src/shared-content/version-control-repository-url-format.include.md
hnrkndrssn: > It might be good to explicitly call out where GitHub falls (I think both are valid?)

Correct, GitHub supports both formats.

hnrkndrssn: > I'm on the fence

as am I...it doesn't feel like it adds anything relevant but thought it worth bringing up nonetheless :)

IsaacCalligeros95: I think we just leave them out.

IsaacCalligeros95 (approved):
LGTM

src/shared-content/version-control-repository-url-format.include.md
hnrkndrssn: We have pretty good [proof](https://github.com/OctopusDeploy/Issues/issues/9267) that GitLab **does not** work with our implementation (not sure what the difference would be) in Octopus when not having `.git`. 

My guess is the `warning: redirecting to https://gitlab.com//cac-test.git` has something to do with it.

PR #2656, opened 2025-03-28
Docs for AWS ECR Oidc feeds

IsaacCalligeros95:
Updates ECR docs to include configuration details for Oidc.


benPearce1 (approved):
That reads nicely. Thanks

steve-fenton-octopus (dismissed):
Hello! Thanks for these great changes. They are so good, we should signal to search engines there has been an update. To do this, update the `modDate` in the frontmatter.

PR #2658, opened 2025-04-01
Update Uptime SLO stats for March 2025

adrianparker:
Updated stats. Gave sentence about fewer than 9 minutes of total downtime per month a little more prominence.


pawelpabich (approved):
LGTM

PR #2661, opened 2025-04-03
Update export/import and migration docs to include CaC projects

hnrkndrssn:
[sc-107306]

### DIFF
--- a/src/pages/docs/octopus-cloud/migrations.md
+++ b/src/pages/docs/octopus-cloud/migrations.md
@@ -53,7 +53,7 @@ Use this checklist to guide you as to the complexity of your migration. The more
 | Do you have specific security requirements or policies around accessing deployment targets or managing API keys? | You will need to manage these through the migration.  |
 | Do you need to retain historical data and task logs? | Historical data and task logs are **not supported by Export/Import Projects**. You will need to continue to host the older version to retain historical data. |
 | How long are your Audit log retention policies? | Octopus Cloud archived audit logs has **a max of 365 days** |
-| Do you have any Projects using Config-as-Code? | Config-as-Code Projects requires Octopus **2025.2.x** |
+| Do you have any Projects using Config-as-Code? | Config-as-Code Projects requires Octopus **2025.2.5601** |
 | Do you store build information?  | You will need to continue to host the older version to retain historical data, as build information **is not migrated**. |
 | Do you have any subscriptions? | Subscription migration is **not supported by Export/Import Projects**. These will need to be migrated manually or with a script. |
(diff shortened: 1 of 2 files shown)


src/pages/docs/octopus-cloud/migrations.md
```diff
@@ -53,7 +53,7 @@ Use this checklist to guide you as to the complexity of your migration. The more
 | Do you have specific security requirements or policies around accessing deployment targets or managing API keys? | You will need to manage these through the migration.  |
 | Do you need to retain historical data and task logs? | Historical data and task logs are **not supported by Export/Import Projects**. You will need to continue to host the older version to retain historical data. |
 | How long are your Audit log retention policies? | Octopus Cloud archived audit logs has **a max of 365 days** |
-| Do you have any Projects using Config-as-Code? | Config-as-Code Projects are **not supported by Export/Import Projects** |
+| Do you have any Projects using Config-as-Code? | Support for Config-as-Code Projects was added in Octopus **2025.2.5601** |
```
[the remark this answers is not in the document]
hnrkndrssn: Yea, we have a selection of different conventions around this it seems. Some places uses `major.minor` others `major.minor.patch`...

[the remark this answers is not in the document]
benPearce1: might be a question for the topic-docs channel

src/pages/docs/projects/export-import/index.mdx
```diff
@@ -207,3 +246,7 @@ You can use the [Octopus REST API](/docs/octopus-rest-api) to export or import O
 To find out more take a look at our examples:
 - [Export projects](/docs/octopus-rest-api/examples/projects/export-projects)
 - [Import projects](/docs/octopus-rest-api/examples/projects/import-projects)
+
+## Older versions
+
+- Prior to version **2025.2.5601**, version-controlled projects were not supported by the **Export/Import Projects** feature.
```
benPearce1, line 252: I am no expert at English, it feels like this could have the logic reversed:

```suggestion
- Importing and Exporting of version-controlled projects is only supported from version **2025.2.5601**
```

benPearce1 (approved):
This looks good, I am not sure of the best approach for stating the supported version: prior to version vs supported from this version

src/pages/docs/projects/export-import/index.mdx
hnrkndrssn: I agree that it feels backward when you read it, but it's the logical process you go through though. And we use that order elsewhere in these docs.

benPearce1: If it is used elsewhere then leave it as is

hnrkndrssn:
> This looks good, I am not sure of the best approach for stating the supported version: prior to version vs supported from this version

The [example](https://octopus.com/docs/projects/version-control/config-as-code-reference#older-versions) I was given uses `Prior to version...` so I went with that. 🤷

PR #2663, opened 2025-04-03
Added hidden page for OctoAI

mcasperson:
This PR adds a hidden page that we will share with select customers who are trialing OctoAI.

### DIFF
--- a/src/pages/docs/administration/octoai/index.md
+++ b/src/pages/docs/administration/octoai/index.md
@@ -1,0 +1,72 @@
+---
+layout: src/layouts/Default.astro
+pubDate: 2025-04-04
+modDate: 2025-04-04
+title: OctoAI
+description: How to use OctoAI
+navOrder: 100
+hideInThisSection: true
+---
+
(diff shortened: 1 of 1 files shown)


robpearson (approved):
LGTM!

src/pages/docs/administration/octoai/index.md
```diff
@@ -0,0 +1,86 @@
+---
+layout: src/layouts/Default.astro
+pubDate: 2025-04-04
+modDate: 2025-04-04
+title: OctoAI
+description: How to use OctoAI
+navOrder: 100
+hideInThisSection: true
```
steve-fenton-octopus, line 8: `hideInThisSection` should be replaced with:

```suggestion
navSearch: false
navSitemap: false
navMenu: false
```

This is granular control over specific types of visibility.

steve-fenton-octopus: @mcasperson just an FYI that ⬆️ is what you'll need if you want the page to be hidden.

PR #2688, opened 2025-04-28
Adding official support for Windows Server 2025 for Octopus Server

MarkSiedle:
Our installation requirements do not yet list Windows Server 2025 in our list of supported platforms for Octopus Server. This PR updates that.

While here, we're also modifying the wording to clarify that we do not guarantee automatic testing any of these platforms (which is not sustainable based on the number of combinations of platforms/agents we'd need to setup), only that we support them.


dictionary-octopus.txt
```diff
@@ -357,6 +357,7 @@ reprovisioned
 reprovisioning
 resolv.conf
 ResolvConfPath
+RHEV
```
borland, line 360: What's a RHEV?

borland (approved):

dictionary-octopus.txt
MarkSiedle: Red Hat Enterprise Virtualization. It's mentioned in the existing docs, but never got captured in spell check previously.

steve-fenton-octopus (approved):

PR #2697, opened 2025-05-12
Update Access Control naming to match CCv2

egorpavlikhin:
Renames Access Control to User Access and Admin Access (for server subscriptions).

[sc-109701]


steve-fenton-octopus (approved):

erinb-octopus (approved):
LGTM

PR #2699, opened 2025-05-12
Add Docker image package variables

harrisonmeister:
This adds variable definitions already available in our docs here - https://octopus.com/docs/deployments/custom-scripts/run-a-script-step#docker-image-package-variables


tonykelly-octopus (approved):
LGTM

PR #2705, opened 2025-05-15
Add warning around Safari and third-party cookies in an iframe

hnrkndrssn:
image

[sc-110184]


src/pages/docs/releases/issue-tracking/jira.md
```diff
@@ -85,6 +85,12 @@ This process is for Jira Cloud, if you are using Jira Server, see [Connecting Ji
 
     From the Atlassian Marketplace, add the [Octopus Deploy for Jira](https://marketplace.atlassian.com/apps/1220376/octopus-deploy-for-jira) app and click 'Get Started' to configure it. Alternately, the app is also available in Jira by navigating to **Jira Settings ➜ Find new apps**.
 
+    :::div{.warning}
+    **Safari and third-party cookies in an `iframe`**
+
+    Please note: The Octopus Deploy for Jira plugin uses cookies and Safari, by default, discards cookies set in an `iframe` unless the host that's serving the `iframe` has set a cookie before, outside the `iframe` which in our case does not happen. Attempting to configure our plugin using Safari will fail with a blank screen and a HTTP 400 error in the network tab, due to this limitation we recommend using a different browser that allows third-party cookies in an `iframe`.
```
benPearce1, line 91:
```suggestion
    Please note: The Octopus Deploy for Jira plugin uses cookies, and Safari, by default, discards cookies set in an `iframe` unless the host that's serving the `iframe` has set a cookie before, outside the `iframe`. In our case, this does not happen. As a result, attempting to configure our plugin using Safari will fail with a blank screen and a HTTP 400 error in the network tab. Due to this limitation, we recommend using a different browser that allows third-party cookies in an `iframe`.
```

benPearce1 (changes requested):
"Breaking up the wall-of-text" suggstion

benPearce1 (approved):

PR #2715, opened 2025-05-16
Minor update to fix Markdown anchor for feeds

twerthi:
The feeds anchor had a space init making it so the markdown didn't interpret it as an achor.


steve-fenton-octopus:
@twerthi good spot!

PR #2718, opened 2025-05-20
Update index.md

zentron:
Based on some [updated plans around deprecations](https://octopushq.atlassian.net/wiki/spaces/RND/pages/2500493448/Managing+Deprecations) we will now instead consider the removal at 6 moths rather than 12 months as standard practice.

### DIFF
--- a/src/pages/docs/deprecations/index.md
+++ b/src/pages/docs/deprecations/index.md
@@ -13,9 +13,9 @@ Occasionally, Octopus will deprecate features that will no longer be supported. 
 
 Deprecations have the following lifecycle:
 
-- Announce deprecation
-- (+6 months) Toggle off deprecated functionality
-- (+1 year) Remove deprecated functionality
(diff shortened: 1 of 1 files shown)


src/pages/docs/deprecations/index.md
```diff
@@ -13,9 +13,9 @@ Occasionally, Octopus will deprecate features that will no longer be supported.
 
 Deprecations have the following lifecycle:
 
-- Announce deprecation
-- (+6 months) Toggle off deprecated functionality
-- (+1 year) Remove deprecated functionality
+- Announce deprecation with in-app warnings, newsletters and/or blog posts
+- (+6 months) Disable deprecated functionality
```
steve-fenton-octopus, line 17: We try to avoid charged terms, will this work instead?

```suggestion
- (+6 months) Switch off deprecated functionality
```

steve-fenton-octopus (changes requested):
I left a suggestion on one small change.

PR #2725, opened 2025-05-23
Updates to some prior docs I changed

Clare-Octopus:
Swapped some pictures around so they matched their titles and added the integrity check to one upgrade doc I missed.


AdamHollow49 (approved):
LGTM

PR #2737, opened 2025-06-04
Clarification of supported characters in variable names

felixi-octopus:
[sc-96109]

# Background
There's a lack of clarity over what kind of output variable names are supported with Octostache.

# Results
Add explanation of the kind of variable naming that will work with Octostache.

### DIFF
--- a/src/pages/docs/projects/variables/getting-started.mdx
+++ b/src/pages/docs/projects/variables/getting-started.mdx
@@ -163,7 +163,27 @@ If you are using the [Structured configuration variables](/docs/projects/steps/c
 ## Variable Recommendations
 
 ### Use standard characters in variable names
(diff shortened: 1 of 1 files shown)


FinnianDempsey:
Hey @felixi-octopus 

Sorry to me it's not quite clear how the proposed doc changes relate to the other docs that indicate variables can use the `##{` syntax to not be evaluated. 

This also seems to imply that you shouldn't use the `#{}` syntax for _any_ variable names:
> For broader compatibility, including support for Octostache expressions, stick to alphanumerics, periods (`.`), hyphens (`-`), underscores (`_`), and spaces. Octostache does not support characters outside this set, so following this convention ensures your variables function reliably across all usage contexts.

FinnianDempsey (approved):
Confirmed that variables accessed via #{} syntax don't allow for special characters and it needs to use get_octopusvariable to resolve the value

PR #2752, opened 2025-06-17
Update index.mdx

donnybell:


Justin-Walsh (commented):
While I appreciate the intention and driver for this change - earlier or equal version is how it's supposed to work. For scenarios where it isn't, we should escalate it to RND.

PR #2758, opened 2025-07-01
fix: updated image and meta props

jbristowe:
- replaced image with updated UX for Octopus AI Assistant


tonykelly-octopus (approved):
LGTM

PR #2771, opened 2025-07-17
Update dependencies

steve-fenton-octopus:
Should resolve Dependabot alerts and get everything up to date.

2x dependabot alerts


KMajkrzakOctopus (approved):
👌

PR #2781, opened 2025-07-29
Adding additional information for the Octopus AI Assistant to work wi…

twerthi:
…th self-hosted instances

Current documentation makes it look like all you need is to allow the IP address in, but you also need the Azure Function to make a callback to the Octopus Server.  This means that the DNS entry of your self-hosted server needs to be resolvable via the Internet.


tonykelly-octopus (approved):
LGTM

PR #2787, opened 2025-08-07
Update dynamic worker docs

adrianparker:
Make it more explicit that customers can lease one VM of each type, and can lease one of each type concurrently.


adrianparker:
Forgot to nudge front matter.

PR #2788, opened 2025-08-07
Update dynamic worker docs

adrianparker:
Make it more explicit that customers can lease one VM of each type, and can do so concurrently.


steve-fenton-octopus (approved):
LGTM

PR #2796, opened 2025-08-20
Update subscription events documentation

wlthomson:
A few additions and tidy ups to bring our subscription docs up-to-date.


src/pages/docs/administration/managing-infrastructure/subscriptions/index.md
```diff
@@ -306,3 +311,5 @@ The following is a list of all options for subscriptions including Event Groups,
 \* License key expiry events are only available in Octopus Server version `2025.2.6402` and later.
```
wlthomson, line 311: Not sure at what point we should consider removing these footnotes? Now that we are on `2025.3`, is it worth considering doing away with them?

benPearce1: For now I would leave it as is. 2025.2 is still LTS for the rest of the year

PR #2805, opened 2025-09-03
Resolve PNPM issue

steve-fenton-octopus:
When running pnpm dev the following error happens (depending on a few environmental factors, so it's possible you would have just the right scenario for it to work).

/Users/fenton/repos/docs/node_modules/.bin/../@pnpm/exe/pnpm: line 1: This: command not found
 ELIFECYCLE  Command failed.

Explanation

This error is caused by a loss of context in the command chain.

pnpm dev -> npm-run-all -> dev:watch -> pnpm watch

The pnpm watch call inside dev:watch has lost the local pnpm instance from its PATH.

On the call we fixed this by removing the local pnpm dependency, but this is a useful way to make sure the right version on pnpm is used for each project (as it may differ) so we needed a solution that resolved the loss of PATH.

Solution

I've tested this out locally and if we remove the hop from dev:watch -> pnpm watch it should resolve the issue.


tonykelly-octopus (approved):
LGTM

PR #2813, opened 2025-09-08
Troubleshooting tentacles rejig

calverp:
Re-organisation of the tentacle troubleshooting docs to make them easier for users to follow


src/pages/docs/infrastructure/deployment-targets/tentacle/troubleshooting-tentacles.md
```diff
@@ -1,268 +0,0 @@
----
```
steve-fenton-octopus, line 1: We usually keep the markdown file, but turn it into a redirect to an alternative page.

See: https://github.com/OctopusDeploy/docs?tab=readme-ov-file#redirects

steve-fenton-octopus (changes requested):
See the inline comment about redirects and redirect chains. Let me know if you need help.

PR #2842, opened 2025-09-24
Experimental "Copy Page" button

egorpavlikhin:


steve-fenton-octopus (changes requested):
See note on markdown availability.

steve-fenton-octopus:
I'll close this PR for now. Get in touch if you have a strong use case for adding this.

PR #2843, opened 2025-09-26
Add new process template and relative positioning examples

shaun-od:
The Policy examples do not currently contain the following examples.
- Process Templates
- Relative step position checking 

This PR adds those examples.


shaun-od:
@Vvasudevan-octopus can you confirm these are the examples you wanted and look good before we merge.

Vvasudevan-octopus (approved):
Examples look good to me.

PR #2848, opened 2025-10-02
Added troubleshooting and versioning info to Gateway doc

tleed5:


APErebus (approved):
awesome! Thanks @tleed5

PR #2855, opened 2025-10-06
ArgoCD: Add default registry documentation

rain-on:
Added a new page, and referenced it from the Image-tag step.


Jtango18 (approved):

src/pages/docs/argo-cd/steps/update-application-image-tags.md
```diff
@@ -18,6 +18,12 @@ ensure the correct sections of the relevant values files are updated. See [Helm
 more information.
 :::
 
+
+:::div{.info}
+If your cluster's default registry has been overridden see [cluster annotations](/docs/argo-cd/annotations/cluster-annotations) to ensure
```
APErebus, line 23:
```suggestion
If the application cluster's default registry has been changed see [cluster annotations](/docs/argo-cd/annotations/cluster-annotations) to ensure
```

src/pages/docs/argo-cd/annotations/cluster-annotations.md
```diff
@@ -0,0 +1,21 @@
...
+to override this setting.
+
+By leaning into the default registry - your Kubernetes yaml becomes simpler, as you can nominate the image of interest
+without needing to prefix with the registry.
+
+While your default registry is `docker.io`, all Argo based steps will continue to function as expected, however, if you
+change your default registry, this change must be communicated with Octopus.
+
+This can be done by adding an annotation to the cluster object in your Argo CD instance:
+`argo.octopus.com/default-container-registry : <registry hostname>`.
```
APErebus, line 21:
```suggestion
By default, a Kubernetes cluster will use `docker.io` as the image registry when none is defined. This means that if an image is defined without the registry, for example `nginx/nginx:latest`, then this image will be resolved from the `docker.io` registry.

This default registry can be changed via the [kubeadm init --image-repository](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/) command & flag.

When executing the [Update Argo CD Image Tags](/docs/argo-cd/steps/update-application-image-tags) step, Octopus will exactly match the specified image using the registry and name.

If the default cluster version has been changed, then Octopus needs to be made aware so that it correctly matches images.

:::div{.info}
If the cluster default registry has not been changed, then you do not need to do the following steps.
:::

## Setting the default container registry annotation

To make Octopus aware of a clusters default registry, an annotation needs to be added to each cluster object in Argo CD instance.

The annotation and value is:

`argo.octopus.com/default-container-registry : <registry hostname>`.

example:

`argo.octopus.com/default-container-registry : my-company-registry.com`.

IMAGE GOES HERE

```

APErebus (changes requested):
The language needs a bit cleaning up and I think an image showing where in the UI the annotation goes on the cluster would be helpful

PR #2870, opened 2025-10-16
remove bad file for argo

rain-on:
We left some cruft files behind.


NickJosevski:
should that path 
   `https://octopus.com/docs/infrastructure/argo-cd`

 now redirect to 
 
  `https://octopus.com/docs/argo-cd` 
  
  ?

PR #2874, opened 2025-10-20
Add callout for runbook triggers and pre-release package versions

hnrkndrssn:
Scheduled triggers for git runbooks only support finding non-prerelease package versions so I'm adding a callout to our docs to make it clear that runbook triggers won't find prerelease packages and that you'd have to hard-code the version if you want to use them.

[sc-124157]


grace-rehn (approved):

HuyPhanNguyen (approved):
LGTM

PR #2886, opened 2025-10-24
Add troubleshooting docs for runbook migrations to cac

grace-rehn:
Customer ran into issues migrating their runbooks due to duplicate/empty slugs in their runbook snapshots. This was likely due to them having published runbooks from before slugs were introduced.

This was not able to be fixed in code as duplicate slugs meant that we could not determine the `processOrder`.

Added docs to help troubleshoot slug related errors during runbook migrations. Included 2 scripts from support that help identify the problematic runbook steps.

[sc-120525]


hnrkndrssn (approved):
LGTM 👍

PR #2894, opened 2025-10-29
Update HTTP to HTTPS for on-premises setup

harrisonmeister:
Customer raised this term (HTTP) as being confusing, as the request is made over HTTPS, so changed the term to avoid confusion.


samwld (approved):
GTG

PR #2896, opened 2025-10-30
Don't update old images.

steve-fenton-octopus:
As docs is image heavy, let's not auto update old generated images.


samwld (commented):
GTG

PR #3094, opened 2026-04-13
Add information to GitHub Connections for allow lists

stevencl840:
# Context 🌇 
We have updated the GitHub App so that all traffic from the App to GitHub now passes through a known IP. This allows customers using GitHub Enterprise to enable allow lists and, by adding this IP, still use the GitHub App to manage connections to their repositories.

chrome_5NfT9Nqo0H

This PR will closes devex-5

### DIFF
--- a/src/pages/docs/projects/version-control/github/index.md
+++ b/src/pages/docs/projects/version-control/github/index.md
@@ -81,7 +81,7 @@ Whenever possible, Octopus uses a token scoped down to minimal permissions in ac
 
 ## Allow List
 
-The Octopus Deploy GitHub App can be used with the GitHub Enterprise allow list feature. To include the app in your allow list either enable "Enable IP allow list configuration for installed GitHub Apps" in the "Authentication security" section, or manually add the IP Address "172.182.208.68" to your allow list.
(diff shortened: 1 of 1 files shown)


src/pages/docs/projects/version-control/github/index.md
```diff
@@ -79,6 +79,15 @@ There are specific GitHub permissions that the Octopus GitHub App requests in or
 
 Whenever possible, Octopus uses a token scoped down to minimal permissions in accordance with the principle of least privilege.
 
+## Allow List
+
+The Octopus Deploy GitHub App can be used with the GitHub Enterprise allow list feature. To include the app in your allow list manually add the IP Address `172.182.208.68`. Due to a limitation in the way that GitHub supports inheritance of IP addresses when performing actions on behalf of a user, the IP address for the GitHub App needs to be configured manually and cannot be inherited from the app settings. For more information please refer to [GitHub's Documentation](https://docs.github.com/en/enterprise-cloud@latest/apps/maintaining-github-apps/managing-allowed-ip-addresses-for-a-github-app#about-ip-address-allow-lists-for-github-apps)
```
geofflamrock: I think I would probably put the bit about the limitation in a separate paragraph, and maybe as some sort of information note like the one below? Just to differentiate it from the main content which is the IP address.

geofflamrock: A few other thoughts:
- Can we add a link to the GitHub docs on how to add the IP address to an enterprise or org? 
- Is "GitHub Enterprise allow list" the name of the feature in GitHub? Might be worth checking what they call it and adjusting if needed.

stevencl840: Having two information blocks in that close of proximity made it look weird (in my opinion). I have split it into two paragraphs and placed the info block between them. I think breaking it up actually draws more attention to the second paragraph

stevencl840: * I have added the link to githubs documentation
* They refer to it as  allow  list, or allow list IP. I have removed the reference to enterprise

geofflamrock (approved):

PR #3098, opened 2026-04-13
Add missing vars for argocd instances tf

evolutionise:
Ran this terraform recently - these variables can be used if using the helm chart for the argo cd gateway, and are already included in the tfvars, just needed to be mapped using the `variables.tf`.


rain-on (approved):
LGTM

PR #3099, opened 2026-04-14
Update username password authentication with limitations

grace-rehn:
Updates the username and password authentication method with some security limitations and point to idp and directory based auth as a recommendation.

Fixes HPY-1345


hnrkndrssn (approved):
LGTM 👍

PR #3100, opened 2026-04-14
Fix incomplete argo manifest

wlthomson:
Fixes FD-212.

Step 4 of the docs for "Installing an Argo CD Application" includes a yaml snippet for a manifest spec, sans the outer envelope and parent `spec:` block.

The wording of the section was originally "The application yaml required to install the helm chart is as follows," but was recently changed to the current phrasing (https://github.com/OctopusDeploy/docs/pull/3043), which includes an optional instruction to "...commit this manifest to your git-ops repository."

This change updates the yaml block to include the full K8 resource envelope.

### DIFF
--- a/src/pages/docs/argo-cd/instances/automated-installation.md
+++ b/src/pages/docs/argo-cd/instances/automated-installation.md
@@ -142,31 +142,38 @@ The application YAML required to install the helm chart is as follows (replacing
 4. Apply the Argo CD application (or commit this manifest to your git-ops repository already synced by Argo CD)
 
     ```yaml
-    project: default
-    source:
-      repoURL: registry-1.docker.io/octopusdeploy
-      chart: octopus-argocd-gateway-chart
-      targetRevision: 1.23.0
-      helm:
-        valuesObject:
-          registration:
-            octopus:
-              name: <display name of gateway in Octopus>
-              serverApiUrl: https://your-instance.octopus.app
-              serverAccessTokenSecretName: octopus-server-access-token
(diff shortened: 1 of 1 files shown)


ATGardner:
that's a good point. thanks for spotting this.
i am not entirely sure, but we might also want to add
```yaml
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
      allowEmpty: false

    syncOptions:
      - PrunePropagationPolicy=foreground
      - Replace=false
      - PruneLast=false
      - Validate=true
      - CreateNamespace=false
      - ApplyOutOfSyncOnly=false
      - RespectIgnoreDifferences=false
```
to the spec.
it might be a bit too opinionated (assuming the user wants auto-sync with pruning and self-heal). so we can also leave it out, and assume the user knows enough of argo-cd to set up their own preferences on the application.

ATGardner (approved):

wlthomson:
> that's a good point. thanks for spotting this. i am not entirely sure, but we might also want to add
> 
> ```yaml
>   syncPolicy:
>     automated:
>       prune: true
>       selfHeal: true
>       allowEmpty: false
> 
>     syncOptions:
>       - PrunePropagationPolicy=foreground
>       - Replace=false
>       - PruneLast=false
>       - Validate=true
>       - CreateNamespace=false
>       - ApplyOutOfSyncOnly=false
>       - RespectIgnoreDifferences=false
> ```
> 
> to the spec. it might be a bit too opinionated (assuming the user wants auto-sync with pruning and self-heal). so we can also leave it out, and assume the user knows enough of argo-cd to set up their own preferences on the application.

I'll leave this one for as a follow up for you guys to decide 👍 .

PR #3110, opened 2026-04-17
Update search.json.ts and improve dot-handling

steve-fenton-octopus:
Updates search.json.ts to the most current Accelerator file.

Changes dot-handling:

- Before: octopus.server.exe -> octopusserverexe
- After: octopus.server.exe -> octopus server exe

This takes advantage of the fact that matching "octopus.server.exe" on a page is effectively 3 search query terms matched.


tonykelly-octopus (approved):
lgtm

PR #3112, opened 2026-04-20
Restructure feature toggle documentation

dylanlerch:
Feature toggle documentation is now broken into 3 sections:

- Overview: Introduction, getting started, and misc (default values).
- Targeting: Targeting controls within an environment.
- Providers: List of provider libraries.

There are minimal content changes in here (other than adding a few icons and moving a few lines around where it made sense after the restructure). I'll be following this up with some other PRs to add content for all of the new features that we have added over the past few months.


steve-fenton-octopus (approved):
LGTM!

PR #3116, opened 2026-04-27
Add scoping information for script module permissions

bec-callow-oct:
Script modules are stored as unscoped variable sets so they cannot be accessed by users with environment scoped permissions. This information is missing from the current documentation.

Fixes: FD-249

### DIFF
--- a/src/pages/docs/deployments/custom-scripts/script-modules.md
+++ b/src/pages/docs/deployments/custom-scripts/script-modules.md
@@ -240,6 +240,8 @@ ramifications of changing a Script Module, as well as track down where it has be
 
 ## Permissions
 
+Script modules are represented as unscoped variable sets. This means that the user role must not be scoped to specific environments. 
+
 To be able to access and use the script modules feature you will require the following permissions:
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/deployments/custom-scripts/script-modules.md
```diff
@@ -235,11 +245,13 @@ the Script Module, as well as the Releases that have a snapshot of the Script Mo
 ramifications of changing a Script Module, as well as track down where it has been used in the past.
 
 :::figure
-![](/docs/img/deployments/custom-scripts/images/script-modules-usage.png)
+![Script Modules Usage](/docs/img/deployments/custom-scripts/images/script-modules-usage.png)
 :::
 
 ## Permissions
 
+Script modules are represented as unscoped variable sets. This means that the user role must not be scoped to specific environments.
```
benPearce1, line 253: It feels like this could be a bug, but for the moment, this might be the right callout.

benPearce1 (approved):
Thanks for the cleanup too
👍

PR #3118, opened 2026-04-27
Document OIDC subject claim behaviour

sathvikkumar-octo:
## Summary

When a selected OIDC subject key has no value at runtime (e.g. `tenant` on an untenanted deployment), both the key and the value are dropped from the generated `sub` claim — not just the value. The docs had a format that included a `tenant:` segment that isn't generated for untenanted deployments. Reported internally after an AWS IAM trust policy was configured against the documented format and didn't match the actual subject.

This PR:

- Documents the behaviour around what key/values are generated
- Shows both the tenanted and untenanted resulting subjects, plus a matched example pair on the AWS account page
- Adds the missing `Project Group` key from the subject generator
- Removes `Feed` from the Deployments and Runbooks allowed parts (Feed has its own subject config)

Source of truth: [`OidcSubjectGenerator.cs`](https://github.com/OctopusDeploy/OctopusDeploy/blob/main/source/Octopus.Core/Authentication/OidcSubjectGenerator.cs)

### DIFF
--- a/src/pages/docs/infrastructure/accounts/aws/index.md
+++ b/src/pages/docs/infrastructure/accounts/aws/index.md
@@ -1,7 +1,7 @@
 ---
 layout: src/layouts/Default.astro
 pubDate: 2023-01-01
-modDate: 2025-08-18
+modDate: 2026-04-27
 title: AWS accounts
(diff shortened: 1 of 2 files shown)


steve-fenton-octopus (approved):
LGTM!

PR #3120, opened 2026-04-29
Document NFS storage deprecation in Kubernetes Agent

APErebus:
Added information about the deprecation of bundled NFS storage in Kubernetes Agent v3, including future removal plans and guidance for existing users.


flin-8 (approved):
LGTM

PR #3122, opened 2026-04-30
Tweak Kubernetes agent docs to support v3

APErebus:
We have released the Kubernetes agent v3 which has a slightly different installation dialog so updated some images and tweaked the language.

Apparently `run dev` also regenerated a stack of other images 😠 

image

image

image

image

image

image

Fixes MD-1565


steve-fenton-octopus (approved):
LGTM!

PR #3135, opened 2026-05-07
Octopus Deploy Project configuration in Deploy instance

mattallford:
This PR includes OCL for the configuration of a new project for the docs, in the Deploy instance.

We will be migrating deployments from tenpillars to deploy. These OCL files are in a subdirectory so they don't interfere with the existing OCL for the project in tenpillars.


jbristowe (approved):
LGTM

PR #3141, opened 2026-05-11
docs(argo-cd): publish HelmReplacementPath OCL property + clarify valueFiles handling for Update Image Tags step

vlussenburg:
Two gaps in the [Update Argo CD Application Image Tags](https://octopus.com/docs/argo-cd/steps/update-application-image-tags) doc that bit me on a CaC project.

**The OCL property name for "Helm image tag path" isn't documented.** The UI field is described, but if you're editing `deployment_process.ocl` by hand you need to know it serialises to `HelmReplacementPath` on the package reference (resolved deployment-time as `Octopus.Action.Package[<ref>].HelmReplacementPath`). I found this by reading [`PackageVariables.HelmReplacementPath`](https://github.com/OctopusDeploy/Calamari/blob/main/source/Calamari.Common/Plumbing/Variables/PackageVariables.cs) in Calamari. The OCL save silently strips unknown property names, so guessing `HelmImageTagPath` (or similar) gives no error — the step just runs as a no-op forever. Added an OCL example.

**The "For Helm charts" section reads as if only the chart's default `values.yaml` is touched.** In practice, when `HelmReplacementPath` is set, the step also writes to every file listed in the Application's `spec.source.helm.valueFiles` (see [`HelmUpdater.ProcessHelmSourceUsingStepVariables`](https://github.com/OctopusDeploy/Calamari/blob/main/source/Calamari/ArgoCD/Conventions/UpdateImageTag/HelmUpdater.cs)). Inline `spec.source.helm.valuesObject` is *not* written — worth calling out, since users with valuesObject-based Applications will otherwise chase missing tag bumps. Rewrote the bullets to spell out both code paths and the valuesObject limitation.

Found while debugging an Octopus + Argo CD setup that ran "Success" but reported `Nothing to update for Application` for days — one guessed OCL property, one valuesObject Application.


rain-on (approved):
LGTM

PR #3243, opened 2026-07-15
Upgrade to Astro 7

borland:
Why: Astro 7 is faster than prior versions (The github actions build looks like it takes 9 minutes now, instead of 11+), and it is good to keep on top of dependency updates

Claude made the changes in this PR, and wrote the content below the line

I verified the output using `pnpm dev` and `pnpm preview` and as far as I can see, everything looks as it should.

-----

## What

Upgrades the docs site from Astro 6 to Astro 7 (`astro@^7.0.9`, `@astrojs/mdx@^7.0.3`) and fixes the breaking changes the upgrade surfaced. `pnpm build` completes cleanly (2695 pages, no warnings).

## Changes

- **`Header.astro` — build failure fix.** Astro 7 hands non-inline `<script>` tags to Vite for bundling, which cannot resolve the `/docs/js/modules/theme-switcher.js` **public** asset and failed the build with `UNRESOLVED_IMPORT`. Marked the script `is:inline` so Astro emits it verbatim and the browser resolves it at runtime — consistent with how the rest of the site references its public JS.

- **`astro.config.mjs` — deprecation migration.** `markdown.remarkPlugins` is deprecated in Astro 7. Moved the remark plugins into `markdown.processor: unified({ remarkPlugins })` from `@astrojs/markdown-remark`, and added that package as a direct dependency (it's now imported by the config; previously only transitive).

- **`astro.config.mjs` — config cleanup.** Hoisted `trailingSlash: 'never'` to the top level, where it's actually honored (it was previously nested under `markdown` and silently ignored), and removed the dead `extendDefaultPlugins` option (a no-op since Astro 2).

## Testing

`pnpm build` — passes, exit code 0, 2695 pages built, no deprecation or config warnings.

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -9,6 +9,7 @@ import llmMdEmitter from './src/integrations/llm-md-emitter.ts';
 // https://astro.build/config
 export default defineConfig({
     site: 'https://octopus.com',
+    compressHTML: true, // preserve astro v6 behavior - https://docs.astro.build/en/guides/upgrade-to/v7/
     integrations: [
         mdx(),
         llmMdEmitter()
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -1,6 +1,7 @@
 import remarkDirective from 'remark-directive';
 import remarkHeading from 'remark-heading-id';
 import { defineConfig } from 'astro/config';
+import { unified } from '@astrojs/markdown-remark';
 import mdx from '@astrojs/mdx';
 import { attributeMarkdown, wrapTables } from '/src/themes/octopus/utilities/custom-markdown.mjs';
 import llmMdEmitter from './src/integrations/llm-md-emitter.ts';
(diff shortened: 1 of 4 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3243.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3243&status=all).

enf0rc3:
Heads up on the Astro 6→7 upgrade: v7 changes the default compressHTML to JSX-style whitespace stripping, which removes whitespace between inline elements in `.astro` templates (our .md/.mdx content is unaffected). 

I've spotted the difference in the breadcrumbs — the crumbs tighten up around the chevron separators since .site-breadcrumbs li is display: inline and relied on source whitespace for spacing. 

It's site-wide as breadcrumbs render on every docs page, and the top nav is likely affected too. 

Screenshots below — if the spacing isn't acceptable we can set `compressHTML: true` to restore the v6 behaviour.

image

More visible when swapping between tabs:
- [original](https://octopus.com/docs/installation/requirements)
- [changes](https://stoctodocspr3243.z22.web.core.windows.net/docs/installation/requirements)

borland:
Thanks @enf0rc3. I've updated the astro config to restore the v6 behaviour. If you refresh the ephemeral environment at https://stoctodocspr3243.z22.web.core.windows.net/docs/installation/requirements you can see it now matches the original

steve-fenton-octopus (approved):
Tested on the ephemeral environment. Looks good to go.

enf0rc3 (approved):
Looks good, thanks for fixing the spacing issue

PR #3258, opened 2026-07-28
Cut docs build time by ~70% (3.7x faster)

borland:
I was unhappy with how long it takes to build our docs site. I prompted claude with the following and let it run.

> I am unhappy with how long it takes to build the docs site. It is built using pnpm build; more details are in the README.md file.
We upgraded to astro 7 and I was expecting a bigger performance uplift. I suspect that some plugins or customisations we've made might be bottlenecking it and slowing things down. Additionally I notice when the build process runs it doesn't seem to use more than a single CPU core.
Can you help figure out what is affecting the build performance. If you can, try and put the repo in a vanilla astro state (no plugins/customisations) but still has a successful build; then binary search the plugin space and tell me where the time is going and what we could do to improve it. Also consider that I may be wrong and it may not be plugins, but I'd like to understand where the time is going

It investigated many things and made some meaningful changes. 

Numbers from GitHub actions:

In a previous PR, the "Astro build and test" phase took **9 minutes 22 seconds**
In this PR, the "Astro build and test" phase took **4 minutes 44 seconds**

While we don't see a 70% reduction there, note that it's including the tests as well, which weren't part of the optimization.

The rest of the PR description beyond this point is written by Claude; I asked it to summarize the changes it made, and provide evidence as to why this change was safe.

-----------------------

## Summary

`pnpm build` is now **~3.7x faster — about 73% less wall-clock time**, with byte-identical output.

Measured across six full builds on one machine (16-core macOS, Node 24): **190.9s → 51.8s** mean, a 3.53x–3.87x range run to run. Absolute numbers will differ on your hardware and in CI; the ratio is the portable figure. The phase that shrank is static route generation: **157s → 17.5s (~9x)**.

The cause was not plugins or Astro 7. ~78% of the build was three pieces of per-page code recomputing data that is identical for every page.

## Where the time went

Profiled a full build with `node --cpu-prof`. Of ~198s:

| Cost | Time | Share |
|---|---|---|
| `Navigation.autoMenu()` → `getChildren()` | **95.2s** | 48% |
| `getEligibleSlugs()` in `CopyMarkdown.astro` | **29.0s** | 15% |
| `Posts.all()` → disk cache `readFileSync` + `JSON.parse` | **22.9s** | 12% |
| markdown/MDX compile (micromark, remark, shiki) | 26.5s | 13% |
| vite/rolldown/esbuild bundling | 20.0s | 10% |
| copying the 609 MB `public/` dir | 5.2s | 3% |

Per-page render times were bimodal: 1,408 redirect pages at ~1ms, and 1,270 content pages at **~120ms each**. Everything above lives in that 120ms.

The three culprits:

1. **`astro-accelerator-utils` `Navigation.getChildren()`** is recursive and runs a full scan of all ~2,700 pages *at every node* — O(n²) per page, for a tree that is identical on every page. Called twice per page: from `Navigation.astro`, and from `ArticleJourney.astro` (which flattens the whole tree just to find prev/next links).
2. **`getEligibleSlugs()`** re-parses frontmatter for all 2,661 docs pages on every page render. Its comment said "No memoization… the walk is sub-second" — true per call, but it's called 1,270 times.
3. **`Posts.all()`** has no in-memory cache. It round-trips the entire page set through `.cache/v1_posts.all.json` (1.1 MB) — `readFileSync` + `JSON.parse` — on *every* call, 6–8 times per page. `Accelerator`'s getters also construct a brand-new `Posts`/`Cache`/`Navigation` on every property access, so memoization was impossible by design.

**It isn't the plugins.** Swapping `Default.astro` for a bare `<html><slot/></html>` while keeping every integration (mdx, the custom remark processor, shiki, `llm-md-emitter`) builds in 39.8s on the same machine — i.e. the plugin surface plus content compilation is ~77% of what the build now costs, and we're within ~12s of that floor. `llm-md-emitter` costs about 1s.

## What changed

- **`src/lib/accelerator.ts`** (new) — one shared `Accelerator` per process with memoized helpers and a memoized `posts.all()`. Returns a shallow copy per call because `Navigation.breadcrumbs()` splices the array it's given. Resolved lazily; see the gotcha below.
- **`src/lib/navigationTree.ts`** (new) — builds the nav tree once. `Navigation.astro` clones only the branch the reader is in (every other section renders collapsed anyway) and recomputes `isOpen`/`ariaCurrent` itself. `ArticleJourney.astro` indexes into a pre-flattened leaf list.
- **`mdxContent.ts`** — memoized `getEligibleSlugs()`.
- 43 files mechanically switched from `new Accelerator(SITE)` to the shared import.

All three memoizations are gated on `import.meta.env.PROD`, so `astro dev` keeps its current always-fresh behaviour — adding or renaming a page still updates the nav without a server restart.

## How to gain confidence this isn't breaking

This is a pure-performance change, so the bar is that the site is *identical*, not merely working. Three independent checks:

**1. Byte-for-byte output comparison — the main one.** I built `main` and this branch and compared every generated file:

```
main: 3953 files | branch: 3953 files | only-main: 0 | only-branch: 0
DIFFERING: 0
```

That covers all 2,697 HTML pages plus the 1,252 emitted per-page `.md` files, `llms.txt`, `llms-full.txt`, `sitemap.xml` and `search.json`. The only normalisation applied was the footer's `data-shared-note="fetched … at <timestamp>"` attribute, which already differs between any two builds of `main` because the footer is fetched live at build time.

To reproduce:

```bash
git checkout main && rm -rf .cache && pnpm build && cp -r dist /tmp/out-main
git checkout perf/build-investigation && rm -rf .cache && pnpm build
diff -rq /tmp/out-main dist   # only the footer timestamp should differ
```

**2. Playwright suite: 90/90 passing**, including `redirect.spec.ts` which checks all **1,389 redirects**, and `llm-endpoints.spec.ts` which independently verifies that the emitted `.md` set, the `CopyMarkdown` dropdown and the `llms.txt` slug set all still agree — that's a direct check on the `getEligibleSlugs()` memoization.

**3. Reasoning about the risky bits.** The two places where shared state could realistically leak between pages:

- `posts.all()` now returns a snapshot. `Navigation.breadcrumbs()` **splices the array it receives**, which is why `all()` hands out `.slice()` rather than the array itself. Copying 2,700 references is microseconds; parsing 1.1 MB of JSON is ~18ms.
- The nav template is shared read-only. `Navigation.astro` spreads each top-level node and `structuredClone`s only the expanded branch before touching any flags; `journeyOrder()` clones before flattening, because flattening sorts each level in place.

If you want more assurance than the byte diff, the highest-value manual spot-check is the left-hand nav on a deep page (correct section expanded, correct item highlighted, others collapsed to their section label) and the prev/next links at the bottom of an article — those are the two components whose logic actually changed.

## Notes for review

- **A gotcha worth knowing.** My first attempt resolved `posts.all()` at module-init time and broke the build: `Posts.all()` runs an eager `import.meta.glob` over every page, so forcing it during module init pulls page modules in before their own imports have initialised, yielding `undefined` entries. Worse, those got written to `.cache/v1_posts.all.json`, and the *next* build read them back as `null` and failed in an unrelated file. It's resolved lazily now, but it shows `.cache/` can poison a subsequent build.
- **`.cache/` is arguably dead weight now.** It persists between builds with a 200-second TTL (`cacheMaxAge` in `src/config.ts`), which also means any build slower than 200s expired its own cache mid-flight and rewrote it. With `posts.all()` memoized in-process it no longer does anything useful for the build. Removing it would be a reasonable follow-up, but it's out of scope here.
- **On the single-core observation.** Astro's `build.concurrency` defaults to `1`, but raising it barely helps: at `concurrency: 8` I measured 50.1s vs 50.7s and still only ~117% CPU. Page rendering is synchronous single-threaded JS, so concurrency only overlaps the async parts. Output was identical, so it's safe if you want the ~3s, but I left it out to keep this diff focused.
- **Remaining profile is flat** — markdown compile 23.4s, bundling 17.6s, asset copy 4.2s. No hotspot left worth chasing. The next-largest lever would be the 609 MB `public/` directory.

### DIFF
--- a/src/components/ArticleHeader.astro
+++ b/src/components/ArticleHeader.astro
@@ -1,10 +1,9 @@
 ---
-import { Accelerator } from 'astro-accelerator-utils';
+import { accelerator } from '@lib/accelerator';
 import type { Frontmatter } from 'astro-accelerator-utils/types/Frontmatter';
 import { SITE } from '@config';
 import { Lang } from '@util/Languages';
 
-const accelerator = new Accelerator(SITE);
 const stats = new accelerator.statistics('components/ArticleHeader.astro');
 stats.start();
 
--- a/src/components/ArticleJourney.astro
+++ b/src/components/ArticleJourney.astro
@@ -1,11 +1,10 @@
 ---
-import { Accelerator } from 'astro-accelerator-utils';
+import { accelerator } from '@lib/accelerator';
 import type { Frontmatter } from 'astro-accelerator-utils/types/Frontmatter';
 import type { NavPage } from 'astro-accelerator-utils/types/NavPage';
-import { SITE } from '@config';
 import { Lang } from '@util/Languages';
+import { journeyOrder } from '@lib/navigationTree';
 
-const accelerator = new Accelerator(SITE);
 const stats = new accelerator.statistics('components/ArticleJourney.astro');
 stats.start();
 
@@ -21,25 +20,12 @@ const { lang, frontmatter, headings } = Astro.props satisfies Props;
 const _ = Lang(lang);
 
 // Logic
-const articles: NavPage[] = [];
-
-function flatten(pageTree: NavPage[]) {
-  pageTree.sort((a, b) => a.order - b.order);
-  pageTree.forEach((page) => {
-    if (page.children.length == 0) {
-      articles.push(page);
-    } else {
-      flatten(page.children);
-    }
-  });
-}
-
 const currentUrl = new URL(Astro.request.url);
-const pages = accelerator.navigation.autoMenu(SITE.subfolder);
-accelerator.navigation.setCurrentPage(pages, currentUrl);
-flatten(pages);
+const articles = journeyOrder();
 
-const current = articles.findIndex((p) => p.ariaCurrent);
+// Was: setCurrentPage() over a freshly rebuilt tree, then findIndex on the
+// ariaCurrent flag it set. Same predicate, no rebuild.
+const current = articles.findIndex((p) => p.url === currentUrl.pathname);
 
 const previous = articles[current - 1] || null;
 const next = articles[current + 1] || null;
--- a/src/components/ArticleNav.astro
+++ b/src/components/ArticleNav.astro
@@ -1,10 +1,9 @@
 ---
+import { accelerator } from '@lib/accelerator';
 import { SITE } from '@config';
 import { Translations, Lang } from '@util/Languages';
-import { Accelerator } from 'astro-accelerator-utils';
 import Separator from './Separator.astro';
 
-const accelerator = new Accelerator(SITE);
 const stats = new accelerator.statistics('octopus/components/ArticleNav.astro');
 stats.start();
 
--- a/src/components/EditOnGithub.astro
+++ b/src/components/EditOnGithub.astro
@@ -1,10 +1,9 @@
 ---
-import { Accelerator } from 'astro-accelerator-utils';
+import { accelerator } from '@lib/accelerator';
 import type { Frontmatter } from 'astro-accelerator-utils/types/Frontmatter';
 import { SITE } from '@config';
 import { Lang, Translations } from '@util/Languages';
 
-const accelerator = new Accelerator(SITE);
 const stats = new accelerator.statistics(
   'octopus/components/ArticleList.astro'
 );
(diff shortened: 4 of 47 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3258.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3258&status=all).

shaunmarx (approved):
LGTM

PR #3259, opened 2026-07-28
Rework the theme selector ahead of the design refresh

enf0rc3:
Rewrites the theme switcher so it is a solid base for the design refresh. A 1:1 port of the existing behaviour, with the bugs fixed and the moving parts in idiomatic Astro places.

## Bugs fixed

1. **The two switchers desynced.** Each instance cached `isDarkMode` at construction, so toggling the mobile one left the header one stale. Resize past 1130px and it showed the wrong position; clicking it then set `data-theme` to the value it already had.
2. **The knob animated on every page load.** Slider position came from a JS-applied `.dark-mode` class, so in dark mode it painted on the light side and slid across after hydration. Now driven by `html[data-theme]`, correct on the first frame.
3. **Keyboard users tabbed to a non-widget.** `tabindex="-1"` on the input plus `tabindex="0"` on the label meant the focusable element was a bare `<label>` — no role, name or state exposed to a screen reader. The real `input[role=switch]` is focusable again, the deprecated `keyCode` handler is gone, and the focus ring is drawn via `input:focus-visible + label`.
4. **`localStorage` was unguarded.** A throw in the head script (Safari private mode, cookie-blocked iframes) killed the IIFE and left the site permanently light regardless of OS preference.
5. **Enter did not toggle the switch.** A native checkbox ignores Enter, so it is bound explicitly and routed through `click()`, keeping the `change` listener the only writer of state. Space and Enter are both tested.

## What changed

| File | |
|---|---|
| `src/lib/theme.ts` | new — attribute names, storage key and types in one place |
| `src/scripts/theme-switcher.ts` | new — the controller, bundled by Astro |
| `src/themes/octopus/components/ThemeScript.astro` | new — pre-paint inline script, values via `define:vars` from `theme.ts` |
| `src/components/ThemeSwitcher.astro` | new — one switcher component, takes an `id` |
| `tests/theme.spec.ts` | new — 9 tests |
| `public/docs/js/modules/theme-switcher.js` | deleted |

- The controller moved from `public/` into `src/scripts/`, so Astro bundles, minifies and dedupes it rather than shipping an unhashed public file loaded via `is:inline`.
- The pre-paint script moved out of the `HEADER_SCRIPTS` string in `config.ts` into `ThemeScript.astro`, rendered first in `<head>` — ahead of the stylesheet links, so the parser is not waiting on a CSS download to run it. It stays `is:inline` deliberately; it must not be bundled or deferred.
- Switcher markup deduplicated into `ThemeSwitcher.astro`, used by both `Header.astro` and `MobileMenu.astro`.
- `astro:after-swap` handler so the switcher keeps working if `<ClientRouter />` is ever added.
- OS preference changes are followed until the user picks a side; other tabs sync via the `storage` event.
- Removed the dead `<body data-theme="dark">`, and the `DOMContentLoaded` wrapper the old script did not need.
- Accessible name is now "Dark mode" — "Toggle theme between light and dark mode, switch, on" read badly.

We tried adding `color-scheme` so native scrollbars would follow the theme, but our existing text-colour transitions made it look wrong, so it is not included.

## Testing

`tests/theme.spec.ts` — 9 tests, all passing: OS preference followed in both schemes, explicit choice beating the OS preference and surviving navigation, both switchers staying in sync, Space and Enter each toggling both ways, fallback when `localStorage` throws, `data-theme` present with scripting disabled, and a regression guard that the knob position is attribute-driven rather than script-driven.

`pnpm build` passes (2697 pages).

I have also manually tested the above bugs and can confirm they are resolved, i found one trivial thing that was fixed being pressing enter on the toggle when tabbing through ui


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3259.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3259&status=all).

playwright.config.ts
```diff
@@ -1,12 +1,15 @@
 import type { PlaywrightTestConfig } from '@playwright/test';
 import { devices } from '@playwright/test';
 
+// Override with PLAYWRIGHT_PORT to run the suite alongside a dev server.
+const port = Number(process.env.PLAYWRIGHT_PORT ?? 3000);
```
enf0rc3: so we can run tests while dev server is running

playwright.config.ts
```diff
@@ -28,8 +31,8 @@ const config: PlaywrightTestConfig = {
   ],
   outputDir: 'test-results/',
   webServer: {
-    command: 'npm run preview',
-    port: 3000,
+    command: `npm run preview -- --port ${port}`,
```
borland: What's up with this change here? Did the previous port key in the config object not work?

borland:
I don't profess to understand the nitty-gritty detail of css and root selectors vs other things, but at a high level it all seems sound and reasonable.

I noticed though on the staging site, when switching between light and dark mode, the color changes on the header items now "pulse" in a way that they didn't before.

To experience this, go to this page: https://octopus.com/docs/infrastructure/accounts/openid-connect on the main site, and toggle between light and dark a few times quickly.

Then do the same on the staging site: https://stoctodocspr3259.z22.web.core.windows.net/docs/infrastructure/accounts/openid-connect and note how all the header text such as "OpenID Connect" animates differently. 

While personally I think it's kind of weird looking, It isn't _unacceptable_ --  we could choose to accept the new color changes. But if we can fix them instead, we should

playwright.config.ts
enf0rc3: Added it so we can run the test while running the dev server, rather than being hard coded to 3000

borland: yeah but it still has port on the line below.

And would this also work?

```javascript
webServer: {
  command: 'npm run preview',
  port: port,
}
```

enf0rc3: Apparently no: 
```
astro preview takes its port from server.port in astro.config.mjs and never reads the Playwright config. webServer.port only tells Playwright which port to poll before starting tests, so setting just that leaves it waiting on a port nothing is serving. The original config worked because both values were independently hardcoded to 3000 and happened to agree. Either way it's gone now — it was only ever for my convenience.
```

but i just reverted it doesn't add much value and if i need to scaffold it again i can just re implement it and not commit, no harm

enf0rc3:
Thanks for finding that bug @borland  So it turns out it was the addition of using "[color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/color-scheme)"  because of pre-existing "transtions" that were defined on main. 

[Super weird to have transitions there](https://github.com/OctopusDeploy/docs/blob/3964829522cd0b3fdf24dd583fef9710c4de50da/public/docs/css/main.css#L180), but it may be there for a reason, so i reverted the scroll bar back to has it was to not use color-scheme

borland (approved):

PR #3267, opened 2026-07-29
Add work-in-progress new TopNav component

rosslovas:
# Summary

Adds an early stage implementation of an upcoming `TopNav` redesign.

No functionality is implemented, just visuals. Only the signed-out variant is implemented, and the search bar is entirely unimplemented.

# Results
## Target design:

image

## As implemented:
See it in action at https://stoctodocspr3267.z22.web.core.windows.net/components#topnav

image

Variations of note from the original design:
- The `Docs` and `Start for free` text are weight `600` rather than `700`, which was discussed and agreed upon as 600 matches the intended tokens.
- The theme switcher icon isn't correct nor does it switch between icons for light/dark mode, it's just a placeholder for now.


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3267.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3267&status=all).

src/components/TopNav.astro
```diff
@@ -0,0 +1,56 @@
...
+          >
+            {label}
+          </a>
+        ))
+      }
+    </nav>
+  </div>
+  <div class="top-nav__trailing">
+    <!-- All of these are just non-functional and for visual/presentation only at this stage -->
+    <Button icon="fa-regular fa-moon" aria-label="Switch to dark theme" />
```
rosslovas, line 50: I expected `fa-regular fa-moon` to be the outline [like their website says](https://fontawesome.com/icons/classic/regular/moon), but it's exactly the same as solid 🤔 Not sure why, but I'm just leaving this as a placeholder for now anyhow, as I'm just trying to get the layout right at this early stage.

src/styles/vars.css
```diff
@@ -254,6 +255,11 @@
   --grid-gap-main: 1rem;
 
   --standard-radius: 0.5rem;
+
+  /* TODO: drop these once color.nav.background.* are available in design-system-tokens */
+  --colorNavBackgroundRest: rgb(from var(--colorScalesBlue500) r g b / 0);
+  --colorNavBackgroundHover: var(--colorScalesSlate100);
+  --colorNavBackgroundActive: rgb(from var(--colorScalesBlue500) r g b / 0.06);
```
rosslovas, line 228: I believe I need the design folks to get these added on my behalf, that would be appreciated whenever possible 🙇

ellen-octopus (approved):
Looks pixel perfect 🤩  thanks @rosslovas! 
Also just noting, I know I haven't provided a responsive layout for the nav yet, but since this is an early implementation, we can add it later.

src/styles/vars.css
ellen-octopus: @rosslovas I'll work with design team next week to make these available

enf0rc3: + dark mode as well

I assume that these ones: https://github.com/OctopusDeploy/OctopusDeploy/blob/main/frontend/packages/design-system-tokens/src/generated/lightTheme.css#L136-L144 

are not for "top menu" but for [Area Nav List?](https://www.octopus.design/latest/components/navigation/area-nav-list/overview-bNV9s3ma)  instead.

ellen-octopus: Those are only for in-product navigation. We won't be using the same for the docs site :)

src/components/TopNav.astro
enf0rc3: I have just had an icon,  that wasn't on the free fontawesome fonts, i opted to just copy the svg from design-system-icons 

https://www.octopus.design/latest/foundations/icons/system-icons-upsPwkne

it has a moon there too, I just opened up a new claude in octopusdeploy repo and got it to write the prompt for me and extract the code, you can do a better prompt but for berevity this prompt gave me the exact svg and the instructions to implement

```
I need to use a u design-system-icons from here specifically the link-on icon, however the website doesnt support react and it uses css by js, for now i just want the one icon but we will look at other ways to get it, for the single icon how can i easily get the link-on in my astro html css js website? https://www.npmjs.com/package/@octopusdeploy/design-system-icons  (thats the package enerated by this repo's package)
```

src/styles/main.css
```diff
@@ -2950,7 +2972,85 @@ img.btn__icon {
 .btn-row + .btn-row {
   margin-block-start: var(--space8);
 }
-/* Button.astro component end */
+/* Button component end */
+
+/* TopNav component */
+.top-nav {
```
enf0rc3, line 2978: Apparently, with Astro, the idiomatic way is to do these in the `.astro` file

https://docs.astro.build/en/guides/styling/#scoped-styles

enf0rc3 (approved):
Happy to go without this change as it isn't enabled yet/

PR #3269, opened 2026-07-29
Make the Infrastructure target-tag matrix table fill the content column

enf0rc3:
## What

The target-tag matrix table on `/docs/infrastructure` rendered **713px wide in an 800px content column**, so its row borders stopped short of the right edge and its headings sat 37–59px right of the data they label. Now 800px of 800px, headings aligned.

## Changes

One opt-in class in `public/docs/css/main.css`:

```css
.table-full-width table {
  width: 100%;
}

.table-full-width th {
  text-align: left;
}
```

and that table wrapped in `:::div{.table-full-width}`.

`th` has no `text-align` anywhere in the stylesheet, so it falls back to the browser default of `center` — unnoticeable while a table is only as wide as its content, wrong once it fills the column. Both properties are in the one class for that reason.

I tested applying it to all tables as default,  and ran side by side comparisons but it made some tables less readable, opted to instead implement this version with just one. 

## Scope

`.table-full-width` appears three times in the repo: that one table, and these two rules. No other table on the site changes, verified on the deployed build against a 12-table control page.

Mobile is unchanged. The table's min-content width is 628px, so below roughly 800px of content column nothing moves.

## Before / After

image
image
image

### DIFF
--- a/public/docs/css/main.css
+++ b/public/docs/css/main.css
@@ -417,6 +417,9 @@ h6:hover .bookmark-link,
 
 table {
   border-collapse: collapse;
+  /* Without this, auto layout shrinks tables to their content width and leaves
+     a ragged gap on the right of the content column. */
+  width: 100%;
 }
 
 table thead {
--- a/public/docs/css/main.css
+++ b/public/docs/css/main.css
@@ -419,6 +419,10 @@ table {
   border-collapse: collapse;
 }
 
+.table-full-width table {
+  width: 100%;
+}
+
 table thead {
   font-weight: 700;
 }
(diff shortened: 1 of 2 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3269.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3269&status=all).

borland (approved):

borland:
We can merge this one now. The changes are so minor that it won't make a difference if people get served old vs new CSS, it'll be fine either way

PR #3270, opened 2026-07-30
Emit bundled Astro assets under the /docs/ prefix

enf0rc3:
# Background

Azure Front Door serves this site at `octopus.com/docs` and declines anything outside `/docs`. 

See: sources it has is only under docs + octopus-public (this is for the footer actually not deployed by this)

image 

Astro emitted its bundled component CSS to `/_astro/`, so it 404s on the CDN.

Error example:
image

This broke the theme toggle selector and the search styles:

image

This was the first [CSS file bundled](https://docs.astro.build/en/guides/styling/#import-a-local-stylesheet) this way, hence why it didn't exist before.

# Result
Sets `build.assets: 'docs/_astro'` so the bundle lands inside the routed prefix.


enf0rc3:
Broken deploy, new pr works https://github.com/OctopusDeploy/docs/pull/3271

PR #3271, opened 2026-07-30
Emit bundled Astro assets under the /docs/ prefix

enf0rc3:
# Background

Azure Front Door serves this site at `octopus.com/docs` and declines anything outside `/docs`. 

See: sources it has is only under docs + octopus-public (this is for the footer actually not deployed by this)

image 

Astro emitted its bundled component CSS to `/_astro/`, so it 404s on the CDN.

Error example:
image

This broke the theme toggle selector and the search styles:

image

This was the first [CSS file bundled](https://docs.astro.build/en/guides/styling/#import-a-local-stylesheet) this way, hence why it didn't exist before.

# Result
Sets `build.assets: 'docs/_astro'` so the bundle lands inside the routed prefix.

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -10,6 +10,11 @@ import llmMdEmitter from './src/integrations/llm-md-emitter.ts';
 export default defineConfig({
     site: 'https://octopus.com',
     compressHTML: true, // preserve astro v6 behavior - https://docs.astro.build/en/guides/upgrade-to/v7/
+    build: {
+        // The site is proxied onto octopus.com under /docs/ only, so bundled
+        // assets must live inside that prefix or the CDN 404s them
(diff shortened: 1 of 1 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3271.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3271&status=all).

enf0rc3:
Tested with `dev` `preview` and static site, astro folder now under `/docs` instead of on root

image

rosslovas (approved):
The ephemeral preview site works. I think we should be good here, nice one 🎉

PR #3272, opened 2026-07-31
Import stylesheets such that Vite bundles/minifies/content-hashes them

rosslovas:
# Summary

A lack of cache busting poses problems for our docs site, and among other things this affects `main.css` and `vars.css` which contain the bulk of our custom styles.

By `import`ing these stylesheets, Vite will bundles/minify/content-hash them such that their contents are served in `docs/_astro/components.<hash>.css` and the hash in the filename achieves the desired cache-busting.

Note: this does not affect third-party CSS we ship such as FontAwesome which still lacks any cache-busting mechanism.

We have also decided to temporarily leave the CSS files in the public folder because html is currently cached for 7 days, so clients will be expecting to find this CSS at this and will 404 if it's moved to  `src/styles/*.css`. (This will be done separately)

# Risk

This PR changes the order that the styles appear in the page. Design tokens and third party Font Awesome stylesheets used to come _after_ our custom styles, but now they appear before.

To the best of my ability to tell this should be inert. To verify this I had a script generate screenshots of every single page in their entirety (except the credits page because it's too big), both light mode and dark mode, before and after. By diffing the results I was able to verify that there were absolutely no differences to any docs pages. I also validated the script actually worked by putting in a few canary edits, which _were_ correctly detected.

The one thing it detected had changed is the contents of the `reports` folder, and this is because `main.css`/`vars.css` were not applied to them at all before. However, (a) as far as I can tell that content is not on the public site anyway, and (b) it just looks slightly different now in a harmless way, for example:

<details>
<summary>Missing metadata page before</summary>

image

</details>

After can be seen at: https://stoctodocspr3272.z22.web.core.windows.net/report/missing-meta (it's more blue 🟦)

# Test
image


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3272.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3272&status=all).

borland (approved):
Change looks simple, I had a look at the staging site and LGTM

src/themes/octopus/components/HtmlHead.astro
```diff
@@ -74,8 +80,6 @@ stats.stop();
   <link rel="stylesheet" href={`${ASSETS}/pu-vars.css`} />
```
borland, line 82: Should we also import these things? I guess because they're third-party it doesn't matter, and we're more likely to delete rather than update them?

enf0rc3: If i'm correct, these come from a different website? 

```
export const ASSETS = import.meta.env.SHARED_ASSETS_BASE_URL;

//.env
SHARED_ASSETS_BASE_URL=https://octopus.com/octopus-public/assets
```

It turns out the footer is hosted elsewhere

borland: Oh right, fair enough then :smile:

PR #3274, opened 2026-07-31
Import JS files such that Vite bundles/minifies/content-hashes them

rosslovas:
# Summary

A lack of cache busting poses problems for our docs site, and among other things this affects our custom JavaScript code.

By `import`ing these files, Vite will minify/content-hash them such that their contents are served in `docs/_astro/file.<hash>.js` files, and the hashes in the filenames achieves the desired cache-busting.

This is the JS counterpart to #3272.

We have also decided to temporarily leave the JS files in the public folder because HTML is currently cached for 7 days, so clients will be expecting to find this JS at this and will 404 if it's moved to src/scripts/modules/*.js. (This will be done separately)

# Risk

Like in the CSS PR, I had a script generate screenshots of every single page in their entirety (except the credits page because it's too big), both light mode and dark mode, before and after. By diffing the results I was able to verify that there were absolutely no differences to any page.

However, JS is inherently more dynamic than CSS so screenshots do not tell the whole story, further testing is required.


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3274.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3274&status=all).

borland (approved):

PR #3276, opened 2026-08-03
Adopt the design system font tokens, update heading spacing, and rebuild the anchor button

enf0rc3:
## Summary

Three related pieces of work:

1. **Adopted the design system font tokens.** `src/styles/main.css` now names tokens from `@octopusdeploy/design-system-tokens` directly, and the docs-local font alias layer in `vars.css` is deleted.
2. **Updated heading typography and spacing to the new design.** Sizes, weights, line-heights and the space around every heading now match the Figma spec.
3. **Rebuilt the heading anchor as a design system icon button**, using `LinkOnIcon` instead of a Font Awesome glyph — groundwork for the button work coming next.

## Closer to the Figma design

| | Before | After |
|---|---|---|
| `.header__title` (page h1) | 26px / 700 / 31px | **32px / 600 / 44px** |
| in-content `h1` | 44px / 800 / 53px | **32px / 600 / 44px** |
| `h2` | 24px / 800 / 34px | **24px / 600 / 32px** |
| `h3` | 20px / 800 / 28px | **20px / 600 / 28px** |
| space above `h2` | 48px | **56px** — `--space56` |
| space above `h3` | 48px | 48px — `--space48` |
| space above `h4` / `h5` / `h6` | 48px | **40 / 32 / 24px** |
| space below any heading | 24px | **12px** — `--space12` |
| `h2` immediately followed by `h3` | 56px | **32px** — `--space32` |
| `body` | 16px / 400 / 25.6px | **16px / 400 / 24px** |
| breadcrumbs | 12.8px | **14px / 400 / 20px** |
| heading `letter-spacing` | `-0.02em` | **removed** |

image

Figma's weight 590 is SF Pro's semibold, which is 600. The 800 → 600 drop lands on every page and is the most visible change here.

Every spacing figure above is the **total** gap between two elements. Headings space themselves with `margin-block-start` rather than padding, so the previous element's own bottom margin collapses into it instead of adding to it — `max(16px, 56px)` is 56px, where padding would have given 72px. That mirrors Figma, where `gap` applies between siblings and contributes nothing after the last one. `.page-content` carries `display: flow-root` so a heading that starts an article keeps its top margin instead of collapsing it out of the container.

Already matching and unchanged: `h1` colour (`--color-heading` → `--colorTextPrimary`), h2 as `--textHeadingLarge`, h3 as `--textHeadingMedium`.

## Not specified in Figma — worth a second opinion

No spec exists for these, so they stay as close to current as the token scale allows. Each is a judgement call.

| Rule | Change | Why |
|---|---|---|
| `h4` | 18 → 16px, 800 → 600 | 18px has no token; 20px is h3's size |
| `h5` | 16 → 14px | heading composites jump 16 → 12px |
| `h6` | 14 → 13px | same gap |
| `blockquote cite`, `.card__title`, `.icon-tile__title`, `.image__caption` | 18 → 16px | 18px sits between `--fontSizeBase` and `--fontSizeLarge`; down avoids colliding with h3 |
| `.card__description` | 16 → 14px | a step below its 16px title, matching `.icon-tile__description` |
| `.bookmark-link` 19.2 → 20px, `.magnify-icon` 20.8 → 20px | icons snapped to `--fontSizeLarge` |

**Beware the name trap:** the old `--font-size-medium` was **16px** while the token `--fontSizeMedium` is **14px**. So several rules look downgraded but are unchanged — `.octo-header`, `.article-nav__link` and `.card__label` are 16px before and after; `.badge` and `.theme-switcher__icon` are 14px before and after.

Block Quote: (sizing only so no light/dark mode required)
image

## Heading anchor button
Before:
image
After:
image 
light mode: 
image

Cards:
image

## Also fixes Inter on the shared chrome

`vars.css` declared `--font-family` and `--font-family-system` at `:root`, where octopus.com's `pu-vars.css` also declares them:

```css
/* pu-vars.css, loads first */
--font-family: "Inter", var(--font-family-system);
```

Ours won the cascade, and `pu-vars.css` styles the chrome with `.navbar-pu, .footer-pu { font-family: var(--font-family) }`. So the navbar and footer rendered in a system font on docs pages while rendering in Inter everywhere else on octopus.com. Both declarations are gone.

These were the only `:root`-level collisions between the two stylesheets — `pu-vars.css` scopes its other shared names under `.navbar-pu`/`.footer-pu`.

Not observable on the preview host: `InterVariable.woff2` returns 200 with no `Access-Control-Allow-Origin`, so a cross-origin `@font-face` fetch is discarded. On production it is same-origin. Check that the font is *requested*, not that it renders.

## Why 14 line-height declarations were added

`body` now carries an absolute `1.5rem` from its composite rather than a `1.6` ratio. A ratio recomputes per element; an absolute value inherits unchanged. So every rule setting a size above 16px states its own leading — without this, `.header__title` at 32px would sit in a 24px line box.

This is forced by the token set rather than a preference: the body composites stop at 16px, so 20px text at weight 400, 500 or 700 has no composite to use and must be assembled from primitives. The token pairings are also deliberately non-uniform (1.67 at 12px falling to 1.17 at 48px), so no single unitless ratio can express them.

## Removed from `vars.css`

`--font-size-xxlarge`, `-xlarge`, `-large`, `-regular`, `-medium`, `-small`, `--font-size-title`, `--font-size-subtitle`, `--heading-font`, `--body-font`, `--bold-font`, `--font-family`, `--font-family-system`. No references remain anywhere in the repo.

`--code-font`, `--code-background` and `--code-color` are also gone, but #3278 removed those on `main` rather than this branch.


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3276.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3276&status=all).

src/styles/main.css
```diff
@@ -152,12 +149,8 @@ em {
 }
 
 h1 {
-  font-size: 2.75rem;
-  line-height: 1.2;
-  font-weight: 800;
+  font: var(--textHeadingXLarge);
```
enf0rc3, line 153: All handled by `font: var(--textHeadingXLarge);`

src/styles/main.css
```diff
@@ -44,7 +40,7 @@ body.no-scroll {
 
 a,
 summary,
-summary > * {
+summary>* {
```
enf0rc3: There was autoformat during first commit

src/pages/index.md
```diff
@@ -26,6 +26,28 @@ This page is not shown on the production site.
...
+## Heading 2 \{#heading-2-b}
+
+### Heading 3 \{#heading-3-a}
+
+### Heading 3 \{#heading-3-b}
+
+<style>
+  #heading-2-b .bookmark-link,
+  #heading-3-b .bookmark-link {
+    opacity: 1;
```
enf0rc3, line 59: To show what it looks like on "hover" as per figma design

src/scripts/modules/headers.js
```diff
@@ -9,7 +9,9 @@ function enhanceHeaders() {
     const linkContainer = document.createElement('a');
     linkContainer.href = `#${elem.id}`;
     linkContainer.className = 'bookmark-link';
-    linkContainer.innerHTML = 'Bookmark';
```
enf0rc3, line 12: Why headers.js changed: innerHTML = 'Bookmark' → aria-label

Bookmark was never visible text — it was the anchor's accessible name. A CSS ::before glyph can't provide one, so without it screen readers announce "link" with no indication of the target.

It was hidden by clipping, which the new rule no longer does:
```
/* before */  width: 1.25em; height: 1.2em; overflow: hidden;
/* after  */  display: inline-flex; min-width: 28px; padding: var(--space4);
```
min-width is a floor, not a cap, and there's no overflow: hidden. A text node would sit next to the icon as a second flex item, widening the button and pushing the icon off-centre. So the text goes, and the name moves to aria-label — which also fixes the wording, since the control jumps to a section rather than bookmarking a page.

One trade-off: real text survives browser page-translation more reliably than aria-label. If translated docs become a requirement, visually-hidden text is the better pattern.

src/styles/main.css
```diff
@@ -374,23 +370,41 @@ strong {
...
 .bookmark-link {
-  border-radius: var(--standard-radius);
-  color: var(--fore-link);
-  display: inline-block;
-  font-size: 1.2rem;
-  height: 1.2em;
-  margin-inline-start: 0.3rem;
-  opacity: 0.5;
-  overflow: hidden;
+  display: inline-flex;
```
enf0rc3, line 394: This replaces the old icon of the link rings to the new button one, this potentially will change in upcoming pr when we implement the interactive click copy link

src/styles/vars.css
```diff
@@ -177,9 +170,7 @@
   --hamburger-line-width: 1.5rem;
   --hamburger-line-height: 0.2rem;
   --hamburger-line-margin: 0.2rem;
-  --hamburger-translate-y: calc(
-    var(--hamburger-line-height) + var(--hamburger-line-margin)
-  );
+  --hamburger-translate-y: calc(var(--hamburger-line-height) + var(--hamburger-line-margin));
```
enf0rc3: auto-format

src/styles/vars.css
```diff
@@ -192,24 +183,14 @@
   /* Card */
   --card-shadow: 0 10px 20px 0 rgba(0, 0, 0, 0.15);
   --card-related-topics-shadow: 0px 10px 35px 0px rgba(133, 152, 165, 0.1);
-  --card-related-topics-hover-shadow: 0px 10px 35px 0px
-    rgba(133, 152, 165, 0.25);
+  --card-related-topics-hover-shadow: 0px 10px 35px 0px rgba(133, 152, 165, 0.25);
```
enf0rc3: auto-format

src/styles/main.css
```diff
@@ -2991,9 +3014,10 @@ a[data-youtube] {
 }
 
 .magnify-icon:before {
+  /* fa-magnifying-glass-plus */
   content: '\f00e';
   font-family: fa-solid;
-  font-size: 1.3rem;
+  font-size: var(--fontSizeLarge);
```
enf0rc3, line 2855: Before:

Image

After:

Image

src/styles/main.css
```diff
@@ -3093,7 +3118,8 @@ figure:focus .magnify-icon {
 
 .rec-book h3 {
```
enf0rc3: Apparently ` rec-book` is not used /dead-code so i cant test this, but not brave enough to remove this.

mandymichael:
You might be interested to know about why we went with 600 font weight instead of the usual 700 for bold. [It's documented in the design system](https://www.octopus.design/latest/updates/decision-log/2024/bold-font-weights-rwWywhZs)

mandymichael (approved):

src/styles/main.css
```diff
@@ -2240,6 +2277,11 @@ a[data-youtube] {
 }
 
 .page-content {
+  /* Heading gaps rely on sibling margin collapsing, which only happens in block
+     flow. Converting this to flex or grid makes every gap additive again -
+     56px above an h2 becomes 72px after a paragraph, 88px after a blockquote. */
+  display: flow-root;
```
enf0rc3, line 2244: The only side-effect to this is 

Before:

Image

After:
Image

No actual change to height of the page content because it was collapsing. 

With nothing painted, 16px inside the box and 16px outside it render identically. Same for the top edge.

Also it means if there is an image added in markdown html with float (since MD can have html in it) it will stretch the container to fit the image. 

but i have checked markdown for this using claude:

```
Float audit result

┌─────────────────────────────────────┬────────────────────────────────────────────────┬────────┐
│               Search                │                     Scope                      │ Result │
├─────────────────────────────────────┼────────────────────────────────────────────────┼────────┤
│ float: left/right                   │ src/pages (all markdown/MDX)                   │ none   │
├─────────────────────────────────────┼────────────────────────────────────────────────┼────────┤
│ float: left/right                   │ src/shared-content, src/components, src/themes │ none   │
├─────────────────────────────────────┼────────────────────────────────────────────────┼────────┤
│ style="…float…"                     │ every built page under /docs                   │ none   │
├─────────────────────────────────────┼────────────────────────────────────────────────┼────────┤
│ align="left|right" on img/table/div │ every built page under /docs                   │ none   │
└─────────────────────────────────────┴────────────────────────────────────────────────┴────────┘

The last one matters because align is the legacy float equivalent — browsers implement it as a float, so it would have been caught by BFC containment even though it never appears in CSS. Nothing there either.

So display: flow-root on .page-content is confirmed safe:

- No floats exist inside it, in CSS or in authored content
- No position: sticky inside it
- It has no background, border, padding or margin, so containing the first and last child margins is visually identical to letting them escape
- main is a grid item and therefore already a BFC, so those margins were contained one level up anyway
- It does not clip, does not create a stacking context, and does not become a containing block for absolutely positioned descendants
```

src/styles/main.css
```diff
@@ -172,37 +165,49 @@ h3:not(:where([data-shared-fragment] *)),
 h4:not(:where([data-shared-fragment] *)),
 h5:not(:where([data-shared-fragment] *)),
 h6:not(:where([data-shared-fragment] *)) {
-  font-weight: 800;
-  line-height: 1.4;
   color: var(--color-heading);
-  padding: 2rem 0 0.5rem 0;
-  font-family: var(--heading-font);
-  letter-spacing: -0.02em;
+  margin-block-end: var(--space12);
```
enf0rc3, line 163: As per figma, is 12 instead of 16 (default)

src/styles/main.css
enf0rc3: removed as per other pr

PR #3278, opened 2026-08-04
Update styling for <code> blocks, wrap text on word boundaries

borland:
Uses the styling from design tokens / ui design figma and applies it to `<code>` blocks

# Before

image

image

Word wrapping:

image

# After

image

image

Word wrapping:

image

### DIFF
--- a/src/plugins/rehype-wbr.js
+++ b/src/plugins/rehype-wbr.js
@@ -1,0 +1,32 @@
+import { visit } from 'unist-util-visit';
+
+// Break *before* each of these characters
+const SEPARATORS = /(?=[.\[:\/\\])/;
+
+function withBreaks(value) {
+  const parts = value.split(SEPARATORS).filter(Boolean);
+  if (parts.length < 2) return null;
+
+  return parts.flatMap((part, i) => {
+    const text = { type: 'text', value: part };
+    return i === 0
+      ? [text]
+      : [
+          { type: 'element', tagName: 'wbr', properties: {}, children: [] },
+          text,
+        ];
+  });
+}
+
+export default function rehypeWbr() {
+  return (tree) => {
+    visit(tree, 'element', (node, index, parent) => {
+      if (node.tagName !== 'code') return;
+      if (parent?.tagName === 'pre') return; // fenced blocks, leave alone
+
+      node.children = node.children.flatMap((child) =>
+        child.type === 'text' ? (withBreaks(child.value) ?? [child]) : [child]
+      );
+    });
+  };
+}
--- a/src/styles/main.css
+++ b/src/styles/main.css
@@ -108,26 +108,27 @@ blockquote cite {
 
 /* TODO: Refactor code snippets. This needs to be a component. */
 code {
-  font-family: var(--code-font);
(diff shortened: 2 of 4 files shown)


src/plugins/rehype-wbr.js
```diff
@@ -0,0 +1,32 @@
+import { visit } from 'unist-util-visit';
```
borland, line 1: Claude wrote this. We need it because css will not wrap on full stops within a code block.

I.e. if you have `Octopus.Long.Thing.Oh.So.Very.Long` then it will never wrap and just make the table cell that it's in wider until it falls off the right edge of the column.

We can set the css `overflow-wrap: anywhere` property, but then it wraps at random points such as halfway within the word "Long".

A solution is to use a processing astro plugin which finds word breaks within a `code` block and inserts the `<wbr>` tag so the browser will wrap at those points

team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3278.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3278&status=all).

src/styles/vars.css
```diff
@@ -128,9 +128,6 @@
 
   --duration-default: 300ms;
 
-  --code-color: var(--color-text);
-  --code-background: var(--colorBackgroundSecondaryDefault);
```
borland, line 132: We discussed before whether to clean up these old vars and move directly to the design system tokens, or to leave them there... and I think I said leave them there, because we want to be disciplined about not changing too many things at once and introducing risk.

However, these `--code-*` vars were only used in a tiny number of places (all contained within this PR) so it wasn't risky to clean them up. I didn't touch other things like `--color-text` though as the blast radius is much larger for those

enf0rc3 (approved):
LGTM, checked against the spec, demo site looks good the border radius is a big improvement

src/styles/vars.css
enf0rc3: +1 remove vars that now have direct design tokens, especially for these edge case ones

mandymichael:
This is a nitpick but in octopus server we add a little tiny bit of spacing on the left and right so it doesnt just abruptly end at the  text for inline code 
e.g.
https://design-system-components.octopus.design/?path=/docs/components-data-codedecorator--docs

src/styles/main.css
```diff
@@ -108,26 +108,27 @@ blockquote cite {
...
 /* TODO: Refactor code snippets. This needs to be a component. */
 code {
-  font-family: var(--code-font);
-  background-color: var(--navy-200);
+  font-family: var(--fontFamilyCode);
+  border-radius: var(--borderRadiusSmall);
+  border: var(--borderWidth1) solid var(--colorBorderPrimary);
+  background: var(--colorBackgroundSecondaryDefault);
+  white-space: pre-wrap;
+  overflow-wrap: break-word;
```
mandymichael, line 116: I'd suggest adding `padding: var(--space4)` here so its got a bit extra spacing between the text and the edge of the background.

PR #3296, opened 2026-08-05
Component: breadcrumbs

enf0rc3:
Closes [NES-277](https://linear.app/octopus/issue/NES-277/component-breadcrumbs).

Migrates the breadcrumbs to idiomatic Astro and applies the [Figma design](https://www.figma.com/design/DTvlouW1QArI3ZTf1PXs4J/Documentation-vision?node-id=1528-43711&m=dev).

Had to do a bit of pre-emptive margin / spacing work on the body content to get it to look "normal", anything to do with this will likely be resolved when we come to do those components / layouts.

## Idiomatic Astro

The breadcrumb rules had lived in `main.css` since the 2023 site import. Astro scopes and bundles component `<style>` blocks, so they now sit next to the markup they style and `main.css` loses 55 lines. Scoping also makes the `.site-breadcrumbs` prefix redundant, and nothing else in the repo referenced that class once the rules moved, so the class is gone too — `nav[aria-label="Breadcrumb"]` is the selector if tests ever need one.

## The design

Every value maps onto a design system token, so both themes come for free through `[data-theme]` and the dark variant in Figma needs no separate rules:

| Figma | Token |
| --- | --- |
| `space/8` | `--space8` (gap between crumbs and separators) |
| `space/6`, `space/0` | `--space6`, `--space0` (6px above, 0 below) |
| `text/body/regular/medium` | `--textBodyRegularMedium` (link text) |
| `text/body/regular/small` | `--textBodyRegularSmall` (separator) |
| `color/text/secondary` | `--color-menu-link` |
| `color/text/link/default` | `--color-menu-link-alt` |
| `color/text/tertiary` | `--color-text-tertiary` (new alias in `vars.css`) |

The colours needed no new plumbing — `vars.css` already pointed `--color-menu-link` and `--color-menu-link-alt` at the exact tokens the design uses for the rest and hover states. Only the separator colour was missing an alias.

The Font Awesome chevron is replaced by the design's `/` text separator, so this drops one more `fa-solid` dependency from the page.

The list and each item share the same 8px gap: the list spaces one crumb from the next, each item spaces its own separator from its link. That reproduces the design's uniform rhythm without adding non-`ListItem` children to the `BreadcrumbList`.

## Heading alignment

Per [node 1540-63194](https://www.figma.com/design/DTvlouW1QArI3ZTf1PXs4J/Documentation-vision?node-id=1540-63194&m=dev), the breadcrumb sits 56px below the header with the `<h1>` flush beneath it (breadcrumb at y=56 with height 26, headings at y=82).

That offset goes on the content column rather than on `.content-group`, because each column starts at a different height below the header — nav at 16px, content at 56px, table of contents at 100px — so there is no shared value to move. `.content-group`'s own `margin-top` is clearance for the fixed header and stays as it is.

## Two fixes along the way

- **`schema.org` positions were off by one.** `BreadcrumbList` positions are 1-based; they were emitted from a 0-based index, so the first crumb was `position="0"` and the markup was invalid for structured data.
- **Empty markup when there are no crumbs.** The design has a `levels=None` variant; the component rendered an empty `<nav>`/`<ol>`, and now renders nothing.

Also dropped the `a[aria-current]` colour override — the design shows every crumb in the secondary colour with no distinct current-page treatment, and `aria-current` still conveys it to assistive tech.

## Verification

`astro build` green: 2697 pages in 3m 12s. Checked the emitted output rather than just the dev server:

```
.content-group main{margin:var(--space56) 0 0;grid-area:content}
li[data-astro-cid-cvpdko3c]+li[data-astro-cid-cvpdko3c]:before{content:"/";color:var(--color-text-tertiary);font:var(--textBodyRegularSmall)}
```

Rendered crumbs on `/docs/administration/data/` are `Docs / Administration / Data` with `position` 1, 2, 3.

## Follow-ups, deliberately not in scope

- Nav column wants 16px and the TOC 100px. The TOC one has a bug behind it: `.side-nav { margin-block-start: 10vh }` is viewport-relative, so it lands at 80px on a short laptop and 144px on a tall monitor where the design says a fixed 100px. Two coupled `calc()` values hardcode the old offset.
- Below 1130px the grid restacks to `top / toc / content / menu`, so the 56px becomes separation between the TOC block and the content. The design node is 1920px and does not cover mobile.
- The last crumb still links to the current page; convention is a plain `<span>` for it. Changing that alters the RDFa `item`/`WebPage` structure, so it wants its own PR.
- `Breadcrumbs.astro` looks up each crumb with `allPages.find()` inside a loop over the crumbs, which is O(n·m) over every page at build time. A `Map` keyed by formatted URL would make it O(n+m).

### DIFF
--- a/src/themes/octopus/components/Breadcrumbs.astro
+++ b/src/themes/octopus/components/Breadcrumbs.astro
@@ -63,7 +63,7 @@ let metaIndex = navPages.length;
 stats.stop();
 ---
 
-<nav class="site-breadcrumbs" aria-label={_(Translations.aria.breadcrumbs)}>
+<nav aria-label={_(Translations.aria.breadcrumbs)}>
   <ol vocab="http://schema.org/" typeof="BreadcrumbList">
     {
       navPages.map((page, index) => (
@@ -113,8 +113,7 @@ stats.stop();
   }
 
   li::before {
-    content: '\f054';
-    /* Chevron Right */
+    content: '\f054'; /* Chevron Right */
     color: var(--color-menu-link);
     font-family: fa-solid;
     font-size: 0.8em;
--- a/src/themes/octopus/components/Breadcrumbs.astro
+++ b/src/themes/octopus/components/Breadcrumbs.astro
@@ -95,11 +95,6 @@ stats.stop();
 </nav>
 
 <style>
-  /* The list gap spaces one crumb from the next and the separator carries the
-     same 8px after itself, which keeps the rhythm uniform. list-style is here
-     because the items are plain list items: nothing in main.css resets ol
-     markers globally, and flex-wrap lets a long trail drop to a second line
-     rather than overflow the column. */
   ol {
     display: flex;
     flex-wrap: wrap;
--- a/src/themes/octopus/components/Breadcrumbs.astro
+++ b/src/themes/octopus/components/Breadcrumbs.astro
@@ -58,51 +58,41 @@ for (let i = 0; i < navPages.length; i++) {
   }
 }
 
-// schema.org BreadcrumbList positions are 1-based, so the nav pages occupy
-// 1..navPages.length and any explicit crumbs continue on from there.
-let metaIndex = navPages.length + 1;
+// The derived nav pages and any explicit crumbs render identically, so they
+// share one list. That keeps the schema.org position sequence, which is
+// 1-based, tied to a single index rather than two counters that have to agree.
+type Crumb = {
+  url: string;
+  title: string;
+  ariaCurrent?: string;
+  rel?: string;
+};
 
-const hasCrumbs = navPages.length > 0 || (breadcrumbs?.length ?? 0) > 0;
+const crumbs: Crumb[] = [...navPages, ...(breadcrumbs ?? [])];
 
 stats.stop();
 ---
 
-{
-  hasCrumbs && (
-    <nav aria-label={_(Translations.aria.breadcrumbs)}>
-      <ol vocab="http://schema.org/" typeof="BreadcrumbList">
-        {navPages.map((page, index) => (
-          <li property="itemListElement" typeof="ListItem">
-            <meta property="position" content={(index + 1).toString()} />
-            <a
-              property="item"
-              typeof="WebPage"
-              href={accelerator.urlFormatter.formatAddress(page.url)}
-              aria-current={page.ariaCurrent ? page.ariaCurrent : null}
-              rel={page.rel}
-            >
-              <span property="name">{page.title}</span>
-            </a>
-          </li>
-        ))}
-        {breadcrumbs &&
-          breadcrumbs.map((crumb) => (
-            <li property="itemListElement" typeof="ListItem">
-              <meta property="position" content={(metaIndex++).toString()} />
-              <a
-                property="item"
-                typeof="WebPage"
-                href={accelerator.urlFormatter.formatAddress(crumb.url)}
-                aria-current={crumb.ariaCurrent ? crumb.ariaCurrent : null}
-              >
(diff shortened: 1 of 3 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3296.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3296&status=all).

src/themes/octopus/components/Breadcrumbs.astro
```diff
@@ -58,44 +58,79 @@ for (let i = 0; i < navPages.length; i++) {
...
-            >
-              <span property="name">{crumb.title}</span>
-            </a>
-          </li>
-        ))
-    }
   </ol>
 </nav>
+
+<style>
```
borland, line 97: Inline CSS is typically considered an anti-pattern. Was there a reason why this stuff moved out of the main.css or did claude just randomly do it?

src/themes/octopus/components/Breadcrumbs.astro
```diff
@@ -58,44 +58,79 @@ for (let i = 0; i < navPages.length; i++) {
...
+// share one list. That keeps the schema.org position sequence, which is
+// 1-based, tied to a single index rather than two counters that have to agree.
+type Crumb = {
+  url: string;
+  title: string;
+  ariaCurrent?: string;
+  rel?: string;
+};
+
+const crumbs: Crumb[] = [...navPages, ...(breadcrumbs ?? [])];
```
borland, line 71: Nice. Running the loop twice was naff

src/themes/octopus/components/Breadcrumbs.astro
```diff
@@ -58,44 +58,79 @@ for (let i = 0; i < navPages.length; i++) {
...
-              <span property="name">{crumb.title}</span>
-            </a>
-          </li>
-        ))
-    }
   </ol>
 </nav>
+
+<style>
+  /* The 8px rhythm is uniform across links and separators, so both the list
```
borland: Delete this overly claude-y comment, it adds no value

borland (commented):
outcome on the staging site looks good. I'm not a fan of the big 56px gap above the breadcrumb, but it's pixel-perfect matching figma.

Holding on approval because of the inline CSS, otherwise LGTM

src/themes/octopus/components/Breadcrumbs.astro
enf0rc3: When in standard html + JS yes, however its the [ idiomatic way of doing it with `.Astro`](https://docs.astro.build/en/guides/styling/#styling-in-astro) components due to the build process that ensures the styles are only used in that given component. 

You would import css if there was multiple components that needed the styles

```
Best practice: inline <style> tag inside the component, not a separate imported file — for component-level styles.

Why:

Astro <style> tags are auto-scoped — no BEM naming, no class collisions. Separate imported CSS files are not scoped and can leak globally.
Co-location — style lives next to the markup it styles, so you're not context-switching files for a small component. Good for ADHD-brain workflow too — one file, one unit of thought.
Scoped styles always win the cascade (highest precedence, applied last) — predictable, no fighting specificity wars with imports.

When to use a separate imported file instead:

Global/site-wide styles (resets, design tokens, typography) — imported once in a Layout.
Shared CSS across many components (rare — usually design tokens/variables cover this).
```

borland (approved):

src/themes/octopus/components/Breadcrumbs.astro
borland: Ah, TIL. Approved

PR #3303, opened 2026-08-05
Make the heading anchor an interactive copy-URL button

enf0rc3:
[NES-282](https://linear.app/octopus/issue/NES-282/component-copy-url-button) · [CopyHeadingUrl design](https://www.figma.com/design/DTvlouW1QArI3ZTf1PXs4J/Documentation-vision?node-id=1749-15176)

Add the copy-url-button to all headings in `.page-content` as per spec:
image

## Results

`scripts/modules/headers.js` already created these elements on load, so it keeps doing that and now wires the behaviour as well:

```js
function enhanceHeaders() {
  addCopyButtons();   // one per heading
  addCopyListener();  // one delegated click handler for the page
}
```

Both glyphs come from `design-system-icons` and are byte-identical to `LinkOnIcon.tsx` and `CheckIcon.tsx`, on the `0 0 40 40` viewBox its `IconWrapper` uses. 

Example:
image
image

## Accessibility

Each heading gets an `aria-label` of its own text. A heading is named from its contents, and those contents include a nested control's accessible name, so every heading was otherwise announced with the button's label appended:

```
before   heading "Install Octopus Server Copy link to this section"
after    heading "Install Octopus Server"
```

That was true of the old anchor too, with `Link to this section`, so this closes a pre-existing defect.

## Also

Removes the heading samples from `index.md`. They existed to show the button in a forced-visible state, which is unnecessary now that it is interactive.

### DIFF
--- a/src/pages/index.md
+++ b/src/pages/index.md
@@ -26,40 +26,6 @@ This page is not shown on the production site.
 - [Oldest content](/report/oldest-content/1)
 - [Taxonomy](/report/taxonomy)
 
-## Headings \{#headings}
-
-In each pair below, the first heading is at rest and the second has the anchor
-button forced visible. Normally it appears on hover or keyboard focus.
-
-<!-- The page title renders as an h1 from frontmatter, so this sample is a second
-     one. Deliberate - this page exists to show the heading styles. -->
-<!-- markdownlint-disable MD025 -->
-
-# Heading 1 (title)
-
-<!-- markdownlint-enable MD025 -->
-
-## Heading 2 \{#heading-2-a}
-
-## Heading 2 \{#heading-2-b}
-
-### Heading 3 \{#heading-3-a}
-
-### Heading 3 \{#heading-3-b}
-
-#### Heading 4
-
-##### Heading 5
-
-###### Heading 6
-
-<style>
-  #heading-2-b .bookmark-link,
-  #heading-3-b .bookmark-link {
-    opacity: 1;
-  }
-</style>
-
 ## Body text \{#body-text}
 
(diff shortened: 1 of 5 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3303.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3303&status=all).

src/scripts/modules/headers.js
```diff
@@ -1,20 +1,121 @@
...
+    status = document.createElement('div');
+    status.className = 'copy-heading-url-status';
+    status.setAttribute('aria-live', 'polite');
+    document.body.append(status);
+  }
+
+  status.textContent = message;
+}
+
+function enhanceHeaders() {
```
enf0rc3, line 119: Called by main.js  after loading the page

src/pages/index.md
```diff
@@ -26,40 +26,6 @@ This page is not shown on the production site.
 - [Oldest content](/report/oldest-content/1)
 - [Taxonomy](/report/taxonomy)
 
-## Headings \{#headings}
```
enf0rc3, line 29: I didn't really like this so opted to remove it, we already have similar stuff on /components

borland (approved):
tested on the staging site, LGTM

enf0rc3:
Pushed trivial pixel-perfect placement of tooltip changes and re-tested. still centered.

image

PR #3301, opened 2026-08-05
Move the heading anchor icon out of public/ and into src/assets/

enf0rc3:
Follows the `Move the inline-SVG icon components out of public/` commit in #3280 — `link-on.svg` is a component asset whose only consumer is the `.bookmark-link` mask in `main.css`, so it belongs next to the stylesheet rather than in `public/`.

| | Before | After |
|---|---|---|
| file | `public/docs/icons/link-on.svg` | `src/assets/icons/link-on.svg` |
| `mask` url | `url('/docs/icons/link-on.svg')` | `url('../assets/icons/link-on.svg')` |

`main.css` is imported by `HtmlHead.astro:24`, so Vite processes its `url()` declarations and the relative path is what lets it bundle the file. At 995 bytes the SVG falls under the 4KB inline limit, so it ends up as a data URI inside `components.*.css` and costs no request at all.

`src/assets/` matches the convention #3267 introduces with `src/assets/octopus-logo-v2.svg`.

The favicons stay in `public/docs/icons/` — they are requested by absolute URL from the HTML head, and #3280 deliberately keeps them too.

Precursor to the CopyHeadingUrl work, which adds a second icon.

`prettier --write` also normalized combinator spacing at `main.css:174` and `:324`, in rules this change does not otherwise touch. Both were already failing `prettier --check` on `main`.

## Verification

- `pnpm build` exits 0, 2697 pages.
- The `mask` in `dist/docs/_astro/components.*.css` resolves to the inlined data URI; `dist/docs/icons/` now holds only the six favicons.
- `cspell` on the changed files and `prettier --check src/styles/main.css`: both clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3306, opened 2026-08-05
Extract the clipboard write shared by the copy features

enf0rc3:
Pure extraction, split out of #3303 so the heading work reviews on its own.

`copy-markdown.js` carries a clipboard write with two code paths: `navigator.clipboard.writeText` where a secure context allows it, falling back to a hidden textarea plus `document.execCommand('copy')` for HTTP and older browsers. The heading copy button in #3303 needs exactly the same logic, so it moves to `src/scripts/modules/clipboard.js` rather than becoming a second copy.

## What changed

- `writeToClipboard` and `execCommandCopyFallback` move out of the `CopyMarkdown` class into a module, unchanged apart from the warning prefix (`[copy-md]` → `[clipboard]`) and JSDoc types.
- `copy-markdown.js` imports it and calls `writeToClipboard(text)` in place of `this.writeToClipboard(text)`.

No behavior change: the same two paths are tried in the same order, and the return contract (`true` on success) is the same.

`code-blocks.js` has its own simpler inline version. It is bound up in a larger icon-swapping flow, so it is left alone — worth a follow-up if this module proves out.

## Verification

- `pnpm build` exits 0, 2697 pages, on this branch alone.
- The extracted code is bundled into the page script (`[clipboard]` appears in `dist/docs/_astro/Default.astro_astro_type_script_index_0_lang.*.js`), so the import resolves rather than being tree-shaken away.
- `prettier --check` and `cspell` on both files: clean.

Stacked on #3301.

### DIFF
--- a/src/scripts/modules/clipboard.js
+++ b/src/scripts/modules/clipboard.js
@@ -1,0 +1,57 @@
+// @ts-check
+
+/**
+ * Copies text to the clipboard, reporting whether it worked.
+ *
+ * navigator.clipboard needs a secure context, so this falls back to the
+ * deprecated execCommand path for HTTP and older browsers.
+ *
+ * @param {string} text
+ * @returns {Promise<boolean>}
+ */
+async function writeToClipboard(text) {
+  if (
+    typeof navigator !== 'undefined' &&
+    navigator.clipboard &&
+    typeof navigator.clipboard.writeText === 'function' &&
+    (typeof window === 'undefined' || window.isSecureContext !== false)
+  ) {
+    try {
+      await navigator.clipboard.writeText(text);
+      return true;
+    } catch (err) {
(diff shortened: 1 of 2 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3306.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3306&status=all).

enf0rc3:
Superseded by #3303. navigator.clipboard.writeText has been Baseline since 2020, so the shared module and its execCommand fallback are unnecessary - the four-line write is inlined at the call site instead.

PR #3313, opened 2026-08-06
Replace the magnify-and-new-tab figure with an inline lightbox

enf0rc3:
Closes [NES-284](https://linear.app/octopus/issue/NES-284/component-lightbox).

Clicking the magnifier called `window.open`, dropping the reader on a bare image file. Figure images now click to zoom in place.

Modelled on [Linear's docs](https://linear.app/docs/conceptual-model), which the designer referenced, and matching the Figma frames under **Lightbox** (`1848:17946`, `1848:18049`), plus a close button.

## Two files

```
src/scripts/modules/figures.js | 194 +++++++++++++++++++++++++++++-----
src/styles/main.css            | 156 +++++++++++-------------------
```

No markup changes — the emitted HTML is byte-identical to `main`. Nothing about how images are served changes: no new files, no image-pipeline changes, `public/` untouched.

## No new dependency

I started on PhotoSwipe, then read Linear's shipped CSS — their whole lightbox is 7 rules and no vendor library.

I did go back and build the PhotoSwipe version properly once review raised scroll-locking and the close button — it's on `willlaugesen/nes-284-photoswipe-spike` and it works. We're not taking it yet, because this version animates `left/top/width/height` rather than `transform: scale()`, and that difference matters: a transform scales its whole subtree, so the 8px radius drew at 43px mid-zoom and the 1px border at 5px, both snapping back at the end. Avoiding that in PhotoSwipe needs counter-scaling the radius per frame. Here it's free.

The spike stays available if we later want pinch-zoom on touch, which this doesn't do.

## How it works

**`figures.js`** reads the image's rect, clones it into a fixed frame at that rect, then animates to a fit-viewport rect over an opaque `var(--color-base-primary)` backdrop — already theme-aware, so dark mode is free. `prefers-reduced-motion` skips the animation.

The image carries the interaction directly. Linear uses a wrapping `<div aria-label="Zoom image">`, which isn't focusable and gives keyboard users nothing; `img.tabIndex = 0` plus an Enter/Space handler is what a wrapper would otherwise provide, without touching the markup.

**`main.css`** loses the `.image` frame, the `.image__img` shadow, the legacy `figure:has(p > img) p` duplicate (which carried a `TODO` to remove it), and all six `.magnify-*` rules. What replaces it is one `figure img` rule covering both the markdown and `<Image>` paths — the duplication that `TODO` was waiting on is gone without migrating 1,452 blocks to a component.

## Review feedback, all addressed

| | fix |
|---|---|
| page scrolled behind the lightbox | `overflow: hidden` plus permanent `scrollbar-gutter: stable`, so hiding the scrollbar reflows nothing |
| no visible way out | close button top right, takes focus on open, and is removed the moment closing starts |
| low-res image blown up | the clone names the widest `srcset` file outright, and hovering prefetches it |
| white images have no edge | `border: var(--borderWidth1) solid var(--colorBorderPrimary)`, inline and zoomed |

## Verified

Full build (2,673 pages, clean), then driven with Playwright against the built site, including under 300KB/s with 300ms latency:

| | |
|---|---|
| resting cursor | `zoom-in` |
| overlay | `rgb(255,255,255)` light, `rgb(17,26,35)` dark |
| lightbox cursor | `zoom-out` |
| border | `#DEE1E6` light, `#2E475D` dark, no overflow (799.97px in an 800px column) |
| scroll while open | moved **0px**; body width 1425px before, during and after |
| image served | 1432px file behind a 1340px box, correct from the first frame |
| close | button, click and Escape; button gone within 30ms of closing |
| keyboard | Enter opens, focus lands on close, returns to the image after |
| both paths | markdown `:::figure` and the `<Image>` MDX component |
| console errors | none |

## Worth knowing

**Mobile pinch-zoom.** The Figma frames are 1920×1080 only. This fits to viewport and stops there, so a dense screenshot is still hard to read on a phone. Linear has the same limitation. The PhotoSwipe spike solves it without changing this markup.

**`naturalWidth` lies on srcset images.** It returns a density-adjusted size — the badges thumbnail reports 842 where the file is genuinely 1432. It nearly read as an upscaling bug. The pages with `srcset` also carry a real `width` attribute, which takes priority in the upscale cap, so behaviour is correct.

**Upscale cap.** A narrow screenshot opens at its true pixel width rather than stretching, which keeps it sharp.

### DIFF
--- a/src/scripts/modules/figures.js
+++ b/src/scripts/modules/figures.js
@@ -123,6 +123,10 @@ function open(img) {
     document.removeEventListener('keydown', onKey);
     overlay.classList.add('zoom-overlay--closing');
 
+    // The way out has been taken, so the button has no job for the length of
+    // the animation. Leaving it up reads as the lightbox failing to close.
+    close.remove();
+
     let closed = false;
     const done = () => {
       if (closed) return;
--- a/src/scripts/modules/figures.js
+++ b/src/scripts/modules/figures.js
@@ -1,26 +1,144 @@
-import { qs, qsa } from './query.js';
+import { qsa } from './query.js';
 
-const activeClass = 'magnify-icon';
+const inset = { wide: 50, narrow: 16 };
+const duration = 400;
+
+/** @type {(() => void) | null} */
+let dismiss = null;
+
+function stillFrames() {
+  return window.matchMedia('(prefers-reduced-motion: reduce)').matches;
+}
 
 /**
- * Enables opening image in new tab
+ * The rect the image animates to: its own aspect ratio, centred in the
+ * viewport, never scaled past the source image's real pixel width.
+ *
+ * @param {HTMLImageElement} img
+ */
+function targetRect(img) {
+  const padding = window.innerWidth < 640 ? inset.narrow : inset.wide;
+  const room = {
+    width: window.innerWidth - padding * 2,
+    height: window.innerHeight - padding * 2,
+  };
+
+  const ratio = img.naturalWidth / img.naturalHeight;
+  const native = Number(img.getAttribute('width')) || img.naturalWidth;
+
+  let width = Math.min(room.width, room.height * ratio, native);
+  let height = width / ratio;
+
+  return {
+    left: (window.innerWidth - width) / 2,
+    top: (window.innerHeight - height) / 2,
+    width,
+    height,
+  };
+}
+
+/**
+ * @param {HTMLElement} node
+ * @param {{ left: number, top: number, width: number, height: number }} rect
+ */
+function place(node, rect) {
+  node.style.left = `${rect.left}px`;
+  node.style.top = `${rect.top}px`;
+  node.style.width = `${rect.width}px`;
+  node.style.height = `${rect.height}px`;
+}
+
+/**
+ * @param {HTMLImageElement} img
+ */
+function open(img) {
+  dismiss?.();
+
+  const overlay = document.createElement('div');
+  overlay.className = 'zoom-overlay';
+
+  const lightbox = document.createElement('div');
+  lightbox.className = 'zoom-lightbox';
+
+  const frame = document.createElement('div');
+  const clone = /** @type {HTMLImageElement} */ (img.cloneNode());
+
+  // A viewport-wide candidate beats the one picked for a column-width thumbnail
+  if (clone.srcset) {
+    clone.sizes = '100vw';
+  }
+
+  frame.appendChild(clone);
+  lightbox.appendChild(frame);
+  document.body.append(overlay, lightbox);
+
+  place(frame, img.getBoundingClientRect());
+  img.dataset.hidden = 'true';
+
+  if (stillFrames()) {
+    place(frame, targetRect(img));
+  } else {
+    frame.getBoundingClientRect();
+    requestAnimationFrame(() => {
+      frame.style.transition = `all ${duration}ms ease-out`;
+      place(frame, targetRect(img));
+    });
+  }
+
+  /** @param {KeyboardEvent} event */
+  const onKey = (event) => {
+    if (event.key === 'Escape') dismiss?.();
+  };
+
+  dismiss = () => {
+    dismiss = null;
+    document.removeEventListener('keydown', onKey);
+    overlay.classList.add('zoom-overlay--closing');
+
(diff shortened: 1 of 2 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3313.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3313&status=all).

ellen-octopus:
So exciting! This looks fantastic 🤩 

@enf0rc3 I've realised we probably need a border because some images are white 😅 
Could you please add `border: var(--borderWidth1) solid var(--colorBorderPrimary);` to images?

borland:
Looking good!

Some UI feedback

- When the image is zoomed, it always fills the screen, even if the original image is small. This looks kind of weird; should we put a max zoom level in or something?
  - Alternatively, maybe we shouldn't offer the lightbox for images below a certain size. The project badges page linked below is an example
  
- Despite the fact that the cursor swaps to a "minus" magnifying class, and the escape key works to close the lightbox, I found myself hunting for some sort of "Close" button and was a bit lost without one.

- The lightbox is just taking the existing inline image and moving it around. For responsive images, like on the project badges page (https://stoctodocspr3313.z22.web.core.windows.net/docs/projects/badges) we shouldn't just take the low-res version and blow it up, we should swap out for the full-size version. 

- When the image is zoomed, you can scroll the page underneath it with the scroll wheel. This is a common bug with lightboxes.

Taking the image-swapping, close and scroll-bugs into consideration, I feel like photoswipe is probably worthwhile.

enf0rc3:
Fixed the border, did a spike to implement the library but required way more customisation, was easier just to do the fixes recommended.

borland:
@enf0rc3 I had a play with the latest version.

The "X" button is cool, and the under-scrolling bug is fixed, and the subtle border looks nice

The image swapping doesn't seem to be there though. I went to the main page on /docs. The responsive image for the dashboard gets loaded into the page via `docs/i/2000/getting-started/dashboard.webp`, but when I click the lightbox I expect it to get swapped out for the original image at `docs/i/x/getting-started/dashboard.png`, but it just takes the 2000 webp and stretches it

src/scripts/modules/figures.js
```diff
@@ -1,26 +1,197 @@
...
+  qsa('figure img').forEach((img) => {
+    img.tabIndex = 0;
 
-    const magnify = document.createElement('button');
-    magnify.classList.add(activeClass);
-    magnify.title = 'Enlarge';
+    // Fetching the full size on the way to the click means the zoom does not
+    // spend its first frames showing the column-sized version stretched
+    const warm = () => {
+      const full = largest(img);
```
borland: Our responsive image thing renders image tags like this:

```
<img src="/docs/i/x/getting-started/dashboard.png" alt="Octopus Dashboard" srcset="/docs/i/600/getting-started/dashboard.webp 600w, /docs/i/1000/getting-started/dashboard.webp 1000w, /docs/i/2000/getting-started/dashboard.webp 1500w" sizes="(min-width: 1680px) 1000px, (min-width: 940px) calc(71.81vw - 192px), calc(100vw - 32px)" class="resp-img" width="1500" height="736" style="--shown: 1;" tabindex="0">
```

The original/largest image is just in the `src` attribute. We should ignore the `srcset` when looking to swap in a full-sized replacement

borland (approved):

src/scripts/modules/figures.js
enf0rc3: Thanks, have resolved, good idea! |

Ai slop always overcomplicates!

enf0rc3:
Hi @borland I have made the changes regarding the full src.

> The image swapping doesn't seem to be there though. I went to the main page on /docs. The responsive image for the dashboard gets loaded into the page via docs/i/2000/getting-started/dashboard.webp, but when I click the lightbox I expect it to get swapped out for the original image at docs/i/x/getting-started/dashboard.png, but it just takes the 2000 webp and stretches it

This one doesn't have an image larger than 1500 so you would need to start on a smaller browser window to recreate this one e.g:
image

The image below this one on the dashboard page goes up to 3k and works.

PR #3316, opened 2026-08-06
Regenerate image derivatives that img.mjs had been skipping

enf0rc3:
Split out of #3313, where these files were surfaced. They fix a pre-existing gap in the asset pipeline and stand on their own.

## The trap

`img.mjs` only processes a source image when it has **no** metadata sidecar:

```js
// Only processes images where there is no json metadata file
if (!fs.existsSync(metaPath)) {
  filesToProcess.push(info);
}
```

Once a sidecar exists without its derivatives, nothing regenerates them. Three images were stuck that way.

## What was stuck

**`releases/timeline/environment-timeline.png` and `-button.png`** were renamed from `timeline.png`. The sidecar travelled with the source, so `img.mjs` skipped both, the derivatives kept the old name, and the sidecar kept the *old* image's dimensions — 1000×535 recorded against an image that is actually 1644×1081. Any consumer trusting those numbers was laying out the wrong box.

**`api-and-integration/github/github-connections-authorize-app.png`** had a sidecar and no derivatives at any size. That's the state you get when `img.mjs` exits mid-write: its `toFile()` calls at lines 105, 111, 116, 147 and 152 are never awaited, while the sidecar write at line 130 is.

## What this does

Deletes the three stale sidecars so `img.mjs` reprocesses them, and commits the 12 derivative files it produced plus the three corrected sidecars.

The old `timeline-*.webp` derivatives are unreferenced now, but I left them rather than widening this into a cleanup.

## Not fixed here

- **The unawaited `toFile()` calls.** `img.mjs` carries a header saying it comes from Astro Accelerator and local edits get overwritten, so the race wants its own decision.
- **`.jpeg` sources.** `img.mjs` writes `.jpg` derivatives while `getDestination` builds the fallback URL from the original extension, so `/docs/i/x/…jpeg` never exists for the three `projects/dashboard` images. #3313 handles this by falling back to the original file.

## Verifying

Nothing references these paths on `main` today — `getImageInfo` runs for `:img{}` directives only, and none of these images use one. So `main` stays green either way; this makes the assets correct for anything that starts requesting them.


borland (changes requested):
This PR is redundant when my cleanup PR merges. There's no value in regenerating the image derivatives in the meantime because nothing references them

team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3316.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3316&status=all).

enf0rc3:
Not required as https://github.com/OctopusDeploy/docs/pull/3281 fixes the need and we will use the new approach

PR #3317, opened 2026-08-07
Give code blocks a header with a language and copy button

enf0rc3:
NES-285 — the [CodeBlock component](https://www.figma.com/design/DTvlouW1QArI3ZTf1PXs4J/Documentation-vision?node-id=1548-67137) from the Documentation vision file.

Every fenced code block gets a header carrying an optional label, its language, and a copy button.

## What changed

**The shell is static HTML.** `src/plugins/shiki-code-block.js` is a Shiki transformer that emits the frame, header, label, language and copy button at build time, for all 1,599 fences across `.md` and `.mdx`. With scripting off, a block still renders complete.

**Labels come from the fence meta.** ` ```powershell Rename a deployment target ` becomes the block's label. Shiki drops that string, so the transformer keeps it. Most blocks have no label yet and render with the language and copy button alone — labelling the rest is content work for follow-up PRs.

**Several languages become a `<select>`.** A `<details data-group>` set whose panels are each a lone code block folds into one block with a language switcher. Groups holding prose as well stay tab lists.

**Long blocks collapse** at 500px behind a gradient fade. Click the code to expand, click away to collapse.

**Dark mode gets real syntax colours.** Shiki now highlights against `light-plus` and `dark-plus`, replacing the `invert(98%) hue-rotate(180deg)` filter.

**One copy button module.** `copy-button.js` backs both the code block and the heading copy-URL button, which were carrying ~55 duplicated lines. `headers.js` drops from 125 lines to 52.

## Pages to test

| Page | What to look for |
|---|---|
| [`/docs/kubernetes/steps/kustomize`](https://octopus.com/docs/kubernetes/steps/kustomize) | Header with a label, language, copy button |
| [`/docs/octopus-rest-api/octopus.client/using-resources`](https://octopus.com/docs/octopus-rest-api/octopus.client/using-resources) | Language switcher, `.md` |
| [`/docs/octopus-rest-api/octopus.client/getting-started`](https://octopus.com/docs/octopus-rest-api/octopus.client/getting-started) | Language switcher, `.mdx`, long option label |
| [`/docs/octopus-rest-api/octopus.server.exe-command-line/configure`](https://octopus.com/docs/octopus-rest-api/octopus.server.exe-command-line/configure) | Collapsed block, fade, click to expand and away to collapse |
| [`/docs/kubernetes/targets/kubernetes-agent/permissions`](https://octopus.com/docs/kubernetes/targets/kubernetes-agent/permissions) | A mixed group that stays tabs |
| [`/docs/projects/variables/certificate-variables`](https://octopus.com/docs/projects/variables/certificate-variables) | Rebuilt table (12 rows were losing data) |

Try each in dark mode, and hover a heading to check the copy-URL button still behaves.

## Worth a look

- **The switcher is a native `<select>`.** The trigger matches the design; the list it opens is the browser's, so that part diverges from the Figma panel. @mandymichael — is that acceptable?
- **~490 blocks display "Text"** because their fence has no language. Accurate, and now visible.
- **Grouped blocks flash**: the static HTML has N blocks that JS folds into one. Removing that needs a rehype plugin, which needs `markdown.processor` untangled first.
- Code still wraps rather than scrolling horizontally, unchanged from before.

## Also in here

Markdownlint runs only on changed files, so touching these surfaced 93 pre-existing violations. Same files on `main` report the same 93. They are fixed: the `certificate-variables` table was genuinely broken, six fences had no language, and `output-variables` used bold text as headings above its fences.

## Testing

`tests/code-block.spec.ts` and `tests/copy-button.spec.ts` — 12 tests covering the header, the label, copying, switching language, keyboard operation of the select, tabs staying tabs, collapse and expand, and the shell rendering with JavaScript disabled.

## Before/after
image

image
image


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3317.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3317&status=all).

borland:
A visual bug I noticed on the staging site. (Note: I haven't looked at the code yet)

Load https://stoctodocspr3317.z22.web.core.windows.net/docs/octopus-rest-api/octopus.client/using-resources

The first code snippet appears, and "PowerShell" is the language.
Flip it to C# and the "C#" part isn't left-aligned. It's not centered either, I'm not sure what's going on with it

image

It also feels like there's a slight delay after changing the select box before the page content updates.
I'm not sure if this is just how browsers deal with select box changes... If it is, then so be it, but if it represents a problem with our code, we should fix it

borland (commented):
Looking at the code here, claude obviously wrote it all. At a high level it seems fine (no malware or bitcoin miners) but I'm not in position to nitpick the detail... and I don't think we need to either. I'd be happy to approve once you've taken a look at my other comment about text alignment and speed

borland (approved):
Thanks for fixing the bug. The "Show more" on the long code blocks looks excellent. Approved

enf0rc3:
Have rebased and re-tested:
image
image

PR #3318, opened 2026-08-07
Re-style the right-hand article navigation panel

borland:
# Results

Claude does all the coding here; I reviewed and re-steered it several times to improve things.

- Updates the styling of the right hand article nav panel to match the figma design
- Overhauls the javascript which hilights the selected area as you scroll. The existing code had several bugs and would sometimes hilight a section other than the one you had clicked on. This bug exists today on octopuscom/docs but it became easier to trigger with the slightly smaller font sizes in the style update.
  - It now updates instantly as you scroll, rather than on a 1000ms timer
  - It now reflects the position within the document accurately

# Before

image

# After

image

# Testing

Tested on macOS in MS Edge (chrome), Safari and Firefox

### DIFF
--- a/src/components/ArticleNav.astro
+++ b/src/components/ArticleNav.astro
@@ -45,6 +45,14 @@ stats.stop();
             </li>
           ))}
         </ol>
+        {/* Slid onto the current heading's link by scripts/modules/toc.js,
+            which measures the links against this element's offset parent —
+            the <details> above. */}
+        <span
+          class="article-nav__indicator"
+          data-article-nav-indicator
+          aria-hidden="true"
+        />
       </details>
     </nav>
   )
--- a/src/components/ArticleNav.astro
+++ b/src/components/ArticleNav.astro
@@ -2,7 +2,6 @@
 import { accelerator } from '@lib/accelerator';
 import { SITE } from '@config';
 import { Translations, Lang } from '@util/Languages';
-import Separator from './Separator.astro';
 
 const stats = new accelerator.statistics('octopus/components/ArticleNav.astro');
 stats.start();
@@ -31,7 +30,6 @@ stats.stop();
         <summary class="article-nav__title">
           {_(Translations.toc.title)}
         </summary>
-        <Separator />
         <ol class="article-nav__list">
           {headings.map((heading) => (
             <li
--- a/src/components/Separator.astro
+++ b/src/components/Separator.astro
@@ -1,5 +1,0 @@
----
-// No props are needed
----
-
-<hr class="separator" />
--- a/src/pages/components.mdx
+++ b/src/pages/components.mdx
@@ -18,7 +18,6 @@ import Card from 'src/components/Card.astro';
 import IconTile from 'src/components/IconTile.astro';
 import Image from 'src/components/Image.astro';
 import Link from 'src/components/Link.astro';
-import Separator from 'src/components/Separator.astro';
 import TopNav from 'src/components/TopNav.astro';
 
 ## Component Usage Guide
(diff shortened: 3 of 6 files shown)


src/components/ArticleNav.astro
```diff
@@ -2,7 +2,6 @@
 import { accelerator } from '@lib/accelerator';
 import { SITE } from '@config';
 import { Translations, Lang } from '@util/Languages';
-import Separator from './Separator.astro';
```
enf0rc3, line 5: Can this be deleted now

team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3318.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3318&status=all).

src/components/ArticleNav.astro
borland: Yeah looks like it's unused except for the components demo page.

I think I'd prefer to leave it though, in case we need to put it back for some reason, and file a card to clean it up at the end, if that's OK?

enf0rc3:
Looks all good overall, would be happy with merge, one thing i did notice is we have lost toe header highlighted when its expanded and the hover affect like the original, also the carat is facing right when its collapsed instead of down saying, click to pull the menu down, all trivial.

image
image

can see here, side by side view (zoom in on both to trigger the responsive view.)
Before: https://octopus.com/docs/installation
After: https://stoctodocspr3318.z22.web.core.windows.net/docs/installation

Oh and there is also a c-spell error

src/components/ArticleNav.astro
borland: Actually, given that the spelling error caused me to need to re-spin the PR, I deleted the component.

Claude keeps writing comments in the CSS which say "colour" but our spelling requires AMERICAN :-(

borland:
> one thing i did notice is we have lost toe header highlighted when its expanded and the hover affect like the original

I'm not sure what this means?

> also the carat is facing right when its collapsed instead of down saying, click to pull the menu down, all trivial.

I didn't realize that the article nav event did anything in the mobile layout! Upon seeing it though, the change is intentional; In the figma it states for the nav that unexpanded sections should have a rightward-facing arrow. Down means expanded. The previous behaviour was wrong. Good job claude

ellen-octopus:
Very nice @borland 👏🏻 

I have two hopefully small bits of feedback:
1. Can we make the gap between the body content and the article navigation 80px?
CleanShot 2026-08-07 at 15 24 01@2x

2. This is a nice-to-have, so happy to put this in the backlog to address as polish, but would it be possible to make the active nav transition smooth - like the way [Linear](https://linear.app/docs/estimates) does it? 

https://github.com/user-attachments/assets/a68005b0-39b5-4abd-ae70-e1ddff8a06d8

enf0rc3 (approved):
Happy with this, did in person review found the UI is consistent with best practice.

Ellen did point out two minor things i would implement, smooth transition animation of the sidebar, now she mentions it it does look choppy. 

Also the 80px, that could be considered apart of layout, but may as well do it now. 

Feel free to merge those above changes without re-review

borland:
Hey @ellen-octopus - I got claude to make those suggested changes. The position indicator in the right-nav now animates smoothly, and the right-nav now anchors 80px to the right of the content column.

An additional change I made was that the content column and right-nav are now **centered together as a group.**, so the actual content is slightly left of center. 

Claude offered this as a suggestion, and I took it, because that's what Linear's docs do, and seeing it in action, it's a lot nicer now that the nav is anchored to the content.

I'll merge this in the interests of unblocking @enf0rc3, who has other changes that rely on the right-nav waiting in the wings.

ellen-octopus:
> An additional change I made was that the content column and right-nav are now centered together as a group., so the actual content is slightly left of center.

Thanks @borland 🙌🏻  that was the intention, so that works out perfectly. Thanks so much, it looks fantastic 🎉

PR #3322, opened 2026-08-10
Add link support to Button component

rosslovas:
# Summary

The new `<Button>` component only supported rendering as a HTML `<button>` in its initial form, but links (`<a href=...>`) are another use case we want it to be able to fulfil; there was one introduced already in the form of the `Edit on GitHub` button, which had to be built up manually rather than making use of `<Button>` prior to this PR.

This PR adds support for that: when a `href` is supplied, it will be rendered as an `<a>` rather than a `<button>` as appropriate, while appearing identical either way.

# Results

No visual changes from existing buttons, just a new capability. See in action at https://stoctodocspr3322.z22.web.core.windows.net/components#links

It's also used by the `Edit on GitHub` button found on any page.

### DIFF
--- a/src/components/Button.astro
+++ b/src/components/Button.astro
@@ -3,8 +3,8 @@ import type { HTMLAttributes } from 'astro/types';
 
 type SharedProps = {
   label?: string;
-  icon?: string; // URL to an SVG or a FontAwesome class
-  iconEnd?: string; // URL to an SVG or a FontAwesome class
+  icon?: string; // URL to an SVG, or a class name e.g. a FontAwesome icon class name
+  iconEnd?: string; // URL to an SVG, or a class name
   size?: 'xSmall' | 'small' | 'medium';
   importance?: 'default' | 'loud';
   disabled?: boolean;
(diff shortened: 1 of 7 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3322.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3322&status=all).

enf0rc3:
> 🤖 Written by Claude (an AI assistant), run by @willlaugesen — worth sanity-checking the specificity numbers.

`:not(.btn)` guards the new `Button.astro`, but the legacy `.button` class stays unguarded, and the extra `:not()` takes the selector to (0,4,1). `a.button.button--primary` is (0,2,1), so it loses.

Currently broken on the live site: "Start for free" and the "Send feedback" label in `Feedback.astro:26` renders `--octo-blue` instead of `--button-primary-text`. Also affects `getting-started/index.mdx:19`, and `.octo-copy-md__option` (the "View as Markdown" `<a>`) once that deploys.

Suggestion — `:where()` contributes zero specificity, so the exclusions still work but the selector keeps its original weight:

```css
/* main.css:80 */
.page-content a:not(:where(.link, .top-nav__link, .btn)) {
  color: var(--octo-blue);
}

/* main.css:84-85 */
.page-content a:hover:not(:where(.top-nav__link, .btn)),
.page-content a:focus:not(:where(.top-nav__link, .btn)) {
  color: var(--body-link-color);
}

/* main.css:58 */
a:hover:not(:where(.btn)),
a:focus:not(:where(.btn)),
```

Future components can then be added to the `:where()` list without ratcheting specificity up each time.

src/styles/main.css
```diff
@@ -77,12 +77,11 @@ hr {
   border-bottom: 1px solid var(--header-separator-color);
 }
 
-.page-content a:not(.link):not(.top-nav__link) {
```
rosslovas, line 80: These exceptions unfortunately introduced a regression as it increased its specificity, causing it to cover more links than it used to such as the "Send feedback" link. [The where pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:where) is there to reduce specificity:
> The difference between :where() and [:is()](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:is) is that :where() always has 0 [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Specificity)

...so by applying it to all anchors with deliberately reduced specificity, this allows it to work as the intended fallback colouring for ordinary links, and more specific things like buttons with their custom styles no longer have to fight against it.

src/components/Button.astro
```diff
@@ -27,24 +33,150 @@ if (!label && !rest['aria-label'] && !rest['aria-labelledby']) {
...
-        <i class={`${iconEnd} btn__icon`} aria-hidden="true" />
-      ) : (
-        <img src={iconEnd} alt="" class="btn__icon" aria-hidden="true" />
-      ))
-  }
-</button>
+  {EndIconTag && <EndIconTag {...endIcon!.props} aria-hidden="true" />}
+</Tag>
+
+<style>
```
rosslovas, line 67: I've moved the styles out of `main.css` to here ([scoped styles](https://docs.astro.build/en/guides/styling/#scoped-styles)). This is what I wanted to do originally when introducing the button component, but it wasn't possible while our caching strategy was broken (cached HTML pages would point to CSS with a content hash that didn't exist if a newer version had been deployed so users would see broken pages). That's all fixed now, so we can finally do this and benefit from the specificity increase, fighting less with existing styles 🎉

src/styles/main.css
```diff
@@ -429,7 +428,25 @@ strong {
   font: var(--textHeadingLarge);
 }
 
-.copy-heading-url.btn {
+.copy-heading-url {
```
rosslovas, line 431: This was piggy-backing off the new button styles. Because they're now scoped to `Button.astro`, they're not available to it any more.

We can't really make these into `<Button>`s AFAIK because these are dynamically created with JS rather than server-side rendered, so instead I just copied the relevant styles and applied them directly to this in particular. This means there's a possibility for drift between this and `<Button>` but I think that's acceptable.

enf0rc3 (approved):
LGTM, start for free back to normal.

PR #3324, opened 2026-08-10
Move feedback into the right side-nav 

enf0rc3:
Replaces the Google Forms link at the bottom of the article with the widget from the [Figma design](https://www.figma.com/design/DTvlouW1QArI3ZTf1PXs4J/Documentation-vision?node-id=1540-61659): Yes/No vote → optional comment → thank you, sitting under the article nav.

This is a stop-gap, front-end only ux change rather than a wholesale change in the way we collect this data. There are benefits to improving the way we collect data e.g. splitting the questions and having the submit solely for the qualitative data, and having the yes / no on helpful use a different mechanism.

# Results

Submissions go to the same Google Form as before:

| Field | Sent |
|---|---|
| `entry.336432709` page | `<title> - <url>`, e.g. `Deploy with Kustomize - https://octopus.com/docs/kubernetes/steps/kustomize`. Keeps the hash, so you can see which section was open; drops the query string. |
| `entry.128617088` scale 1–5 | Yes → `5`, No → `1` |
| `entry.434783109` comment | textarea, or a space when empty (the form marks it required) |

Also in here:

- `.btn[aria-pressed='true']` added to `src/styles/button.css` so the chosen vote holds the pressed background. That file is new from #3331, so it is worth a look.
- The styles and the startup script are scoped to `Feedback.astro`. The widget no longer starts from `main.js`, so its code ships as its own bundle, requested only by pages that render it.
- 5 Playwright tests in `tests/feedback.spec.ts` covering the disclosure, the vote toggle, and both submit outcomes.

Known limits, all worth a follow-up:

- **Hidden below 1130px.** The table of contents column restacks above the article at that width, so the widget would ask the question before anything had been read. The old link sat in the article body and so worked on mobile — mobile readers now have no way to give feedback at all. Needs a design call.
- Google Forms sends no CORS headers, so the POST is `no-cors` and the response is opaque. An HTTP rejection is unreadable, so anything Google answers at all counts as accepted. A request that never leaves the browser — offline, or blocked by an extension — does reject, and the widget keeps the form up with Send live again rather than thanking the reader for nothing.
- **Nothing is captured until Send is pressed.** A reader who votes and leaves submits nothing. Posting on the vote instead would also need the Send path to stop creating a second response, so it is a bit more than a one-liner.

# Before (old feedback)
image

# After
## Before Vote:
image

## After Vote:
image

# TODO:
- [x] Test and confirm the new UI actually works (i suspect it may not if you need email or something)
- [ ] Decide what mobile readers get, given the widget is hidden below 1130px


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3324.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3324&status=all).

enf0rc3:
Confirmed the form submission worked:
image
image

src/scripts/modules/feedback.js
```diff
@@ -0,0 +1,98 @@
...
+ * the response comes back opaque. There is no way to read whether the form
+ * accepted it - a rejected submission looks identical to an accepted one, so
+ * anything Google answered at all counts as accepted. tests/feedback.spec.ts
+ * checks the form's fields in place of the response nobody can read.
+ *
+ * @param {string} rating
+ * @param {string} comment
+ * @returns {Promise<void>}
+ */
+async function submit(rating, comment) {
```
enf0rc3: This is really a "hack" to get the UI in before actually wiring up the real deal.

We should look at GTM or something to publish on click of thumbs up / down etc and maybe look at a different mechanism for the content rather than forms.

borland (approved):
LGTM. I tested it on the staging site and it appears to work

Note: I don't have access to the backend google form so I can't be sure, but if it turns out not to work we can fix it later

enf0rc3:
Thanks for testing: (with improved which page did you view)

Confirmed:
image

PR #3326, opened 2026-08-10
Document AddDays and AddHours date filters

NickJosevski:
Documents the `AddDays` and `AddHours` variable filters added in [Octostache 3.9.3](https://github.com/OctopusDeploy/Octostache/pull/124), shipping in Octopus Server **2026.3.11319**.

## Changes

`variable-filters.md` — an `### AddDays and AddHours` section under Date filters, plus rows in the date-filters table.

Covers negative amounts, fractional amounts (which must be quoted, since an unquoted filter argument can't contain a decimal point), time-zone preservation, and the echo-on-bad-input behaviour. Every value in the examples was produced by running the expression through the implementation rather than written by hand.

The section ends with a worked ServiceNow change-window example, and a warning that `Octopus.Task.QueueTime` is **queue** time rather than execution start — so a window derived from it starts counting down while the task is still queued.

## Second commit

Fixes pre-existing markdownlint violations in this file. CI lints only changed files, so editing the page surfaces violations that already fail on main. Scoped to what the linter actually reports — two misaligned tables, asterisk list markers, a hard tab, trailing whitespace, duplicate blank lines, a missing blank line after a heading. Each table keeps its own separator style, so the diff stays close to what the rules require.

## Note

An earlier revision of this PR also added the filters to the flat list in `variable-substitutions.mdx`. That list has since been removed on `main` — the page now links to this one for the full set — so the change was dropped when resolving the merge.


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3326.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3326&status=all).

sathvikkumar-octo (approved):

PR #3330, opened 2026-08-10
Add an Open in Claude split button to the page actions

enf0rc3:
Adds an **Open in Claude** split button to the page actions row, between Copy as markdown and Edit on GitHub, per the Figma ordering.

Implements the [Open in LLM](https://www.figma.com/design/DTvlouW1QArI3ZTf1PXs4J/Documentation-vision?node-id=1561-20228&m=dev) design, using the `SplitButton` component from #3340. Based on `main`, since that has merged.

The primary half opens Claude with a prompt pointing at the page's `.md` URL. The caret opens a menu with Open in Claude and Open in ChatGPT, as link-styled rows with a provider logo and an external-link icon. The prompt carries an absolute URL, because the assistant fetches the page itself rather than following a link from this site.

Gemini is deliberately absent: `gemini.google.com` ignores prompt parameters, so the entry could only prefill by pointing at AI Studio, which is not where "Open in Gemini" says it is going.

Four files, additions only — the component, its strings, one line in the layout, and its tests. The icons and `getPageMarkdownPath()` both come from what has already landed.

## Testing

Clean `astro build`, 1253 `.md` emitted. `tests/open-in-llm.spec.ts` 4 passed, and 17 passed alongside `split-button`. Playwright is not wired into CI, so that is a local run.

The positioning tests wait for the menu's open animation to settle, the same way `split-button.spec.ts` does — measured mid-animation the icon box reads 12x9 rather than the 16x16 it lands on.

Measured where the menu ends up, to confirm those tests are actually holding something down:

| viewport | menu | opens towards |
|---|---|---|
| 1200 | 641–834 | start edge |
| 700 | 321–514 | start edge |
| 430 | 159–353 | **flips to the end edge** |

430 is where the row wraps and the fallback fires, which is the case `split-button.spec.ts` could not reach on its own: the component gallery gives the control room on both sides at every width.

### DIFF
--- a/src/components/OpenInLlm.astro
+++ b/src/components/OpenInLlm.astro
@@ -1,0 +1,90 @@
+---
+import { SITE } from '@config';
+import { Lang, Translations } from '@util/Languages';
+import { getPageMarkdownPath } from '@util/mdxContent';
+import SplitButton from './SplitButton.astro';
+
+type Props = {
+  lang: string;
+};
+const { lang } = Astro.props satisfies Props;
+
+const _ = Lang(lang);
+
+const pageMdUrl = getPageMarkdownPath(Astro.url.pathname);
+
+// The assistant fetches the page itself, so the prompt carries an absolute URL
+// rather than the site-relative one the rest of the site links with.
+const prompt = pageMdUrl
+  ? _(Translations.octopus_open_in_llm.prompt).replace(
+      '{url}',
+      new URL(pageMdUrl, SITE.url).href
+    )
+  : '';
+
+const assistants = [
+  {
+    label: _(Translations.octopus_open_in_llm.open_in_claude),
+    href: 'https://claude.ai/new?q=' + encodeURIComponent(prompt),
+    icon: 'octo-llm__icon octo-llm__icon--claude',
+    iconEnd: 'octo-llm__icon octo-llm__icon--external',
+  },
+  {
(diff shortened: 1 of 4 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3330.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3330&status=all).

src/components/OpenInLlm.astro
```diff
@@ -0,0 +1,90 @@
...
+    --octo-llm-glyph-size: 0.8rem;
+  }
+
+  .octo-llm :global(.octo-llm__icon--claude) {
+    mask: url('../assets/icons/claude.svg') center / var(--octo-llm-glyph-size)
+      no-repeat;
+  }
+
+  .octo-llm :global(.octo-llm__icon--chatgpt) {
+    mask: url('../assets/icons/chatgpt.svg') center / var(--octo-llm-glyph-size)
```
enf0rc3, line 82: im thinking we need a global 'icons.css' for these, but not keen to do it now this pr has lived too long

enf0rc3:
Superseded by a fresh PR from the same branch, to clear leftover stack metadata from when this was based on `wl/split-button`. No content change.

PR #3331, opened 2026-08-10
Move button styles to a shared file that can be imported

rosslovas:
Multiple components could benefit from these styles. By breaking them out into their own file and using `@import`, we get the ability to share while keeping the benefits of scoped styles rather than global styles which run risks of conflicting with the arbitrary specificity of other global styles.


enf0rc3 (approved):
straightforward lift and shift, will unblock future buttons with shared styles, e.g. copy markdown and select LLM, thankyou! 

Approved pending smoke test on staging.

team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3331.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3331&status=all).

PR #3333, opened 2026-08-11
C# script (dotnet-script) steps require the .NET SDK

NickJosevski:
C# scripts (`.csx`) run through dotnet-script, which always shells out to `dotnet restore` — even for a script with no NuGet references — so the machine running the step needs the .NET **SDK**, not just the runtime. That machine is a deployment target, a worker, or the Octopus Server.

This has been true since the scriptcs → dotnet-script migration, but was only ever stated in a [2021 blog post](https://octopus.com/blog/rfc-migrate-scriptcs-dotnet-script) that is now stale on the version. A runtime-only target fails with:

```text
Unable to restore packages from '/root/.cache/dotnet-script/work/net8.0/script.csproj'
Make sure that all script files contains valid NuGet references
```

which points people at their NuGet references rather than the missing SDK. The docs now quote that message so a search for it lands here.

## Changes

- `deployments/custom-scripts/index.md` — new "C# script requirements" section
- `infrastructure/deployment-targets/linux/ssh-requirements.md` — its `.NET` section said the opposite, which is true for self-contained Calamari but not for C# scripts
- `projects/variables/system-variables.md` — only one NuGet source can be supplied via `Octopus.Action.Script.CSharp.NuGetSource`, and it replaces the default rather than adding to it

No SDK version is named. dotnet-script derives the generated project's target framework from whichever runtime the `dotnet` on the path resolves, so the requirement is relative to that runtime rather than a number that would go stale the way the blog post's `net6.0` did.

### DIFF
--- a/src/pages/docs/deployments/custom-scripts/index.md
+++ b/src/pages/docs/deployments/custom-scripts/index.md
@@ -31,6 +31,35 @@ Support for ScriptCS in Octopus will be removed from `2025.3`.
 To view previous and upcoming deprecations, please visit our [deprecations page](https://octopus.com/docs/deprecations).
 :::
 
+## C# script requirements {#csharp-requirements}
+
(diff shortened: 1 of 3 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3333.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3333&status=all).

zentron (approved):

PR #3335, opened 2026-08-11
Fill out WIP TopNav, make it possible to preview with ?newnav

rosslovas:
# Summary

This PR fills out some of the functionality in the new `TopNav` component:
- The correct page is highlighted in the navigation links (_Docs_/_API_/_CLI_)
- Some of the links are real now (still missing _Learn_ and _Changelog_)
- The theme switcher functions now (though it still has the wrong icons)

This PR also adds the ability for you to add `?newnav` to any URL and it will temporarily replace the old header with the new work-in-progress `TopNav`. It will remember this for an hour before automatically reverting back to the old one, or you can add `?newnav=0` to go back to the old one at any point. This will let us play with it in production before it's fully ready to go.

# Results

See it in action at https://stoctodocspr3335.z22.web.core.windows.net/docs?newnav

### DIFF
--- a/src/components/TopNav.astro
+++ b/src/components/TopNav.astro
@@ -4,20 +4,22 @@ import Avatar from './Avatar.astro';
 import Button from './Button.astro';
 
 type Props = {
-  active?: string;
   links?: { label: string; href: string }[];
 };
 
-// TODO: Make these real links
 const {
-  active = 'Docs',
   links = [
-    { label: 'Docs', href: '#' },
-    { label: 'Learn', href: '#' },
-    { label: 'API', href: '#' },
-    { label: 'CLI', href: '#' },
+    { label: 'Docs', href: '/docs' },
+    { label: 'Learn', href: '#' }, // TODO: Make this a real link
+    { label: 'API', href: '/docs/octopus-rest-api' },
+    { label: 'CLI', href: '/docs/octopus-rest-api/cli' },
   ],
-} = Astro.props;
+} = Astro.props satisfies Props;
+
+const pathname = Astro.url.pathname.replace(/\/+$/, '') || '/';
+const currentLabel = links
+  .filter(({ href }) => pathname === href || pathname.startsWith(`${href}/`))
+  .sort((a, b) => b.href.length - a.href.length)[0]?.label;
 ---
 
 <header class="top-nav">
@@ -35,10 +37,10 @@ const {
           <a
             class:list={[
               'top-nav__link',
-              label === active && 'top-nav__link--active',
+              label === currentLabel && 'top-nav__link--active',
             ]}
             href={href}
-            aria-current={label === active ? 'page' : undefined}
+            aria-current={label === currentLabel ? 'page' : undefined}
           >
             {label}
           </a>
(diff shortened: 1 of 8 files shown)


src/pages/docs/octopus-rest-api/index.mdx
```diff
@@ -26,12 +26,12 @@ We designed the Octopus REST API to:
 
 ## Octopus Command Line (CLI)
 
-The Octopus CLI is a command line tool that builds on top of the [Octopus Deploy REST API](https://octopus.com/docs/octopus-rest-api). With the Octopus CLI, you can push your application packages for deployment as either zip or NuGet packages, and manage your environments, deployments, projects, and workers.
+The Octopus CLI is a command line tool that builds on top of the [Octopus Deploy REST API](/docs/octopus-rest-api). With the Octopus CLI, you can push your application packages for deployment as either zip or NuGet packages, and manage your environments, deployments, projects, and workers.
```
rosslovas, line 29: Drive-by fixup just so these links work on staging sites

src/styles/main.css
```diff
@@ -648,7 +648,7 @@ nav.skip-links a:focus {
 .octo-header-bkg {
   box-shadow: 0 5px 30px rgba(0, 0, 0, 0.2);
   background-color: var(--header-bg);
-  min-height: 82px;
+  min-height: var(--header-height);
```
rosslovas, line 651: Made this a variable because the new nav has a different height, and there are a few bits and pieces that were hardcoding the old 82px height.

src/components/TopNav.astro
```diff
@@ -73,4 +84,83 @@ const {
 
 <script>
   import '../scripts/signed-in-user';
+  import '../scripts/theme-switcher';
 </script>
+
+<style>
```
rosslovas, line 90: Made these [scoped styles](https://docs.astro.build/en/guides/styling/#scoped-styles) just like I recently did with `Button`. I intended to do this all along, but I couldn't do it when I first added this because our caching strategy was broken and CSS with a content-hash that no longer exists could be pointed to by an old cached HTML page. That's all fixed now.

src/components/TopNav.astro
```diff
@@ -4,20 +4,22 @@ import Avatar from './Avatar.astro';
...
-// TODO: Make these real links
 const {
-  active = 'Docs',
   links = [
-    { label: 'Docs', href: '#' },
-    { label: 'Learn', href: '#' },
-    { label: 'API', href: '#' },
-    { label: 'CLI', href: '#' },
+    { label: 'Docs', href: '/docs' },
+    { label: 'Learn', href: '#' }, // TODO: Make this a real link
```
rosslovas, line 13: Not sure just yet what this is supposed to point to, I think the others are good though

src/components/TopNav.astro
```diff
@@ -4,20 +4,22 @@ import Avatar from './Avatar.astro';
...
+    { label: 'API', href: '/docs/octopus-rest-api' },
+    { label: 'CLI', href: '/docs/octopus-rest-api/cli' },
```
rosslovas, line 15: I've assumed CLI means the CLI that talks to an Octopus Server instance over HTTP, not the CLI of the Octopus Server binary itself, which is only applicable to self-hosted

src/components/TopNav.astro
```diff
@@ -47,14 +49,23 @@ const {
...
-    <!-- TODO: Make these real links -->
+    <!-- TODO: Fix the theme switcher icons -->
+    <Button
+      icon="fa-regular fa-moon"
+      aria-label="Dark mode"
+      aria-pressed="false"
+      data-theme-toggle-button
+    />
+    <!-- TODO: Make this a real link -->
     <Button label="Changelog" href="#" />
```
rosslovas, line 60: Not sure what this is supposed to point to just yet, I couldn't actually find release notes for non-self-hosted releases available anywhere

team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3335.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3335&status=all).

src/components/TopNav.astro
borland: I'm also assuming, but I think you're right. IMHO the `octopus.server` cli is an aspect of maintaining/administering your system, and not really a "CLI" for octopus in the way that the normal CLI is

borland (approved):
Feedback from testing.

When you click the button to switch between light and dark mode, the page body does it via a CSS animation.
This animation applies to the old nav, but not the new one, so the instant color-switch on it looks jarring.

Apart from that, everything LGTM. Approved so you can merge; Given this is in preview, we can ship some imperfections and clean them up in future PR's - or you can clean up now per your preference

rosslovas:
Thanks, yeah something didn't look right to me either and I see the missing transition now 👍 Since I have to sort out the icons anyway I'll address it in a follow-up PR :)

PR #3338, opened 2026-08-11
Add a Copy as markdown page action

enf0rc3:
Implements the [Copy to clipboard](https://www.figma.com/design/DTvlouW1QArI3ZTf1PXs4J/Documentation-vision?node-id=1561-20055&m=dev) design from Documentation vision.

Based on `main`. Part of a four-PR stack.

| PR | What | Base | State |
|----|------|------|-------|
| #3337 | markdown emitter posix paths | `main` | merged |
| **this** | Copy as markdown page action | `main` | |
| #3340 | SplitButton component | `main` | |
| #3330 | Open in Claude, old dropdown removed | `wl/page-actions` | |

## What changed

Under the page header, next to **Edit on GitHub**, there is now a **Copy as markdown** action. It fetches this page's `.md` companion and puts it on the clipboard, swaps to a check icon and "Copied" for two seconds, then reverts.

It is the `Button` component, so borders, radius, hover, pressed and focus states come from the design-system tokens.

Its two icon rules live in `main.css` beside `.octo-github-edit__icon`, which is the same Button-plus-icon-class arrangement. They cannot live in a scoped block in `CopyAsMarkdown.astro`: Astro stamps a component's scope on its own elements and on a child component's root, so the icon `<span>` inside `Button` never receives the parent's scope and a scoped selector cannot match it. Keeping them in the component would mean `:global()` on both, which is global CSS in a file that looks like it owns it.

The existing "Use Octopus docs with AI" dropdown at the foot of the article is left alone, so both offer to copy the page until #3330 replaces that dropdown. Two consequences for that window, both gone once #3330 lands: a page carries two ways to copy its markdown, and because the dropdown ships its own inline `aria-live` region, a page that has been copied from once carries two polite live regions. Renaming or trimming the dropdown here was considered and dropped — #3330 deletes the component, its script and its CSS outright, so the work would not survive the next PR in the stack.

Worth flagging for review: #3330 replaces the dropdown with Open in Claude / Open in ChatGPT and does not carry over its other two options, "Open this page as markdown" and "Open all docs as markdown". Losing those is a product call that belongs on that PR rather than this one, but it is easier to spot from here.

## Reusing the shared copy module

`copy-button.js` (from #3317) grows two things rather than this action reimplementing them:

- `showResult` writes to `.btn__label` when there is one, and falls back to the tooltip otherwise. The design wants the result in the label; the two existing icon-only consumers are unchanged and cannot do this.
- The result text is now translatable. `resultLabels()` reads `data-copied-label` / `data-error-label` off the button and falls back to the module's constants, so this button passes `octopus_copy_as_md.copied` and `.error` while the two icon-only consumers keep the defaults. The dropdown being replaced localized these three strings; the shared module hardcoded them, so a button showing its result as visible text would have regressed on that.
- `copyFetchedOnClick` covers text that has to be fetched. It hands `ClipboardItem` a **pending promise** so the click's user activation survives the request — awaiting the fetch first and then writing loses it in Safari, which is the trap the module's own comment warns about.

The delegated listener, result swap and announcement are shared by both entry points. The two write strategies stay apart on purpose: the promise trick only earns its keep when there is an await to survive, and text read off the page has none, so it keeps the simpler and more widely supported `writeText`.

The button also locks its measured width before the label shortens, so what sits beside it does not shuffle. Measured rather than hardcoded, so it holds for any translation. The lock is released on revert, so a later copy measures again instead of holding a width taken at an older zoom level.

A third button could have had its own module instead. It shouldn't: `status` in `copy-button.js` is the page's only `aria-live` region, and a second module scope means a second announcer for screen readers to track, plus a duplicate of the clear-then-set trick that makes two copies in a row both announce. The module header now records this.

The eligibility check moved to `pageMarkdownUrl()`, and `CopyMarkdown.astro` now calls it too, so the slug is derived in one place rather than two.

## Testing

`astro build` from a clean `dist`: exit 0, 2674 pages, 0 parse or write failures.

Checked that moving the icon rules to `main.css` did not change what ships: in `dist/docs/_astro/*.css` both `.octo-copy-as-md__icon` and the `[data-copied]` swap come out with their SVGs inlined as data URIs, structurally identical to `.octo-github-edit__icon`. `/docs/argo-cd` renders the button with `data-copy-md-url="/docs/argo-cd.md"`; the ineligible `/docs/kubernetes` has no occurrence of the class at all.

`tests/llm-endpoints.spec.ts` gains two tests: one that clicks the copy button and reads the clipboard back to confirm the page markdown actually arrives, and one that the action is absent on an ineligible page.

Note that nothing in `.github/workflows` runs Playwright, so these are local-only. The numbers below are from my machine.

**13 passed / 0 failed**, against a build carrying the emitter posix fix. That fix has since merged as #3337, so `main` has it and this branch picks it up on merge — the earlier caveat about a Windows checkout emitting only 847 of 1253 `.md` companions no longer applies. This branch is not itself rebased onto that commit, so a local run from this branch alone still needs `main` merged in first.

### DIFF
--- a/src/components/CopyAsMarkdown.astro
+++ b/src/components/CopyAsMarkdown.astro
@@ -24,15 +24,3 @@ const pageMdUrl = pageMarkdownUrl(Astro.url.pathname);
     />
   )
 }
-
-<style>
-  /* `:global()` crosses into Button's scope */
-  :global(.octo-copy-as-md__icon) {
-    background-color: var(--colorIconPrimary);
-    mask: url('../assets/icons/copy.svg') center / contain no-repeat;
-  }
-
-  :global(.octo-copy-as-md[data-copied] .octo-copy-as-md__icon) {
-    mask: url('../assets/icons/check.svg') center / contain no-repeat;
-  }
-</style>
--- a/src/components/CopyAsMarkdown.astro
+++ b/src/components/CopyAsMarkdown.astro
@@ -21,6 +21,8 @@ const pageMdUrl = pageMarkdownUrl(Astro.url.pathname);
       label={_(Translations.octopus_copy_as_md.copy)}
       size="small"
       data-copy-md-url={pageMdUrl}
+      data-copied-label={_(Translations.octopus_copy_as_md.copied)}
+      data-error-label={_(Translations.octopus_copy_as_md.error)}
     />
   )
 }
--- a/src/components/CopyAsMarkdown.astro
+++ b/src/components/CopyAsMarkdown.astro
@@ -1,0 +1,38 @@
+---
+import { Lang, Translations } from '@util/Languages';
+import { pageMarkdownUrl } from '@util/mdxContent';
+import Button from './Button.astro';
+
+type Props = {
+  lang: string;
+};
+const { lang } = Astro.props satisfies Props;
+
+const _ = Lang(lang);
+
+const pageMdUrl = pageMarkdownUrl(Astro.url.pathname);
+---
+
+{
+  pageMdUrl && (
+    <Button
+      class="octo-copy-as-md"
+      icon="octo-copy-as-md__icon"
+      label={_(Translations.octopus_copy_as_md.copy)}
+      size="small"
+      data-copy-md-url={pageMdUrl}
+    />
+  )
+}
+
+<style>
+  /* `:global()` crosses into Button's scope */
+  :global(.octo-copy-as-md__icon) {
+    background-color: var(--colorIconPrimary);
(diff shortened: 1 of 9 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3338.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3338&status=all).

PR #3340, opened 2026-08-11
Add a SplitButton component

enf0rc3:
A **SplitButton** component: a Button with a second, narrower half that opens a menu of related actions.

Built for the Open in Claude page action in #3330 — nothing on the site uses it yet. Docs and a live demo are on `/components`.

Both halves are the `Button` component, so the styling comes from the design-system tokens. `Button` picks up an `as` prop so the caret half can render as a `<summary>` and open the menu without script.

# Result

Test here: https://stoctodocspr3340.z22.web.core.windows.net/components#splitbutton


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3340.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3340&status=all).

src/components/SplitButton.astro
```diff
@@ -0,0 +1,267 @@
...
+          );
+        })
+      }
+    </ul>
+  </details>
+</div>
+
+<script>
+  // Astro bundles and dedupes this, so a page with several split buttons loads
+  // the controller once, and a page with none does not load it at all.
```
rosslovas: Minor: feels very arbitrary to have this comment here. It could equally be put on basically any other component that has a `<script>`, there's nothing specific here to split button nor is there any important reason it needs to be called out for this in particular. Anyone who wants to know about how this works can go to the docs to find out, this feels unnecessary

rosslovas (approved):
Looks good, but I found a few things that didn't seem right:
- The menu component this is supposed to be based on animates when it appears/disappears, but this implementation lacks that; I assume it should have that, unless it was specified otherwise somewhere I missed
- Said menu also disappears when you use tab or shift+tab to navigate off of the menu, but this implementation stays
- Tabbing off of the menu can be used to break the page a bit, rendering the menu in the middle of nowhere or behind other components. Here you can see I open the menu, scroll it off the page, press tab 3 times, then scroll around some more:

https://github.com/user-attachments/assets/a36176eb-8cd1-49ba-a3ea-f4f10d04dffb

Approving as what's here generally works, but appears to be lacking a few things + the aforementioned bug :)

rosslovas (approved):

PR #3341, opened 2026-08-11
Split "Copy as markdown" out of the AI dropdown into a page action

enf0rc3:
Copying the page is the action people reach for most, and burying it in a dropdown at the foot of the article put it three interactions away. It now sits in the page actions row beside "Edit on GitHub", built from the shared `Button` component so it matches the [design](https://www.figma.com/design/DTvlouW1QArI3ZTf1PXs4J/Documentation-vision?node-id=1561-20055&m=dev).

The dropdown keeps the two links it was always best at — opening this page and the whole corpus as markdown — so its component, script, styles and translation keys are renamed from `copy-md` to `markdown-links` to say so.

## What changed

- **New** `src/components/CopyAsMarkdown.astro` — `Button` at `size="small"` with a copy icon and the label "Copy as markdown". Rendered in `.page-actions` ahead of "Edit on GitHub".
- **New** `src/scripts/copy-as-markdown.ts` — fetches the `.md`, writes to the clipboard (keeping the `execCommand` fallback for insecure contexts), and sets a state attribute for 2s before reverting.
- **Renamed** `CopyMarkdown.astro` → `MarkdownLinks.astro` and `copy-markdown.js` → `markdown-links.js`, with `octo-copy-md*` → `octo-md-links*` and `data-copy-md-*` → `data-md-links-*`. The copy option is gone from the menu.
- **Translations split** into `octopus_markdown_links` (label, view_raw, download_all) and `octopus_copy_md` (label, copied, error).
- **Shared helper** `getPageMarkdownPath()` in `mdxContent.ts`, so both components derive the `.md` URL the same way.

## The button does not resize when it reports a result

Swapping the label text for "Copied" made the button narrower and reflowed the page actions row. All three labels are rendered and stacked in a single grid cell instead:

```css
.octo-copy-md__labels { display: grid; }
.octo-copy-md__label  { grid-area: 1 / 1; visibility: hidden; }
```

The cell is sized by its widest child, so the width holds in any language without measuring anything in JavaScript. `visibility` rather than `opacity` keeps the inactive labels out of the accessible name. `Button.astro` gained a `<slot />` (and a label guard that accepts slotted content) to allow a label that is markup rather than a string — additive, and no existing caller passes children.

This departs from the Figma copied state, which is drawn narrower than the rest state. The reflow reads as a glitch in motion.

## Testing

`astro build` passes (2674 pages, 1253 `.md` companions emitted).

`playwright test` — 123 passed. A new test in `llm-endpoints.spec.ts` measures the button's `boundingBox().width` either side of a copy and asserts it is unchanged.

Note for anyone running the suite locally: the clipboard specs are flaky under the default 11 parallel workers, with a different one failing each run as they contend for the single system clipboard. All 15 pass with `--workers=1`. That predates this branch.


src/components/CopyMarkdown.astro
```diff
@@ -1,87 +0,0 @@
...
-  ? docsPath.slice(subfolderPrefix.length)
-  : null;
-const eligibleSlugs = getEligibleSlugs();
-const pageMdUrl =
-  docsSlug && eligibleSlugs.has(docsSlug) ? docsPath + '.md' : null;
-const allDocsUrl = SITE.subfolder.replace(/\/$/, '') + '/llms-full.txt';
----
-
-{
-  pageMdUrl && (
```
enf0rc3, line 25: Split out the "Copy" line item, into its own button and left the existing markdown links

team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3341.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3341&status=all).

borland (approved):
The "Open this page as markdown" and "Open all pages as markdown" menu is still looking ugly there down the bottom of the page. Should we remove it now? Or is that for forthcoming PR's? Mentioning this because those options don't appear in the UI design, it looks like they should simply be gone

Otherwise LGTM, I tested on the staging site and everything seems good

enf0rc3:
@borland Yeah they they dont exist in the design, its pre-existing, I was wanting to determine their fate (outside of this pr / button) way as there is no other way to get to those links, we could potentially update it to use the upcoming split button on the copy, but again worth a separate discussion

PR #3343, opened 2026-08-12
Consolidate "copy" functionality across all usages

enf0rc3:
Follows on from #3341 and implements final styling as per design and consolidates copy logic across all copy buttons

The "Copy as markdown" action fetched the page and then wrote the result. That is the pattern `copy-button.js` already warned against in a comment: Safari grants a click one shot at the clipboard and spends it on the first `await`. The `execCommand` fallback needs the same user gesture, so both paths failed and the button did nothing on that browser.

`ClipboardItem` accepts a **promise** for the value precisely so the gesture survives the fetch — and that pattern serves a synchronous read just as well. Fixing the bug and unifying the three buttons is therefore the same change, not two.

## What changed

The write moves into `copy-button.js` as one implementation, tried in order:

1. `clipboard.write([new ClipboardItem({ 'text/plain': promise })])`
2. `clipboard.writeText(text)` — browsers without `ClipboardItem`
3. `document.execCommand('copy')` — insecure contexts, where `navigator.clipboard` is undefined

The heading and code block buttons gain step 3, which previously only the markdown button had; a docs page served over plain http could copy from the page action but not from a code block.

`copyOnClick` now takes a reader that may return a promise, plus an optional reporter that puts the result on the button and returns the message to announce:

```js
copyOnClick(selector, read, { show })
```

`headers.js:49` and `code-blocks.js:228` are **unchanged** — the tooltip reporter is the default. `copy-as-markdown.ts` drops from 101 lines to 37, keeping only its label-stack reporter and its fetch.

## Testing

`astro build` passes (2674 pages). `playwright test --workers=1` — **126 passed**.

All three buttons now have a test asserting what actually reaches the clipboard:

| button | assertion | where |
|---|---|---|
| heading URL | clipboard contains `#<heading-id>` | `copy-button.spec.ts:23` |
| code block | clipboard contains `kustomization.yaml` | `code-block.spec.ts:86` |
| Copy as markdown | clipboard equals the served `.md` | `llm-endpoints.spec.ts` (new) |

Two further new tests: the failure path (the `.md` fetch is aborted, and the button must show `data-failed` and not `data-copied`), and an assertion that the copy takes the `ClipboardItem` path rather than falling back — `copy-button.js` logs on its way down the chain, so a quiet console is the evidence.

**What I could not verify:** the Safari behaviour this fixes. `playwright.config.ts` runs Chromium only and I am on Windows. The fix follows the spec and the repo's own documented constraint, and the fallback chain means a browser that rejects step 1 still copies — but someone on a Mac or iOS should click the button on the preview build to confirm.

Note for anyone running the suite locally: the clipboard specs are flaky under the default 11 parallel workers, with a different one failing per run as they contend for the single system clipboard. Use `--workers=1`. That predates this branch.

### DIFF
--- a/src/scripts/modules/clipboard.js
+++ b/src/scripts/modules/clipboard.js
@@ -71,11 +71,17 @@ function showTooltipResult(button, ok) {
   const message = ok ? COPIED : FAILED;
   const rest = restLabel(button);
 
+  const clearState = () => {
+    delete button.dataset.copied;
+    delete button.dataset.failed;
+  };
+
+  clearState();
   button.dataset.tooltip = message;
-  button.dataset.copied = '';
+  button.dataset[ok ? 'copied' : 'failed'] = '';
 
   revertAfter(button, () => {
-    delete button.dataset.copied;
+    clearState();
 
     // Dropping the attribute starts the bubble fading, and swapping the label
     // underneath it would be visible for as long as that runs. A pointer or
--- a/src/scripts/modules/clipboard.js
+++ b/src/scripts/modules/clipboard.js
@@ -137,7 +137,7 @@ function execCommandCopy(text) {
   try {
     copied = document.execCommand('copy');
   } catch (error) {
-    console.warn('[copy-button] execCommand fallback threw', error);
+    console.warn('[clipboard] execCommand fallback threw', error);
     copied = false;
   }
   textarea.remove();
@@ -172,7 +172,7 @@ async function writeToClipboard(value) {
       await clipboard.write([new ClipboardItem({ 'text/plain': blob })]);
       return true;
     } catch (error) {
-      console.warn('[copy-button] clipboard.write failed, falling back', error);
+      console.warn('[clipboard] clipboard.write failed, falling back', error);
     }
   }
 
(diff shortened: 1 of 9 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3343.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3343&status=all).

src/styles/main.css
```diff
@@ -1813,6 +1813,7 @@ html[data-theme='light'] .theme-switcher__icon--dark svg path {
 .octo-copy-md__labels {
   display: grid;
   padding-inline: var(--space4);
+  text-align: start;
```
enf0rc3, line 1826: Sneaky hotfix: copy button ui improvement that was missed (it was centered)

src/scripts/modules/clipboard.js
```diff
@@ -0,0 +1,229 @@
...
+ * @param {HTMLElement} button
+ * @param {boolean} ok
+ * @returns {string}
+ */
+function showTooltipResult(button, ok) {
+  const message = ok ? COPIED : FAILED;
+  const rest = restLabel(button);
+
+  button.dataset.tooltip = message;
+  button.dataset.copied = '';
```
rosslovas: I don't think this is a regression but I think this means we still get a ☑️ even if it fails, right? Doesn't have to change with this PR, but it'd be nice to not have that mixed signal

rosslovas (approved):
Looks good including the rebase after my changes :)

PR #3344, opened 2026-08-12
Complete implementation of new theme switcher in new TopNav

rosslovas:
# Summary

This PR completes the theme switcher in the new `TopNav` which was in a WIP state prior to now:
- Changes the icons to the intended sun/moon icons
- Ensures the top nav and buttons gradually fade between the light and dark theme like most other things, rather than switching immediately
- Adds a "Toggle theme" tooltip on hover
- Fixes tooltips to have the intended animation when appearing/disappearing

# Results

https://github.com/user-attachments/assets/742446e0-815f-4b0c-8b3c-865f3a4140a6

And the fixes to existing tooltips:

https://github.com/user-attachments/assets/2868b15e-2c79-4685-9b4f-662721a773b3

Try it out at: https://stoctodocspr3344.z22.web.core.windows.net/docs?newnav


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3344.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3344&status=all).

src/styles/button.css
```diff
@@ -69,6 +72,11 @@ img.btn__icon {
   transform: translateY(-0.0625rem);
 }
 
+/* A toggle button that is currently on holds the pressed background. */
+.btn[aria-pressed='true']:not(:disabled, [aria-disabled='true']) {
+  background: var(--colorBackgroundPrimaryPressed);
+}
```
rosslovas, line 78: Reordered just so a pressed button still reacts to hover (as those rules come after). Not really related to this PR, just spotted it along the way and it didn't seem right.

src/styles/main.css
```diff
@@ -3118,19 +3134,51 @@ a.button.button--primary:active {
...
+  border-block-start: 0;
+  border-block-end: var(--tooltip-arrow-size) solid
+    var(--colorBackgroundInversePrimary);
+}
+
+[data-tooltip]::before,
+[data-tooltip]::after {
+  transition:
+    opacity 375ms cubic-bezier(0.165, 0.84, 0.44, 1),
+    translate 375ms cubic-bezier(0.165, 0.84, 0.44, 1);
```
rosslovas, line 3167: Tooltips were supposed to have an animation, but as originally implemented it was missing. This adds it, for both the new "Toggle theme" tooltip and the existing "Copy to clipboard" on code blocks.

src/scripts/theme-switcher.ts
```diff
@@ -76,8 +80,20 @@ function syncControls(theme: Theme) {
...
+let transitionTimer: ReturnType<typeof setTimeout> | undefined;
+
+function markTransition() {
+  root.setAttribute(THEME_TRANSITION_ATTRIBUTE, '');
+  clearTimeout(transitionTimer);
+  transitionTimer = setTimeout(() => {
+    root.removeAttribute(THEME_TRANSITION_ATTRIBUTE);
+    transitionTimer = undefined;
+  }, TRANSITION_MS);
+}
```
rosslovas, line 92: The reason for this is because buttons (and perhaps other similar things) should gradually fade their background colour between light and dark _when toggling the theme_, but they should not be gradually fading their background colour on mouse hover, and you can only have one transition that covers both. With this change, they will gain the transition briefly as the theme switches, but at all other times they don't have the transition so that the hover state behaves correctly.

src/scripts/modules/copy-button.js
```diff
@@ -46,9 +49,21 @@ function showResult(button, message) {
...
+        return;
+      }
+
+      timers.set(
+        button,
+        setTimeout(() => {
+          button.dataset.tooltip = rest;
+          timers.delete(button);
+        }, FADE_MS)
+      );
```
rosslovas, line 66: Now that the fade animation exists, the text needs to not switch from "Copied!" back to "Copy to clipboard" until an additional `FADE_MS` milliseconds has passed

src/components/TopNav.astro
```diff
@@ -49,11 +49,11 @@ const currentLabel = links
     </nav>
   </div>
   <div class="top-nav__trailing">
-    <!-- TODO: Fix the theme switcher icons -->
     <Button
-      icon="fa-regular fa-moon"
-      aria-label="Dark mode"
-      aria-pressed="false"
```
rosslovas, line 56: This change broke this button:

https://github.com/OctopusDeploy/docs/pull/3324/changes#diff-925adbca99b83579fd3a5036440082d645eac442c49ead881979a876ef108fb3R83-R87

...because it made the theme switcher visibly get stuck "pressed" while in dark mode, which was not the intention. Whether it was "pressed" or not was never meant to be visible. I've changed it so it's no longer a toggle button, so it's not affected by that change any more.

enf0rc3: Oh snap, man, everything is so "touchy" with CSS!

enf0rc3 (approved):
LGTM, Thanks for finding that bug with the theme switcher!

Keen to merge this before the "https://github.com/OctopusDeploy/docs/pull/3343" PR so i can get your transition changes and re-apply them and re test them.

src/components/TopNav.astro
rosslovas: Yeah no kidding! I don't blame you at all btw, I easily could've done the same, and this isn't in production yet either so it's harmless :)

PR #3355, opened 2026-08-13
Component: ApiBadge

enf0rc3:
Implements the ApiBadge design from Figma ([node 1528-43714](https://www.figma.com/design/DTvlouW1QArI3ZTf1PXs4J/Documentation-vision?node-id=1528-43714&m=dev)), in the four methods the design covers: `get`, `post`, `delete` and `put`. Each method has a labelled and an icon-only mode.

There is existing Badge and Label components but these are different, the fact there is new color tokens imply that we aren't using the design system components, these will be unique to docs.

We will however use the design system tokens colours (until such a day when they create the actual tokens)

## What is here

- `src/components/ApiBadge.astro` — takes `method` (`get` / `post` / `delete` / `put`) and `hasLabel` (default `true`).
- `src/assets/icons/` — `arrow-down-left.svg`, `arrow-up-right.svg`, `xmark.svg` and `pen-line.svg`, exported from Figma.
- `src/styles/vars.css` — the `--colorApiBadge*` tokens for both themes.
- `src/pages/components.mdx` — an ApiBadge section showing both modes, on `/components`.

# Results
Visible here: https://stoctodocspr3355.z22.web.core.windows.net/components#apibadge

Dark Mode:
image

Light Mode:
image


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3355.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3355&status=all).

src/styles/vars.css
```diff
@@ -232,6 +232,27 @@
   --colorNavBackgroundRest: rgb(from var(--colorScalesBlue500) r g b / 0);
   --colorNavBackgroundHover: var(--colorScalesSlate100);
   --colorNavBackgroundActive: rgb(from var(--colorScalesBlue500) r g b / 0.06);
+
+  /* TODO: drop these once color.apiBadge.* are available in design-system-tokens */
```
borland: These apiBadge tokens are specific snowflakes for the docs site, they're never going to go into the shared design system. As such we should remove the TODO and maybe replace it with a comment explaining the above.

Same deal with colorNavBackgroundRest above, it shouldn't have a TODO

src/styles/vars.css
```diff
@@ -276,4 +297,25 @@ html[data-theme='dark'] {
 
   /* TODO: drop once color.nav.background.* are available in design-system-tokens */
   --colorNavBackgroundHover: var(--colorScalesNavy800);
+
+  /* TODO: drop these once color.apiBadge.* are available in design-system-tokens */
```
borland: As above no todo

src/components/ApiBadge.astro
```diff
@@ -0,0 +1,125 @@
+---
+import type { HTMLAttributes } from 'astro/types';
```
borland: Having this be an astro component is annoying, because it means every docs page must be a `.mdx` instead of a `.md`, and the docs-generator must output `<ApiBadge >` into the markdown, which will render as broken when people try and view/preview the markdown locally or in the github diff.

I would strongly prefer that everything in the ApiDocs remains simple HTML with CSS styling, because Octopus Server devs are going to see it in their PR's over in the OctopusDeploy repo, where there is no astro.

Can we do this as something like `<span class="api-get">GET</span>` and have CSS do all the styling without needing components? It seems like that should be possible? Doesn't need to have perfect fidelity with the mockups

src/styles/api.css
```diff
@@ -0,0 +1,82 @@
+:is(.api-get, .api-post, .api-delete, .api-put) {
```
enf0rc3, line 1: I suspect that the API layout will have more "non-astro" components that will need styling, so I opted to have a new file separate from main.

src/components/ApiBadge.astro
enf0rc3: Good call, it will likely follow something similar with how we do the CodeBlock enhancements, so won't have astro button support.

I did suspect that we would build the nav-menu a different way that would support the AstroComponents, though, but why have two different components, and we can come to that when we need it!

src/styles/api.css
borland: Nice. I was thinking it'd be cool if we could have a very-mini "api docs" site within the OctopusDeploy repo to optionally help developers see what the docs MD changes will look like when rendered. Probably we won't do it, but keeping API css separate would make it easier if we did

borland (approved):
Looking good!

PR #3349, opened 2026-08-13
Return to the page you were on after signing in

rosslovas:
octopus.com/signin will take you to the billing page by default. With this change users will be returned back to wherever they just were in the docs after signing in.

PR #3362, opened 2026-08-14
Return to /docs after signing out

rosslovas:
Ideally we'd return to whatever page we were actually on like when signing in (#3349) but the signout page has a server side restriction, controlled by a third party (Auth0), that does not support wildcards.

image

As such, the best we can realistically do is go back to octopus.com/docs.

### DIFF
--- a/dictionary-octopus.txt
+++ b/dictionary-octopus.txt
@@ -548,6 +548,8 @@ struct
 stylesheet
 subcomponent
 subcontext
+subpath
+subpaths
 sunsetting
 swaggerui
 systemprofile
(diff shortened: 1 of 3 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3362.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3362&status=all).

enf0rc3 (approved):
Fair enough!

PR #3364, opened 2026-08-14
Stabalize scrollbar gutter so the header stops shifting when search opens

enf0rc3:
Focusing the search hides the body overflow to lock scrolling. That takes the scrollbar with it and widens the viewport, so both fixed headers grew and their contents jumped 15px to the right.

`scrollbar-gutter: stable` on `html` holds that width open, so the viewport stays a constant width whenever anything locks scrolling.

This was "worse" on the new navbar and resolutions that are "tight" on space.

## Before:

https://github.com/user-attachments/assets/46a2ee08-ae7e-4e80-ab82-d3ec6587323b

## After:

https://github.com/user-attachments/assets/c93c2e69-2c12-4e06-ad03-f6d8b5ffb794

### DIFF
--- a/src/scripts/search.js
+++ b/src/scripts/search.js
@@ -138,35 +138,17 @@ function initializeSearch() {
   // Dropdown accessibility controls
   document.addEventListener('keydown', handleDropdownKeyboardNavigation);
 
-  // Locking the page hides the body's overflow, which takes the scrollbar with
-  // it and widens the viewport. The header is fixed, so it is sized to the
-  // viewport and its contents would jump outward. Record the width the
-  // scrollbar was occupying so the CSS can hold that space open.
-  function lockScroll() {
-    const gutter = window.innerWidth - document.documentElement.clientWidth;
-    document.documentElement.style.setProperty(
-      '--scroll-lock-gutter',
-      gutter + 'px'
-    );
-    document.body.classList.add('search-scroll-lock');
-  }
-
-  function unlockScroll() {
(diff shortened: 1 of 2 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3364.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3364&status=all).

borland:
Do we need this?

- I can't repro the problem your video shows, on octopus.com/docs or anywhere else.
  - Observing it in the video, it's a really minor bug that people are unlikely to notice
- The new nav and upcoming search work is likely to change this stuff anyway, these additional changes might just get in the way

enf0rc3:
This bug actually showed itself during testing of new-nav, as it looked worse with the new design, but @rosslovas  suggested:

> I think we're gonna revamp the actual search functionality if I'm not mistaken so we can check for this issue with whatever we replace it with, and fix if necessary

And i figured:

> Ill ask claude to do a pr, if its easy fix we can merge
--- 

I agree with you, it is a minor impact.

I do not think the change will "get in the way" given its fix was trivial css changes that would be easy to revert anyway.

# Proof of new nav fix: left (changes, right original)

https://github.com/user-attachments/assets/6813745d-4747-4994-bf79-c3355a1e4115

borland (approved):

PR #3365, opened 2026-08-17
Update the favicon to match the product favicon

enf0rc3:
The docs favicon was a dark document tile with an octopus on it. The product at `deploy.octopus.app` uses the plain blue octopus mark, so the docs now use the same mark.

The icon set is regenerated from `public/docs/img/octopus.svg`, which is the same artwork the product favicon is rendered from — same `#0d80d8` fill, same 30:32 aspect.

The set is also trimmed from six files to three:

| File | Role |
| --- | --- |
| `icon.svg` | Scalable mark, used by Chrome, Firefox and Edge |
| `favicon-32x32.png` | Raster fallback, used by Safari before v26 |
| `apple-touch-icon.png` | 180x180 opaque, used by the iOS home screen |

The `android-chrome-*` PNGs did nothing without a web manifest, and there is no manifest in the repo. The `favicon.ico` is gone because the only thing an ICO buys is the hardcoded `/favicon.ico` fallback, and that always resolves against the domain root, so a file under `/docs/icons/` could never serve it. `octopus.com` itself ships a PNG favicon and no ICO.

`apple-touch-icon.png` was 96x96 with an alpha channel. It is now 180x180 on an opaque white background, since iOS upscales anything smaller and composites transparent home screen icons onto black.

Verified against the dev server: all three files serve with the right MIME types and the head links resolve.

## Before:

image

## After:
image

Closes DPT-3124


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3365.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3365&status=all).

borland (commented):
You mentioned that nothing references the android-chrome or other PNG icon files.
Rather than re-generating them, we should delete them.

Re: the favicon.ico, I notice that octopus.com doesn't use a favicon.ico file at all, but rather does this:

```html
<link rel="icon" type="image/png" href="/octopus-public/images/(global)/favicon.png" />
<link rel="apple-touch-icon" href="/octopus-public/images/(global)/apple-touch-icon.png" />
```

Is there a reason why we shouldn't just delete all our favicons, then copy those two lines and link to the same icon file as the main site?

borland:
Additional favicon context from claude:

> When a browser loads a page, it looks for `<link rel="icon">` tags in the `<head>`. If it finds one it can handle, it uses that. If it finds none, it falls back to blindly requesting /favicon.ico from the site root — that fallback is hardcoded behavior dating back to IE5, and it's why the file lives at the root by convention rather than because anything requires it.

> The catch is that plenty of software only does the hardcoded fallback and never parses your HTML: RSS readers, link-preview generators, some crawlers, older embedded browsers, and various "get this site's icon" APIs. Those will hit `/favicon.ico`, get a 404, and show a blank square.

It's important though to remember when we consider that context, the fallback is for the **root site**, so it's always going to go for `octopus.com/favicon.ico`. There's no reason for us to ever have a `favicon.ico` in the docs site, unless we started hosting it on a different domain.

borland (approved):
After latest changes, favicon looks great.

PR #3366, opened 2026-08-17
Consolidate Card component

ellen-octopus:
## Card component
- Consolidated `Card` into a single component:
   - Added mutually-exclusive `icon` / `imgSrc` variants
   - Added `featured` variant that can be used on homepage to emphasise content (has larger title and description).
   - Dropped the unused `padded` and `related-topics` variants.
- Removed `IconTile` component → now fully covered by `Card`.
- Removed unused `Related` component. 
- Updated `ArticleList` to use card styling → this also isn't currently in use but I left it in place for now.

> [!IMPORTANT]
> I relied on Claude to fully remove the `Related` and `IconTile` components, and to update `ArticleList` to use card styling. 

### Default size

| No image or icon | Image | Icon |
| --- | --- | --- |
| image | image | CleanShot 2026-08-17 at 12 44 58@2x |

### Featured size

| No image or icon | Image | Icon |
| --- | --- | --- |
| CleanShot 2026-08-17 at 12 57 07@2x | image | CleanShot 2026-08-17 at 12 58 21@2x |

### Patterns
[deployments/patterns](https://stoctodocspr3366.z22.web.core.windows.net/docs/deployments/patterns/) updated to use card with no icon or image.

| Before | After |
| --- | --- |
|  CleanShot 2026-08-17 at 10 27 02@2x | CleanShot 2026-08-17 at 10 26 22@2x |

### Variables
[projects/variables/getting-started](https://stoctodocspr3366.z22.web.core.windows.net/docs/projects/variables/getting-started#learn-more) replaced card with normal list.

| Before | After |
| --- | --- |
| CleanShot 2026-08-17 at 10 29 31@2x | CleanShot 2026-08-17 at 10 29 58@2x |

### Kubernetes
[kubernetes](https://stoctodocspr3366.z22.web.core.windows.net/docs/kubernetes#getting-started) updated to icon variant.

| Before | After |
| --- | --- |
| CleanShot 2026-08-17 at 10 21 24@2x | image |

### Components showcase
[components](https://stoctodocspr3366.z22.web.core.windows.net/components#cards) updated with new card showcase.

### Icons
- Added Docker, Kubernetes, Helm, Kustomize, Windows, and YAML icons


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3366.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3366&status=all).

src/pages/docs/projects/variables/getting-started.mdx
```diff
@@ -25,19 +25,24 @@ In this example, we'll add a variable to a Hello World project that runs a scrip
...
-![Adding a project variable in Octopus Deploy](/docs/img/projects/variables/images/add-project-variable.png)
-:::
+
+   :::figure
+   ![Adding a project variable in Octopus Deploy](/docs/img/projects/variables/images/add-project-variable.png)
+   :::
+
 7. Save the variable by clicking **Save**.
-8. In this example, we’ll reference this variable from a **Run a Script** step.
+8. In this example, we'll reference this variable from a **Run a Script** step.
```
enf0rc3, line 34: All these will likely conflict with the "mass format" here: https://github.com/OctopusDeploy/docs/pull/3358 

should avoid unrelated docs changes like this in the future.

I would suggest waiting and then specifically calling out this pr what it does and rebasing, did you actually make any docs changes? apart from automatic ones?

ellen-octopus: Thanks @enf0rc3
On this page I had only wanted to update the Learn More section to remove the cards it was using, the rest of the changes were automatic ones

src/components/icons/Helm.astro
borland, line 1: This looks like it's just a copy of helm.svg but in a `.astro` file.
Can we just commit the SVG itself?

src/pages/docs/kubernetes/index.mdx
```diff
@@ -31,36 +35,32 @@ Octopus Deploy makes it easy to manage your Kubernetes resources, whether you're
...
 With Octopus, you can choose between managing your application configuration with Helm charts or YAML.
 
-::::div{.docs-home .simple-grid}
+::::div{.docs-home .col-2}
 
 <Card
   description="Learn more about how you can use Octopus variables for Helm values in our Helm documentation."
-  imgAlt="Helm logo"
-  imgSrc="/docs/img/helm-logo.png"
+  icon={Helm}
```
borland: One thing which bit me about this kind of thing. Referencing an image directly as a component causes it to be emitted directly into the HTML output.

It's only referenced one time in the document, so it isn't a big deal, but if the image were referenced, say, 10 times, we'd end up with 10 duplicates of the image which makes the HTML larger and correspondingly slower for the end user (and terrible for LLM's if they happened to consume the HTML variant of the page). I did this by accident with the chevrons in the nav-list. There were dozens of them 🤦 

Except for niches I think we'd be better off to reference images via svg/png _files_ rather than inlining the HTML

src/components/icons/Helm.astro
ellen-octopus: I need the `.astro` file in order to specify the icon colour and make sure it respsects the theme.

Claude's technical explanation:
> Each path has `fill="currentColor"`, which lets the icon inherit its color from CSS. That only works if the SVG is inlined in the DOM, which is what the `.astro` wrapper gives us — loading it as a raw `` would break that, since `currentColor` doesn't work through an image reference.

borland: Hmm. We do color-theming for images that are part of the interface, like the chevrons on the nav menu, or the toolbar images... 

But this is content within an individual page. We have about 2000 png files and we don't have light/dark mode variants of them, wouldn't things like the helm logo fall in that bucket?

src/pages/docs/kubernetes/index.mdx
ellen-octopus: I got Claude to fix this (pushed in the latest commit) so `icon` now takes the URL of an SVG file instead of a component. I've removed the `.astro` icon files I introduced 🙂 

The card component still supports themed icons (this is a deliberate design decision) but I've added a `branded` variant that preserves the colours in the source SVG file. This can be used for logos:
CleanShot 2026-08-18 at 12 00 01@2x
CleanShot 2026-08-18 at 12 00 24@2x

borland (approved):

PR #3368, opened 2026-08-17
Consolidate structured data into JSON-LD

rosslovas:
# Summary

JSON-LD is Google's recommended format for structured data. Source: https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data

> In general, Google recommends using JSON-LD for structured data if your site's setup allows it, as it's the easiest solution for website owners to implement and maintain at scale (in other words, less prone to user errors).

Prior to this PR, we had a mix of all three types: microdata, RFDa, and JSON-LD. Microdata and RFDa were to some degree broken / contradicting the JSON-LD in places. This PR consolidates it all on JSON-LD, and makes an effort to improve it without introducing any regressions.

- Everything now emits `TechArticle` instead of `Article` as the article type (previously only `/getting-started` and `/installation` had this type) as I think it's more accurate given the nature of our docs. (This can be opted out of if there are pages that warrant it)
- JSON-LD is now the source of truth for breadcrumbs rather than RFDa. They use absolute URLs (seems to be the correct approach from what I can tell, Google's own examples also follow this). The breadcrumb trail is shared between the rendered breadcrumbs and the metadata so they can't drift.
- I verified across all rendered pages there's no difference to the breadcrumbs, and to the best of my ability to tell all JSON-LD data is valid per schema.org.

There are a few pages no longer emitting breadcrumb data but three have `noindex` so they can't show up in search results anyway and thus are irrelevant, one is octopus.com/docs/guides which is a redirect to `/docs`, and the last is `/docs` itself which only has one crumb, itself, which doesn't mean anything.

### DIFF
--- a/src/layouts/Default.astro
+++ b/src/layouts/Default.astro
@@ -3,7 +3,7 @@ import { accelerator } from '@lib/accelerator';
 import { PostFiltering } from 'astro-accelerator-utils';
 import type { Frontmatter as OriginalFrontmatter } from 'astro-accelerator-utils/types/Frontmatter';
 import { SITE } from '@config';
-import { getImageInfo } from '@util/custom-markdown.mjs';
+import { buildCrumbs, type Crumb } from '@util/breadcrumbs';
 
 // Theme components
 import Head from '@components/HtmlHead.astro';
@@ -30,18 +30,16 @@ import DocsSearch from '../components/DocsSearch.astro';
 type Props = {
   frontmatter: OriginalFrontmatter;
   headings: { depth: number; slug: string; text: string }[];
-  breadcrumbs?: { url: string; title: string; ariaCurrent?: string }[] | null;
+  breadcrumbs?: Crumb[] | null;
 };
 const { frontmatter, headings, breadcrumbs } = Astro.props satisfies Props;
+
+const crumbs = buildCrumbs(Astro.url, breadcrumbs);
 
 const lang = frontmatter.lang ?? SITE.default.lang;
 const textDirection = frontmatter.dir ?? SITE.default.dir;
 
 // Logic
-const metaImage = frontmatter.bannerImage
-  ? getImageInfo(frontmatter.bannerImage.src, '', SITE.images.contentSize)
-  : null;
-
 const title = await accelerator.markdown.getInlineHtmlFrom(
   frontmatter.title ?? SITE.title
 );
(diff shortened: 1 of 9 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3368.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3368&status=all).

src/themes/octopus/utilities/structured-data.ts
```diff
@@ -292,6 +292,39 @@ export function buildArticle({
...
+  canonicalURL: string
+): BreadcrumbListNode | null {
+  const crumbsWithTitle = crumbs.filter(
+    (c) => typeof c.title === 'string' && c.title
+  );
+  if (crumbsWithTitle.length < 2) return null;
+
+  const itemListElement = crumbsWithTitle.map((crumb, index) => {
+    const isLast = index === crumbsWithTitle.length - 1;
+    const item = isLast ? undefined : toAbsoluteUrl(crumb.url);
```
rosslovas, line 306: I think it's pretty intuitive that the last breadcrumb is just the current page so no link is necessary but just to substantiate:

https://developers.google.com/search/docs/appearance/structured-data/breadcrumb

> If the breadcrumb is the last item in the breadcrumb trail, item is not required. If item isn't included for the last item, Google uses the URL of the containing page.

src/layouts/Default.astro
```diff
@@ -92,20 +95,11 @@ const lastUpdated = frontmatter.modDate ?? frontmatter.pubDate ?? null;
               lang={lang}
             />
           </div>
-          <div class="page-content anim-show-parent" itemprop="articleBody">
+          <div class="page-content anim-show-parent">
             <slot />
             <Authors frontmatter={frontmatter} lang={lang} />
             <Taxonomy frontmatter={frontmatter} lang={lang} />
             <MarkdownLinks lang={lang} />
-            {metaImage && <meta itemprop="image" content={metaImage.src} />}
```
rosslovas, line 101: The only page that had a banner image (which is what this is, `metaImage` is set to `frontmatter.bannerImage` above) was `/docs/search`, which points to https://octopus.com/devops/img/search.png

It's `noindex` so it doesn't show up in search engines, and is not relevant in that regard.

It still has that same image provided as an Open Graph image (`og:image`), which is used for link previews in other places e.g. Slack.

borland (approved):

PR #3369, opened 2026-08-17
Component: DocsSearch overlay

enf0rc3:
Replaces the inline results dropdown with a modal search overlay built to the [Figma design](https://www.figma.com/design/DTvlouW1QArI3ZTf1PXs4J/Documentation-vision?node-id=880-19909&m=dev): tabbed by section with live counts, arrow-key navigation, and a result row per page carrying its type icon, excerpt and breadcrumb trail.

Both navs point their fields at one dialog, so there is a single result list rather than the two that shared `search.js` before. Try it with `?newnav=1` and `?newnav=0`.

Results are unchanged. The existing `search.json` sits behind a `SearchEngine` interface, which is the seam the Pagefind and Orama spikes swap out next. The index is now fetched when the overlay opens rather than on page load, taking 1.86MB off every page view.

The overlay is a native `<dialog>`, so the focus trap, scroll lock and backdrop come from the browser. Focus stays in the input and `aria-activedescendant` moves down the list, which the old dropdown had none of.

`search.js`, `search-dialog.js` and the orphaned theme `Search.astro` are deleted — the adapter supersedes them.

There is a `DocsSearch` section on `/components` with a fixture-backed instance, and `tests/docs-search.spec.ts` covers it.


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3369.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3369&status=all).

borland (approved):
Skimmed the JS/HTML -- there's a LOT going on, and we could probably simplify some of it, but I don't think we really need to. I did some functional testing on the demo site and it's looking nice. Let's merge, we can iterate if we need to

PR #3371, opened 2026-08-17
Improve site search: replace the search index with Pagefind

enf0rc3:
# Background

The current search has always been "lackluster" and often would give you completely irrelevent search results.

We want to improve this by using a library that specialises in searching on static sites and gives us the ability to tweak it. 

This is a type of task that can have a lot of iteration, so i think getting something out that is better than the current should be the goal. 

# Results:
Site now uses Pagefind. It indexes the HTML the build emits, and does some smarts and ranking to better rank the content we want, we did tweak it but found the defaults and headings were suffice. 

## AI Generated Details of the pr

- **Body text search.** 1,251 pages indexed from their article content, so a phrase buried in the middle of a page finds it.
- **BM25 ranking.** Pagefind scores with BM25, so term frequency, document length and term rarity all count. The overlay adds a reorder on top: a page the query names by title or slug wins outright, and guides outrank the API and CLI reference unless the query asks for one.
- **Headings as results.** Up to two matched headings under each of the leading results, each one arrow-selectable and landing on its anchor.
- **Section tabs** with live counts, so a query shows what Docs, API and CLI each hold.
- **The whole result set.** The list extends as it is scrolled. A query matching 1,057 pages offers all of them.
- **Technical punctuation.** `<head>`, `.nupkg` and `#{Octopus.Environment.Name}` are findable as written.
- **Honest empty states.** A query on nearly every page asks the reader to narrow it; a query with no answer says so.
- **Search analytics.** The `searched` event `Plausible.astro` has been waiting for now fires, so what readers type is finally measurable.
- **A much smaller payload.** The old index shipped 423 KB brotli to every visitor before the first query could run. Pagefind ships 114 KB of runtime and fetches index chunks per query.

## How this was measured

Relevance was tested by replaying the site's own search log — the terms readers actually typed, weighted by how many typed each one — plus a second set derived from page analytics covering 12,558 visitors. The harness types into the real overlay in a browser, so what it measures is what a reader gets.

Against production, the top five answers **84% of real searches, where the current engine answers 81%**. Latency is unchanged.

The harness is `tools/search-bakeoff`, kept so the comparison can be re-run.


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3371.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3371&status=all).

enf0rc3:
## Verified on the ephemeral environment

`https://stoctodocspr3371.z22.web.core.windows.net/docs` — this closes the one open risk from the plan.

**Every Pagefind file type serves.** `pagefind.js` as `application/javascript`, `pagefind-entry.json` as `application/json`, the wasm as `application/x-gzip` (Azure content-sniffs it, which is harmless because Pagefind reads it as bytes rather than streaming-compiling). The index chunks and fragments load too. Nothing was skipped by the upload step and no MIME allowlist was needed.

**Search is correct on real hosting.** `tentacle` → 248 results, All 248 / Docs 183 / API 14 / CLI 51 — identical to local. Deployed `pagefind-entry.json` reports `page_count: 1251`, matching the build. Cold first search **2.31s** over the internet.

**Result links stay on the host.** Zero absolute hrefs; clicking a result landed on `stoctodocspr3371.../docs/octopus-rest-api/tentacle.exe-command-line/`.

## One thing to know about the hosting

Azure Blob static website does **no** compression — `Accept-Encoding: gzip` returns everything with no `Content-Encoding`. It barely matters here, because Pagefind's `.pf_*` chunks are already compressed binary and `pagefind.js` is 45 KB. It matters a great deal for #3372, which pays 11.4 MB uncompressed rather than the 2.4 MB I estimated. Details in that PR.

## Bake-off result

| | Legacy | **Pagefind** | Orama |
| --- | --- | --- | --- |
| First search, over the wire | 1,860 KB | **278 KB** | 11,989 KB |
| Each later search | 0 KB | 40–160 KB | 0 KB |
| Cold on staging | — | 2.31s | 2.16s |
| At 1.6 Mbps (arithmetic) | — | ~1.4s | ~60s |
| Typo tolerance | no | no | no (did not work) |
| Hosting risk | — | **disproven** | none |

Both feel the same on a fast connection. The difference is what happens below that, and payload was the defect this work set out to fix.

Recommend merging this and closing #3372.

src/layouts/Default.astro
```diff
@@ -59,10 +60,14 @@ const searchUrl = search && accelerator.urlFormatter.formatAddress(search.url);
 const isSearchPage =
   accelerator.urlFormatter.formatAddress(Astro.url.pathname) === searchUrl;
 
+// Gates the header's search field only. The overlay below is always rendered:
+// /docs/search has its own field and opens that same overlay.
 const showSearch = !isSearchPage;
```
enf0rc3, line 65: This can be removed if / when we remove the search page (no longer needed imo)

src/scripts/docs-search.ts
```diff
@@ -99,8 +132,86 @@ function setup(dialog: HTMLDialogElement) {
     return item;
   }
 
+  function drawSection(
```
enf0rc3, line 141: Essentially to allow scrolling to pull more results

rosslovas:
Haven't done a review yet but just started to play on the staging site, and I apparently found 1 API result yet no results found, what might be going on here?

image

borland (approved):
In terms of code, it all looks like claude and it all seems fine from the skim-review I gave it.

In terms of experience, I had a play on the staging site, and the search results are noticeably better. Nice stuff.

Performance-wise, the fragmented indexing definitely resulted in less data being transferred, however it seems like the index is hyper-fragmented into heaps of tiny files. 

I did a test using the MS Edge dev tools to set the network speed to "Slow 4G", and pagefind was noticeably slower:

I hard-refreshed the front docs page, clicked on the search box, and typed "kubernetes"

- The old search on octopus.com loads the 1.8 MB search.json (compressed to 415kb) in 2.9 seconds. This was enough that I noticed the delay; I finished typing the word and still had to wait for the results to appear, but it wasn't too bad.

- The new search in the staging site only loads 95kb of indexes, and they're already compressed, so that's a huge win. However it took over 5 seconds for results to appear, feeling noticeably slower. It loaded **32 fragments**, each of which was about 1-2 kilobytes, and it seemed like only about 6 would download at any one time; this concurrency-limiting behaviour seemed to be the cause of the slowness.

Claude tells me that the pf_fragment files aren't the search results, but rather the preview fragment data for each page that the search finds, so we can't chunk or merge them, but we can reduce the number we download by tweaking the number of visible results rows we fetch. The UI only shows **Five** results on screen. We want to over-fetch a little bit to allow for scrolling, but maybe 10 would be a better number than 32.

In saying that, maybe it was just down to the staging site lacking Azure Front Door, or maybe it was an artifact of the browser network throttling tooling. I'm going to take an educated guess that it's probably fine and say let's merge it, but we might need to think about tuning how many results we fetch or something.

enf0rc3:
Thanks for the review:

Breaking down the timings, I put a majority of the delay with the server "waiting for server response" so it will improve on production. We can definitely fine-tune this more. 

## Prod
image

 
## Staging
image
image


I will do a quick look to see if there are any quick wins or changes we can't make to improve this.

enf0rc3:
@borland Given this PR is already so big, i will do a follow up PR for further tweaking to the search algo, I have implemented a fix for the issue @rosslovas raised above, so that will come with this merge.

PR #3372, opened 2026-08-17
Spike: Orama

enf0rc3:
Spike 2 of 2 for the search engine bake-off. Branched off #3369, so that PR's commit is in this diff. Pagefind is #3371.

**Closing unmerged. Pagefind won the bake-off — the comparison is at the bottom of this description.** The branch stays for reference.

## What it does

Replaces the hand-rolled `search.json` engine with Orama, behind the `SearchEngine` seam from #3369. **Body text is searchable for the first time.**

The corpus is the markdown `llm-md-emitter` already writes into `dist/docs/`, so there is no second extraction pipeline to keep in step — the emitter runs the same eligibility predicate search uses, so redirect stubs and `navSearch: false` pages are gone before this sees them.

The whole index is downloaded and restored into memory once, in a worker, and every query after that costs nothing. Restore is CPU-bound and would freeze the page while the overlay is open, which is why it is off the main thread.

Deletes 645 lines of custom search code: a Porter stemmer, a synonym map, the string helpers and the legacy engine.

## Results

Measured through the search overlay against expected URLs written before each run. Two query sets, both from analytics and both weighted by traffic.

| Query set | Engine | Top 5 | Rank 1 | MRR |
| --- | --- | --- | --- | --- |
| Real searches, 57 terms | Orama | 80% | 52% | 0.635 |
| | Legacy | 56% | 38% | 0.471 |
| Top pages, 98 terms | Orama | 90% | 79% | 0.842 |
| | Legacy | 89% | 57% | 0.716 |

Orama's real strength is the shape of the queries it handles, which is where an as-you-type overlay lives:

| | Orama | Pagefind | Legacy |
| --- | --- | --- | --- |
| Typo queries | **67%** | 42% | 42% |
| Top 5 on a partial query | **72%** | 57% | 49% |
| Head terms | 94% | **97%** | 81% |
| Natural language | **73%** | **73%** | 67% |

| | Legacy | Orama |
| --- | --- | --- |
| Index on the wire (brotli) | **0.35 MB** | 0.66 MB |
| Parsed per page load | **1.74 MB** | 5.33 MB |
| Cold first search, adjusted | **699 ms** | 3950 ms |
| Warm search | 486 ms | **326 ms** |
| Return visit | 482 ms | **288 ms** |
| Keystroke p95 | 307 ms | **324 ms** |

Once the index is in memory Orama is the fastest thing measured. Getting it there is the problem: it needs about five seconds of reading time before the first search is ready.

## Configuration that matters

Four settings that are easy to get wrong and were, at first:

- **The worker must create its database with the same tokenizer the index was built with.** `load` restores the data but the tokenizer comes from `create`, so building with the English stemmer and restoring without it compares unstemmed query terms against stemmed index terms. Searching `variables` matched 5 pages instead of 321 and no typo matched at all, while search kept returning results throughout.
- **Stop words are not on by default.** Orama's tokenizer defaults `stopWords` to an empty list, so `how`, `do` and 178 others were live search terms. The list ships separately as `@orama/stopwords`.
- **`sort: { enabled: false }`** — nothing here sorts, and the sort store was 2.69 MB of the index.
- **The stored body is trimmed to 200 characters** after `save()`. The inverted index keeps every term; this is only the copy the result excerpt is cut from.

Together the last two took the index from 11.43 MB to 5.33 MB with rankings unchanged. Ranking was then tuned separately: `boost: { title: 8, ... }`, and the same landing-page reorder Pagefind needed, because a page ranking below its own children was a defect in every engine tested.

## Why Pagefind was chosen

Relevance did not decide it. On the real search log the two are level — 81% against 80% in the top five, and identical at rank one. On the terms that carry the traffic Pagefind is clearly ahead, 98% against 90%.

Three things decided it:

1. **Payload.** 0.66 MB on the wire and 5.33 MB parsed, against Pagefind's 0.08 MB and 0.11 MB. Every visitor pays that on every page load whether they open search or not.
2. **Cold start.** 3950 ms against 1821 ms, and Orama needs roughly five seconds of reading time before its first search is ready.
3. **The 8 MB compression cliff.** Front Door only compresses responses between 1 KB and 8 MB. At 5.33 MB there is headroom, but the corpus grows, and crossing it would stop compression silently and return the cold start to about 5.7 s. That is a latent failure tied to a number nobody watches.

What was given up is real and worth recording: typo tolerance drops from 67% to 42%, and top-five on a partial query from 72% to 57%. An as-you-type overlay spends most of its life on partial queries, so this is the cost of the decision rather than a rounding error.

## Notes

- **8 `.mdx` pages are missing from the index.** All of them import `.astro` components, which the markdown emitter cannot resolve. Not fixed here.
- `@orama/plugin-data-persistence` cannot be bundled for a browser worker — it reaches for Node's filesystem and buffers, and the worker dies on import with `Class extends value undefined`. Core `save`/`load` do the same job and are what this uses.
- Facet counts come from Orama's own `facets` rather than being counted from returned hits, which capped them at the worker's result limit.
- Orama returns whole documents rather than excerpts, so the result row cuts its own window around the first matched term.
- `Plausible.astro` still listens for a `searched` event that nothing fires.


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3372.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3372&status=all).

enf0rc3:
## Correction and staging results

**The 2.4 MB gzipped figure in the description is not achievable on this hosting.** Azure Blob static website hosting does no on-the-fly compression — requesting `Accept-Encoding: gzip` returns the full 11,989,438 bytes with no `Content-Encoding` header. Verified on this PR's own ephemeral site:

```
curl -H "Accept-Encoding: gzip" .../docs/search-index.json
200  wire=11989438b  type=[application/json]   # no Content-Encoding
```

Nothing on that host is compressed, Pagefind's assets included. So the real wire cost for Orama is **11.4 MB, not 2.4 MB**. Getting the compressed figure would need the `microsite-deployment` pipeline to pre-compress and set `Content-Encoding` at upload. Production sits behind Front Door which can compress, but 11.4 MB is above its documented 8 MB ceiling, so it probably would not help there either — worth confirming before relying on it.

## It works, and on a fast connection it feels fine

`https://stoctodocspr3372.z22.web.core.windows.net/docs`

`tentacle` → 200 results, All 200 / Docs 151 / API 7 / CLI 42. Cold search **2.16s**, which is actually a shade faster than Pagefind's 2.31s on the same connection. Results are relative and stay on staging.

That is the honest picture at ~47 Mbps: the download is 2 seconds and you would not notice. The gap is entirely about what happens below that.

| Connection | Orama 11.4 MB | Pagefind 278 KB |
| --- | --- | --- |
| 47 Mbps (measured) | 2.0s | ~0.05s |
| 10 Mbps (4G) | 9.6s | 0.2s |
| 1.6 Mbps (Fast 3G) | **60s** | 1.4s |

Download time only; arithmetic from the measured wire sizes, not measured directly.

## Recommendation

Close this in favour of #3371. Payload was always the deciding axis and this makes it worse, not better — every visitor who searches pays 11.4 MB uncompressed, against 1.86 MB today. Typo tolerance, the one thing that could have justified the cost, returns nothing for `kuberntes`.

Worth keeping from this spike if Pagefind is ever revisited: reusing `llm-md-emitter`'s output as the corpus worked cleanly and needed no second extraction pipeline, and Orama ranked `target tag` and `RBAC` better than Pagefind does.

enf0rc3:
Closing unmerged. We are going with Pagefind (#3371).

Relevance did not decide it. On the real search log the two engines are level — 81% against 80% in the top five, and identical at rank one. On the terms that carry the traffic Pagefind is ahead, 98% against 90%.

Payload and cold start decided it. Orama ships 0.66 MB on the wire and 5.33 MB parsed against Pagefind's 0.08 MB and 0.11 MB, and every visitor pays that on every page load whether they open search or not. First search is 3950 ms against 1821 ms, and Orama needs about five seconds of reading time before it is ready. Behind both sits the 8 MB compression cliff: Front Door only compresses between 1 KB and 8 MB, so a corpus that grows past it would silently lose compression and take the cold start back to about 5.7 s.

What we are giving up is real. Typo tolerance drops from 67% to 42%, and top-five on a partial query from 72% to 57%. An as-you-type overlay spends most of its life on partial queries, so that is the price of the decision rather than a rounding error. Worth revisiting if search analytics later show typo and partial queries are a bigger share of traffic than the log suggests.

The branch stays for reference. The updated description above has the full numbers and the configuration notes — the tokenizer mismatch in particular is the kind of thing that fails silently and would cost someone a day.

enf0rc3:
Archived, so this survives the branch being deleted.

- `search-spike-orama` — tag on `438570fc5`, this branch as measured. The only tuned Orama integration we have: matched tokenizer, stop words, sort store dropped, body trimmed, ranking swept.
- `search-bakeoff-2026-08` — tag on the measurement harness, with both traffic-weighted query sets, every tuning tool and the decision doc. Also on branch `worktree-search-bakeoff-harness`.

To re-run the comparison later, start at `tools/search-bakeoff/README.md`, section **Picking this up later**. The staging environments both PRs were measured on are gone, so it covers pointing the harness at local builds instead.

Worth re-running when there is real search analytics. Pagefind was chosen knowing it is worse at typos (42% against 67%) and partial queries (57% against 72%), and nobody had data on how much traffic those account for. `Plausible.astro` listens for a `searched` event that still nothing fires; once it does, that assumption is checkable.

PR #3376, opened 2026-08-17
Include programatically generated API docs - initial first step

borland:
This is the first part of a staged introduction of our API docs into the documentation site.

It adds
- A new folder `/docs/api` with all the auto-generated API docs, copy pasted from the Octopus Server repo
- Supporting infrastructure
  - The API pages have a different layout to the main one
  - They have their own navigation tree as well
  - Includes CSS
  - Example requests and responses need special handling, so there's a `satteri-api-examples` plugin which processes them
- The docs/api folder has been excluded from Spellcheck and Markdown lint as it doesn't comply with either. To achieve compliance, we would need to edit as many as 1,800 files in the Octopus Server repo, which is intractable. We'll chip away at this over time from here.

## How to review this PR

- Ignore all the stuff under `pages/docs/api`. It is programatically generated based on the Octopus Server codebase, and it is equivalent to our swagger content on https://demo.octopus.app/swaggerui/index.html. There are many flaws in the documentation content, however it is already public material and we haven't changed that, so it is not the responsibility of this PR to fix it.

- Claude wrote the rest of the code. It may not be ideal, technically, but it is very limited in what it needs to do, and the blast radius is contained (a bug will only impact a small section of the site and should not affect other things) so it should be good enough

Treat this as a stepping stone. Once it merges I will proceed with the next steps:

## Things this PR does not include

- We need a landing page for `/docs/api` which can help people find common things, such as introductory material and explaining that this documentation is for **octopus cloud** and that LTS versions might be different.
- We need to move the existing content from `/docs/octopus-rest-api` into `/docs/api` and mix it in with the generated content. The generated content may well move to a subfolder such as `/docs/api/reference`. Will cross that bridge when we get to it.
- We need some non-generated pages to explain things like authentication, if the existing content is insufficient
- We need a process to automatically apply changes from the Octopus Server repo. That side hasn't merged yet, and in the interim I'll just apply changes by hand.

**When this PR merges, the new files under `/docs/api` will be live, but _nothing will link to them_**. We will start linking to them and "light them up" as part of that other work.

## The API docs format

This is best explained by taking one of the files as an example

````markdown
---
layout: src/layouts/Api.astro
pubDate: 2026-08-11
modDate: 2026-08-11
title: Access Tokens
---

## Create an access token for the current user

:endpoint{method="POST" path="/api/users/access-token"}

**Response**

`200` — Contains the created access token.

- **`AccessToken`** :span[string]{.type-label}  
  Minimum length 1.

:::api-example{label="Response"}
```json
{
  "AccessToken": "string"
}
```
:::
````

It's normal markdown, with these enhancements:

- We use `:endpoint{method="POST" path="/api/users/access-token"}` to emit ``<span class="api-post">POST</span> `/api/users/access-token` ``. CSS styles the API badge to make it look nice, and the navigation generator also has something it can latch onto for reliable nav-generation
- We similarly use `:span[string]{.type-label}` to emit `string` with the `type-label` CSS class
- There is a special `:::api-example{label="Request|Response"}` directive which emits a div for example request/responses. We need something special here because the examples go in a right-hand side column, and Markdown itself has no way to express columns or layout. The satteri plugin sees this directive and emits the example HTML in the right-hand column rather than in-line with the other content.

It looks like this

image

### CODE REVIEW

**Comment by team-marketing-branch-protections** [comment]
Pull request environment is available at https://stoctodocspr3376.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3376&status=all).

File: src/layouts/Api.astro [hunk 1]
```diff
@@ -0,0 +1,127 @@
...
+          headings={headings}
+          lang={lang}
+          breadcrumbs={breadcrumbs}
+        />
+        <article itemscope itemtype="https://schema.org/Article">
+          <ArticleHeader lang={lang} subtitle={subtitle} title={title} />
+          <div class="page-actions">
+            {
+              /* Copy as markdown temporarily disabled until we can get it working with the API docs. Deliberately no "Edit on GitHub" because these are generated and should not be hand edited */
+            }
```
enf0rc3, line 92: Worth creating a ticket for this so we don't lose it, someone else can work on it?

rosslovas (dismissed):
Haven't done a review, just had a quick play on the test site:

> **When this PR merges, the new files under `/docs/api` will be live, but _nothing will link to them_**. 

I'm able to find & navigate to them via search results, blocking the PR just in case

Image

enf0rc3:
I have done some simple testing:

Feel free to take these offline into separate PRs, or do later if you feel its out of scope of this pr (hence why adding as a comment) 

- Left sidebar nav doesnt have "expand" carats, indicating it can expand until it already is
- Left sidebar nav doesnt keep location when clicking on a top level, i need to re-scroll down to find the opened one. 
- When "Responsive new nav" is merged double check sidebar correctly works. (also make sure we wire up the search)

rosslovas:
@enf0rc3 

> - Left sidebar nav doesnt keep location when clicking on a top level, i need to re-scroll down to find the opened one.

FYI that's the case for the main site as well ([example page](https://octopus.com/docs/octopus-rest-api/octopus-cli/install-global-tool)). It's been on my mind for a while as something I'd really like to see fixed!

src/layouts/Api.astro
```diff
@@ -0,0 +1,127 @@
+---
+import { accelerator } from '@lib/accelerator';
+import { PostFiltering } from 'astro-accelerator-utils';
+import type { Frontmatter as OriginalFrontmatter } from 'astro-accelerator-utils/types/Frontmatter';
+import { SITE } from '@config';
+import { getImageInfo } from '@util/custom-markdown.mjs';
```
enf0rc3: **Fix now.** This layout was forked from `Default.astro` before the breadcrumb refactor, so it still uses the old `Breadcrumbs` API. The net effect is that every API page renders an empty `<nav><ol></ol></nav>` and the BreadcrumbList JSON-LD is missing. There is no `astro check` in CI, so nothing catches the prop mismatch.

This is the first of four small suggestions that together restore it (they batch-apply). First, the import:

```suggestion
import { getImageInfo } from '@util/custom-markdown.mjs';
import { buildCrumbs, type Crumb } from '@util/breadcrumbs';
```

While you are here, line 31 declares `breadcrumbs` inline as `{ url; title; ariaCurrent? }[]`. `Default.astro` uses `Crumb[] | null`, which the import above brings in.

src/layouts/Api.astro
```diff
@@ -0,0 +1,127 @@
...
+type Props = {
+  // `apiMethods` is not written by hand: plugins/satteri-api-examples.js leaves
+  // it on the frontmatter for the left nav, an entry per endpoint.
+  frontmatter: OriginalFrontmatter & {
+    apiMethods?: { text: string; method: string | null }[];
+  };
+  headings: { depth: number; slug: string; text: string }[];
+  breadcrumbs?: { url: string; title: string; ariaCurrent?: string }[] | null;
+};
+const { frontmatter, headings, breadcrumbs } = Astro.props satisfies Props;
```
enf0rc3, line 34: Second of four. Build the crumbs, matching `Default.astro`:

```suggestion
const { frontmatter, headings, breadcrumbs } = Astro.props satisfies Props;

const crumbs = buildCrumbs(Astro.url, breadcrumbs);
```

src/layouts/Api.astro
```diff
@@ -0,0 +1,127 @@
...
+  accelerator.urlFormatter.formatAddress(Astro.url.pathname) === searchUrl;
+
+const showSearch = !isSearchPage;
+
+// The footer prints this; the article keeps it in its schema.org markup.
+const lastUpdated = frontmatter.modDate ?? frontmatter.pubDate ?? null;
+---
+
+<html dir={textDirection} lang={lang} class="initial" data-theme="light">
+  <Head frontmatter={frontmatter} headings={headings} lang={lang} />
```
enf0rc3: Third of four. `HtmlHead` passes `crumbs` straight to `<JsonLd>`, so without this the API pages ship no BreadcrumbList structured data:

```suggestion
  <Head
    frontmatter={frontmatter}
    headings={headings}
    lang={lang}
    crumbs={crumbs}
  />
```

src/layouts/Api.astro
```diff
@@ -0,0 +1,127 @@
...
+         left nav lists this page's endpoints, so there is no table of contents
+         to show, and the width is what the examples column is built out of. -->
+    <div class="content-group content-group--api">
+      <main id="site-main">
+        <Breadcrumbs
+          frontmatter={frontmatter}
+          headings={headings}
+          lang={lang}
+          breadcrumbs={breadcrumbs}
+        />
```
enf0rc3: Fourth of four, and the actual bug. `Breadcrumbs.astro` takes `{ lang, crumbs }` — it never reads `frontmatter`, `headings`, or `breadcrumbs`. So `crumbs` arrives `undefined` and `(crumbs ?? []).map(...)` renders an empty list on every API page.

```suggestion
        <Breadcrumbs lang={lang} crumbs={crumbs} />
```

src/pages/docs/api/_index.md
```diff
@@ -0,0 +1,110 @@
+# API reference
```
enf0rc3: **Fix now.** This file has no frontmatter at all — it and `README.md` are the only two of the ~100 that don't. Consequences:

- No `layout`, so `/docs/api/` renders as bare markdown with no header, nav, footer, or styles.
- No `title`, so it has no `<title>` and nothing for `ArticleHeader` to render.
- No `pubDate`/`modDate`, so `sitemap.xml.ts` emits `<lastmod>undefined</lastmod>` for it.
- `isApiPage()` returns false, so it is missing from the API left nav *and* is not pruned from the main site nav by `withoutApiPages()` — the one page in the section that leaks into the main tree.

Dropping the H1 as well, since `ArticleHeader` renders the title and none of the other generated pages carry one:

```suggestion
---
layout: src/layouts/Api.astro
title: API reference
pubDate: 2026-08-11
modDate: 2026-08-11
---
```

**Separately, the 106 links on this page all 404.** They are written `](./access-tokens.md)`. Astro does not rewrite `.md` links, no plugin in `src/plugins` or `src/integrations` does, and this is the only file in the repo using that style — everything else uses absolute `/docs/...`. They resolve to `/docs/api/access-tokens.md`.

```sh
sed -i 's|](\./\([a-z0-9-]*\)\.md)|](/docs/api/\1)|g' src/pages/docs/api/index.md
```

Both fixes need to go into the generator too, or the next run reverts them.

If you would rather not carry an index page until the real landing page lands, the alternative is renaming this `_index.md` — Astro skips underscore-prefixed files, so `/docs/api/` simply 404s until you light the section up. I'd lean toward fixing it, since the list is useful and both fixes are mechanical.

src/pages/docs/api/_README.md
```diff
@@ -0,0 +1,112 @@
+# API reference documentation
```
enf0rc3: **Fix now.** Anything under `src/pages` becomes a route, so this publishes at `/docs/api/README/` on octopus.com — including "**Do not edit these files by hand**" and the generator internals. `prune-dist` only removes top-level directories outside `/docs`, so it survives the build.

Rename it `_README.md`. Astro excludes underscore-prefixed files from routing, and it stays exactly where it is for anyone reading the folder. Worth changing in the generator as well.

src/pages/docs/api/access-tokens.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: src/layouts/Api.astro
+pubDate: 2026-08-11
+modDate: 2026-08-11
+title: Access Tokens
+---
```
enf0rc3, line 6: **Decide now.** Using this file as a stand-in — it applies to all ~100.

`search.json.ts` and `sitemap.xml.ts` both glob `./**/*.md` under `src/pages/docs/`, and nothing here opts out. So on merge these pages are searchable on the docs site and submitted to Google, which doesn't quite match "nothing will link to them". Given the content has known flaws you plan to chip away at, getting them indexed before the section is finished seems like the wrong order.

There's also a size question: `search.json.ts` runs keyword extraction over every page's compiled content, and this adds 76k lines to a file every visitor downloads. Worth measuring the before/after.

`PostFiltering` reads `navSearch` and `navSitemap`, so opting out is two keys — added to the generator's frontmatter template, and removed when you light the section up:

```suggestion
---
layout: src/layouts/Api.astro
pubDate: 2026-08-11
modDate: 2026-08-11
title: Access Tokens
navSearch: false
navSitemap: false
---
```

Happy for this to be a deliberate "no, index them now" — just flagging that it's easier to withhold than to de-index.

enf0rc3 (commented):
Reviewed the infrastructure code only, skipping the generated content per your note. The plugin design is sound — deriving "first H2" from the tree rather than closure state keeps it stateless, and the ordering note for `apiExampleDirective` after `attributeMarkdown` in `astro.config.mjs` is exactly the kind of thing that gets lost otherwise. The Playwright tests are genuinely good, especially running them with `javaScriptEnabled: false` and asserting the column-stack breakpoint by bounding box.

One real bug and three routing issues. Grouped by urgency, keeping in mind that nothing links to `/docs/api` yet:

**Fix now** — these ship publicly on merge regardless of whether anything links to them:

1. Breadcrumbs render empty on every API page (wrong props to `Breadcrumbs`). Pure code bug, ~4 lines.
2. `README.md` publishes at `/docs/api/README/`, internal generator instructions and all.
3. `index.md` has no frontmatter, so `/docs/api/` renders as a bare unstyled page, and all 106 of its links 404.
4. A decision on `navSearch`/`navSitemap` — the ~100 pages enter the sitemap and site search on merge, so Google will index them before you light the section up.

**Can wait** — the dead CSS rule, `Astro.request.url`, the crawl skip regex, and the microdata/JSON-LD inconsistency. All flagged inline as deferrable.

Worth noting on 3 and 4: the durable fix belongs in the generator in the Server repo, since these files get overwritten. The suggestions below fix the checked-in copy so this PR is correct today.

I verified the component signatures against source rather than by building, and confirmed the markdownlint ignore works by reproducing your CI invocation locally — that part is fine.

src/styles/api.css
```diff
@@ -1,3 +1,156 @@
...
+}
+
+/* Inline code is dense enough on these pages that the box around each span
+   reads as noise, so it keeps the monospaced type and drops the chrome. Same
+   specificity as the callout rule in main.css, and later in the cascade, so it
+   wins inside callouts too. */
+.content-group--api code:not(pre code) {
+  /* border: none;
+  background: none; */
+}
```
enf0rc3: **Deferrable.** The rule body is entirely commented out, so this is an empty ruleset under a comment describing behaviour that doesn't happen — inline code on API pages keeps its border and background.

Either delete it:

```suggestion
```

or uncomment the two declarations if dropping the chrome was the intent. Left as-is it reads as working styling to the next person who comes through.

src/components/ApiNavigation.astro
```diff
@@ -0,0 +1,125 @@
...
+  return text.replace(/\$\{/g, '{').trim();
+}
+
+const _ = Lang(lang);
+
+// Only the page the reader is on lists its endpoints. Expanding every page
+// would put thousands of anchors in the markup of all ~100 of them, and it is
+// the rule the site nav follows too: the branch you are inside is the one that
+// opens.
+const currentPath = new URL(Astro.request.url).pathname.replace(/\/$/, '');
```
enf0rc3: **Deferrable, trivial.** `Astro.url` is what the rest of the codebase uses, including `Api.astro:57` two files over:

```suggestion
const currentPath = Astro.url.pathname.replace(/\/$/, '');
```

package.json
```diff
@@ -16,7 +16,7 @@
         "dev:watch": "onchange 'src/**/*.{js,mjs,ts,astro,css}' -- prettier --write --plugin=prettier-plugin-astro {{changed}}",
         "astro": "astro",
         "build": "astro build",
-        "crawl": "linkinator ./dist --skip \"^(?!http://localhost)\" --recurse  --verbosity error",
+        "crawl": "linkinator ./dist --skip \"^(?!http://localhost)\" --skip \"/docs/api(\\$|/)\" --recurse  --verbosity error",
```
enf0rc3: **Deferrable, but worth a note.** Two things:

1. The JSON `\\$` survives as a literal `$` under Windows `cmd`, which is what npm/pnpm shell out to, so the `$` alternative never matches there. The `/` alternative still covers every subpage, so the only gap is the bare `/docs/api` URL. On CI's `sh` it works as intended.

2. More importantly, this skip is why the 106 broken links on `index.md` don't fail the build. Once those are fixed it's worth checking whether the skip is still needed, or whether it can be narrowed — right now nothing validates links across 76k lines of generated content, and the generator could start emitting bad ones without anyone noticing.

Fine to leave for now; just worth not letting it become permanent.

src/layouts/Api.astro
```diff
@@ -0,0 +1,127 @@
...
+        <article itemscope itemtype="https://schema.org/Article">
```
enf0rc3: **Deferrable.** This uses schema.org microdata (`itemscope`/`itemprop`), which `Default.astro` dropped in favour of JSON-LD via `<JsonLd>`. Once the `crumbs` fix above is in, the site will emit two different structured-data schemes depending on which layout a page uses.

Not urgent and harmless for search engines, but worth collapsing onto JSON-LD next time this file is touched so the two layouts don't drift further apart.

enf0rc3:
Following on from my review above — I read the generator in OctopusDeploy/OctopusDeploy#46319 and left a longer review there about the format contract between the two repos:

**https://github.com/OctopusDeploy/OctopusDeploy/pull/46319#pullrequestreview-4957018861**

Summarising the part that lands on this repo, since some of it changes what I'd do with my earlier comments.

## The short version

The generated Markdown encodes this site's implementation details — the Astro layout path and CSS class names — alongside the API content. The clearest symptom is here in `src/plugins/satteri-api-examples.js`:

```js
function methodFromBadge(node) {
  return METHODS.find((method) => hasClass(node, `api-${method}`)) ?? null;
}
```

The generator has a `MethodCssClasses` dictionary mapping `GET → api-get`. So Swagger says `get`, the Server repo encodes it as `.api-get`, and we decode the class back into `get` to build the nav badge and its `aria-label`. The semantic value round-trips through a CSS class across two repositories.

The consequence for us: restyling a method badge currently requires a PR in the Server repo and 107 regenerated files. Adding PATCH support is the same. We can't style the endpoint path at all, because it arrives as plain inline code indistinguishable from any other backtick span.

## What I've proposed over there

The generator emits data, we decide the paint:

```markdown
:endpoint{method="POST" path="/api/users/access-token"}
:::deprecated
```

Plus dropping `layout:`, `pubDate:` and `modDate:` from the generated frontmatter entirely. Deferring the 16,893 `.type-label` spans and the bold pseudo-headings — big diffs, little reader-visible gain today.

## How this changes my earlier review

**The `index.md` and `README.md` fixes get better if we move `/docs/api` to an Astro content collection** with a single `[...slug].astro` route, rather than patching the files:

1. The layout comes from the route, so generated files need no `layout:` key — which is what lets the Server repo stop writing one.
2. Zod validation on the collection makes a malformed file **fail the build loudly**. The `index.md` bug — no frontmatter, renders as a bare unstyled page — becomes impossible rather than something we happened to catch in review.
3. `README.md` can't publish as a public page. It's already listed as hand-written on the Server side, so the shovel just doesn't copy it.

So I'd treat my `index.md` frontmatter suggestion and the `_README.md` rename as **interim fixes if this merges first**, and as unnecessary if we do the collection move now.

**Still worth fixing here regardless of any of the above:** the empty breadcrumbs (four suggestions above — that's a plain bug, unrelated to the format question), and the `navSearch`/`navSitemap` decision.

## One thing to correct in this PR either way

The comment on `methodFromCode` in `satteri-api-examples.js` describes it as handling "the method an older page spells as inline code". It's actually the live fallback for any method the generator has no CSS class for — `MethodOrder` over there already lists PATCH/HEAD/OPTIONS/TRACE while `MethodCssClasses` covers four. The first PATCH endpoint added to the Server gets no badge and no `aria-label`, silently, in both repos. Zero exist today so it's dormant, but the comment as written invites someone to delete that branch as legacy.

## Suggested sequencing

Given nothing links to `/docs/api` yet, the cheapest order is to agree the format with Orion first, then land both sides together — roughly a day each. If that's more coordination than you want right now, merging this as-is and treating the format as a follow-up is reasonable, with the caveat that the byte-for-byte approval tests over there make every later format change cost a full regeneration.

borland:
@enf0rc3 @rosslovas I've pushed a commit which should address the concerns raised. It temporarily excludes the API folder from search and the sitemap, fixes search integration, json-ld, breadcrumbs, etc

> Left sidebar nav doesnt have "expand" carats, indicating it can expand until it already is

This wasn't intentional (claude just did it and I didn't notice), but thinking about I think I'd prefer to leave it as-is.

In the main site, the carats mean "Expand" but that typically happens if a node represents a **folder** with multiple documents within it.

The nav here is one level down. Each entry is a **single document** and the only thing within the document is multiple H2's. This is equivalent to a normal page in the main docs, and those don't have expanders.

Also, in the near future we're going to intermix human-written content into the api area, and that might well have nested folders, where the expander would have meaning.

src/pages/docs/api/access-tokens.md
borland: Rather than doing this I got claude to have an `Under Construction` check for the docs folder

enf0rc3: This is better than what I suggested, for a reason that matters beyond this PR: it keeps site-specific keys out of the generated files. My version would have had the Server generator emitting `navSearch`/`navSitemap`, which is more of exactly the coupling I was arguing against elsewhere. `underConstruction.ts` puts the decision in the consuming repo where it belongs, and it's one file to delete when the section goes live rather than a frontmatter change across 106 files plus a generator change.

Checked both call sites — `search.json.ts` and `sitemap.xml.ts` both skip before any processing, so the pages stay out of the index and the sitemap while still building. That's what I was after.

enf0rc3:
*Edited: an earlier version of this comment flagged a brace-escaping divergence between the two repos. That has since been fixed on the Server side and the section is updated below to say so.*

Everything I raised is addressed, and a few of the fixes are better than what I suggested.

**Verified fixed:**

- **Breadcrumbs** — and `buildApiCrumbs` is the right call rather than the plain `buildCrumbs` I suggested. Since `_index.md` is underscore-prefixed there's no page at `/docs/api`, so the generic walk would have skipped straight from Docs to the endpoint page. Splicing in a urlless section crumb is correct.
- **`_index.md` / `_README.md`** — both out of routing, and removed on the Server side too.
- **`Astro.url.pathname`**, and the dead `code:not(pre code)` rule is gone.
- **Microdata** — no `itemscope`/`itemprop` left in `Api.astro`, so the two layouts no longer emit different structured-data schemes.
- **Search and sitemap** — `underConstruction.ts`, replied on that thread. Better than my suggestion.
- **The `crawl` skip is gone**, so linkinator now validates links across the API section. That was only possible because the 106 broken `./*.md` links left with `index.md`.
- **`:endpoint`** — the plugin reads `data-method`/`data-deprecated`, so the CSS-class round trip through `methodFromBadge` is gone. Reporting an error on a missing method and a warning naming `api.css` on an unbadged one closes the silent PATCH gap properly.

## Brace escaping — resolved

Briefly the two repos disagreed here: the `:endpoint` paths in this repo were escaped and the ones committed to the Server PR were not, which would have truncated every path at the first `}` once a shovel ran.

Fixed on the Server side by `EscapeDirectiveValue` in `ApiReferenceGenerator.cs` (commit "We need to escape directive values"). Re-checked after the force-push:

- `source/ApiDocs/channels.md` and `src/pages/docs/api/channels.md` are now **byte-for-byte identical**.
- All 20 endpoints in that file are escaped, and both copies parse through Sätteri to the same `path` attribute, which `unescapePath` turns back into `/api/{spaceId}/channels`.

`tests/api-endpoint.spec.ts:26` covers this from now on — it asserts the rendered path reads `/api/{spaceId}/feeds`, so a regression on either side goes red here.

## One small nit

`buildApiCrumbs` in `src/lib/apiNavigation.ts`: if `findIndex` doesn't locate the `/docs` crumb it returns `-1`, and `splice(insertAt + 1, ...)` then inserts at index 0 — so the section crumb lands *before* Docs rather than the insert being skipped. I haven't found a route where the docs crumb is actually missing, so this may well be unreachable. Guarding it with `if (insertAt === -1) return crumbs;` would make it fail safe either way.

## Still open upstream, affects this repo

`pubDate`/`modDate` are still hardcoded to `2026-08-11` in the generator, so every page here reports that as "last updated" in the footer, and it's what `sitemap.xml.ts` would emit as `<lastmod>` once the section comes out from under construction. Chased it on the Server PR. Not blocking here.

enf0rc3 (approved):
Approving. The infrastructure code is in good shape and everything blocking from my earlier reviews is fixed.

**Verified against `a1fb952`:**

- Breadcrumbs render properly via `buildApiCrumbs`, and `Head` gets `crumbs` so the JSON-LD is there.
- `_index.md` and `_README.md` are out of routing, so nothing publishes unstyled and the internal README stays internal.
- The `crawl` skip is gone, so linkinator validates links across the API section.
- `underConstruction.ts` keeps the section out of site search and the sitemap until it goes live, and both call sites skip before any processing.
- `:endpoint` carries `method`, `path` and `deprecated` as data. The plugin errors on a missing method and warns naming `api.css` on an unbadged one, so a future PATCH endpoint surfaces at build time instead of silently losing its badge.
- Brace escaping is resolved on both sides. I diffed five generated pages against `source/ApiDocs` at `a28f5d5` — byte-for-byte identical, and the only file that differs across the two repos is `_README.md`, which correctly stays in the Server repo.

**Deliberately deferred**, recorded here so they are findable later rather than lost:

1. `pubDate`/`modDate` hardcoded to `2026-08-11` in the generator, so pages report a fixed "last updated" date. Matters most once the section leaves under-construction and reaches the sitemap.
2. `layout: src/layouts/Api.astro` written by the generator, naming a docs-repo path from the Server repo. Worth revisiting when the shovel is built, along with a possible move to an Astro content collection.
3. `buildApiCrumbs` inserts at index 0 if `findIndex` returns `-1`, putting the section crumb before Docs. I could not find a route where that happens, so this is defensive only.

None of these affect readers today. Companion PR is OctopusDeploy/OctopusDeploy#46319.

PR #3379, opened 2026-08-18
Fix macOS check when assigning cmd/ctrl+K keybind

rosslovas:
Prior to #3369 there was a single macOS check which was used both for visually displaying a `⌘K` to indicate that shortcut would focus the search bar, and assigning a `keydown` listener for that key combination.

https://github.com/OctopusDeploy/docs/blame/a530d440d6627449c113e4c670097ebe548c2f33/src/components/TopNavSearch.astro#L104-L106

#3369 added a second, different check, which doesn't work in Chrome. In that browser `⌘K` still displays, but you have to press Ctrl+K due to the check failing. This PR changes that second check to match the one that's used to display the key combination so it works in Chrome again.

### DIFF
--- a/src/scripts/docs-search.ts
+++ b/src/scripts/docs-search.ts
@@ -316,10 +316,10 @@ function setup(dialog: HTMLDialogElement) {
   if (name !== SITE_SEARCH) return;
 
   // Cmd+K on Apple platforms, Ctrl+K everywhere else.
-  const platform =
-    (navigator as Navigator & { userAgentData?: { platform: string } })
-      .userAgentData?.platform ?? navigator.platform;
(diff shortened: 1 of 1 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3379.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3379&status=all).

borland (approved):
Looks good. Out of curiousity I checked for iPads (an iPad with a keyboard attached is quite common); and since iOS 13 an iPad will report itself as a mac so we should be good

PR #3384, opened 2026-08-18
Turn "Copy as markdown" into a split button and delete "Use Octopus docs with AI"

enf0rc3:
[NES-294](https://linear.app/octopus/issue/NES-294/turn-copy-as-markdown-into-a-split-button-and-delete-use-octopus-docs)

"Open this page as markdown" moves from the menu at the bottom of the article into a dropdown on the copy action, built on the shared `SplitButton` that "Open in Claude" uses. `MarkdownLinks.astro` and its script are deleted.

"Open all docs as markdown" is dropped rather than moved — `llms-full.txt` is nearly 2 million tokens and unusable with any current model. The route and its tests stay; only the UI link goes.

**Behaviour change:** a failed copy no longer reports anything. It used to show "Copy failed"; it now leaves the button as it was and announces nothing. The heading and code-block copy buttons keep their own failure tooltips.

`SplitButton` takes two changes and no new props: `href` becomes optional, and the primary gets a `<slot />`. The class and `data-copy-md-url` sit on the control, and the copy listener matches the primary half so the caret does not copy the page on its way to opening the menu.

The copy action's styles move out of `main.css` into the component, alongside how `OpenInLlm` and `Menu` already do it, so `main.css` only loses lines here.

The menu item uses design's `markdown.svg`, checked at 16px in both themes.

# Before 
image
image

# After
image


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3384.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3384&status=all).

src/components/MarkdownLinks.astro
```diff
@@ -1,57 +0,0 @@
----
```
enf0rc3, line 1: Image 

This component

src/components/SplitButton.astro
```diff
@@ -40,6 +41,12 @@ if (items.length === 0) {
   throw new Error('A SplitButton with no menu items should just be a Button');
 }
 
+// Button types the two halves of its own union separately - an anchor needs an
+// `href`, a plain button forbids one - so the link attributes are handed over as
+// a set rather than as three possibly-undefined props. `target` and `rel` say
+// where a link opens, and mean nothing on a button.
+const primaryLink = href ? { href, target, rel } : {};
```
enf0rc3, line 48: allows "non href" button that doesn't link anywhere. (client side only)

src/assets/icons/markdown.svg
rosslovas, line 1: Ellen provided an icon in Slack that looked quite different to this one. I see in the PR description it's mentioned that it should be switched.

enf0rc3: Oh didn't notice that, ill update it to use the one she provided, thanks for spotting that!

enf0rc3: @rosslovas Done thanks!

rosslovas (approved):
Looks good!

PR #3388, opened 2026-08-19
Ensure top right state is correct on initial render

rosslovas:
# Summary

The initial state of the elements in the top-right of the new nav bar (theme switcher, sign-in buttons or avatar) was always initially rendered in the light theme + signed-out state. As soon as the theme and the signed-in state was known, it would change to the correct state.

This manifested as a visible flicker that could be annoying/distracting.

This ensures the appropriate scripts to determine the theme and signed-in state run ASAP synchronously, before the browser even gets to the point of rendering these elements, such that they're always rendered in the correct state immediately and never change.

# Results

Test it out at https://stoctodocspr3388.z22.web.core.windows.net/docs; navigate between a few pages in dark mode and observe that the top right should never show the moon icon to change to the dark theme, it should always show the sun icon indicating it's currently in dark mode.

Same deal with the signed-in state. Note that this isn't _quite_ perfect as the initials/avatar image do not load synchronously, so they can flicker into existence, but the scope of this PR is just that the _Sign in_ and _Start for free_ buttons **should never appear when signed in**. It can be tested with:

```js
document.cookie = 'OctopusSignedInUser=' + encodeURIComponent(JSON.stringify({fullName: 'John Doe', email: 'jd@example.com'})) + ';path=/';
window.dispatchEvent(new CustomEvent('octopus:user-changed'));
```

Undo:
```js
document.cookie = 'OctopusSignedInUser=;path=/;max-age=0';
window.dispatchEvent(new CustomEvent('octopus:user-changed'));
```

### DIFF
--- a/src/components/ProfileMenu.astro
+++ b/src/components/ProfileMenu.astro
@@ -93,10 +93,6 @@ const items: MenuItem[] = [
     display: flex;
   }
 
-  .profile-menu[hidden] {
-    display: none;
-  }
-
   .profile-menu__trigger {
     display: flex;
     border-radius: var(--borderRadiusCircle);
--- a/src/components/TopNav.astro
+++ b/src/components/TopNav.astro
@@ -102,12 +102,12 @@ const sectionItems: MenuItem[] = links.map(({ label, href }) => ({
       <div class="top-nav__drawer-theme">
         <span class="top-nav__drawer-theme-label">Theme</span>
         <Button
-          icon="theme-switcher__moon_icon"
+          icon="top-nav__theme-icon"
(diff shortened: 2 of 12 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3388.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3388&status=all).

borland (approved):

PR #3391, opened 2026-08-19
Keep crawler submissions out of the feedback responses

enf0rc3:
The feedback responses are carrying rows nobody submitted: pages from dev and preprod, and page URLs ending in prototype pollution probes (`#__proto__[sssied]=sssieda&...`). Something crawls the docs and works the widget.

Two changes:

- A scripted click, or a click in a browser reporting `navigator.webdriver`, gets the same thanks a reader gets and sends nothing.
- The URL fragment is only kept when it names an element on the page, so appended junk is dropped and real anchors survive.

Best effort, so a determined crawler still gets through. `npx playwright test tests/feedback.spec.ts` covers both, 10 passing.

Additional "hardening" coming in another pr.

### DIFF
--- a/src/scripts/modules/feedback.js
+++ b/src/scripts/modules/feedback.js
@@ -8,6 +8,22 @@ import {
 } from './feedback-form.js';
 
 /**
+ * Crawlers walk the docs on every environment and work each control they find,
+ * and their submissions land among the readers' in the same responses. Two
+ * things separate them from someone reading the page: a click a script
+ * dispatched carries `isTrusted` false, and a browser being driven says so on
+ * `navigator.webdriver`, which the scanners running headless Chrome set as
+ * surely as our own Playwright run does. Neither is proof and a determined
+ * crawler defeats both, so this thins the noise rather than sealing anything.
+ *
+ * @param {Event} event
+ * @returns {boolean}
+ */
+function automated(event) {
+  return navigator.webdriver === true || !event.isTrusted;
+}
+
+/**
  * Google Forms sends no CORS headers, so the POST has to go out as no-cors and
  * the response comes back opaque. There is no way to read whether the form
  * accepted it - a rejected submission looks identical to an accepted one, so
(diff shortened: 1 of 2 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3391.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3391&status=all).

borland (changes requested):
This is not the right approach to solve this problem. It looks like the PR code changes are designed to defeat an automated test suite using something like playwright.

But I don't think that's what's happening here. Security scanners don't play by the same rules as automated testing tools.

There's no way to know exactly what they're doing without asking SecOps exactly which scanners they're running - and even then the scanner might change it's behaviour over time.

Based on experience and looking at the analysis from claude, I think it's unlikely that the scanners are running any kind of browser or javascript at all. Rather, I'd guess they are directly loading this JS here: https://octopus.com/docs/_astro/Feedback.astro_astro_type_script_index_0_lang.BCHktk8f.js and then observing the google form parameters and just firing data directly.

> 1. Parameter discovery / reflection probing — hundreds of key=value pairs where every value is sssiedh<paramname>xsx. That's a canary wrapper: a fixed prefix/suffix around the param name so the tool can grep the response for reflections and identify which params the app actually reads. The name lists are generic wordlists (token, password, cookie, admin, redirect, SAMLResponse, __RequestVerificationToken) **blended with names harvested from your own JS** — THEME_STORAGE_KEY, loadMarketo, __marketoLoader, googleTranslateElementInit, newNavRequested, useNewNav, themeSwitchers, dreamdataTrackingId, plausible.q, storyblokInstance. That's Burp Suite's Param Miner behaviour almost exactly (client-side param discovery mode).

If that's the case, then we cannot defeat them using javascript-based techniques like this

src/scripts/modules/feedback.js
```diff
@@ -7,6 +7,22 @@ import {
...
+ * dispatched carries `isTrusted` false, and a browser being driven says so on
+ * `navigator.webdriver`, which the scanners running headless Chrome set as
+ * surely as our own Playwright run does. Neither is proof and a determined
+ * crawler defeats both, so this thins the noise rather than sealing anything.
+ *
+ * @param {Event} event
+ * @returns {boolean}
+ */
+function automated(event) {
+  return navigator.webdriver === true || !event.isTrusted;
```
borland, line 23: It is highly unlikely that a security scanner would politely identify itself as webdriver or set the isTrusted flag to false. This check seems somewhat worthless

enf0rc3:
> Posted by Claude (an AI agent) on Will's behalf.

Closing this — you were right that it's the wrong approach.

The `isTrusted` check does nothing here. The response data points at a browser-driven crawl: 6–11 seconds between submissions, real in-page anchors like `#move-imports-folder`, and fragments assembled from identifiers harvested off our own pages (`THEME_STORAGE_KEY`, `useNewNav`, `theme-switcher-mobile`). A browser driven over CDP produces trusted events, so that check never fires. `navigator.webdriver` depends on how Chrome is launched, which makes it a coin flip.

The scanner is also ours, so the first fix is a SecOps exclusion rather than code.

Notes for whoever picks this up if that doesn't work out:

- The junk started 88 minutes after #3324 merged. The old `<a>` link was never touched, because submitting meant leaving the domain.
- Every page carries `<link rel="canonical">`, so the reported URL can come from the page instead of the address bar and can't be wrong.
- Reject on a fragment that names no element rather than stripping it. Stripping turns scanner rows into clean rows that look like real feedback.
- The obfuscation idea is the right fallback if SecOps can't exclude us. It defeats automated endpoint harvesting, which is exactly what it's aimed at.

#3392 closed alongside this.

PR #3392, opened 2026-08-19
Honeypot, reading pause and a once-a-day limit on feedback

enf0rc3:
Stacked on #3391. Three cheap guards, all of them client side:

- A honeypot input a reader never sees, reaches by keyboard or hears read out. Anything in it means something filled in every input it found.
- A minimum gap between the vote click and Send. The Send button only appears on the vote, so a reader has to travel to it.
- One answer per page, per browser, per day, kept in `localStorage`.

The comment field is left alone. Readers paste URLs into it legitimately, so a rule against `http` would cost real feedback, and a length cap alone buys nothing.

**Open question before this is worth merging.** Every guard here, and in #3391, requires the sender to be running the page. If the scanner reads the entry IDs out of the public bundle and posts to the Google Form endpoint directly, none of this touches it. The counter-evidence is that the junk rows carry well-formed `Title - URL` values with the correct per-page title, which only our JS builds. That points at a browser, though it does not prove one.

Either way, the only defence that survives a direct post is response validation on the form itself.

`npx playwright test tests/feedback.spec.ts` covers all three, 14 passing.

### DIFF
--- a/src/components/Feedback.astro
+++ b/src/components/Feedback.astro
@@ -52,6 +52,22 @@ stats.stop();
     </div>
   </div>
 
+  {
+    /* Nothing a reader sees, reaches by keyboard or hears read out, so anything
+      in it was put there by something filling in every input it found. Held
+      outside the comment block, which starts hidden, as a crawler working the
+      page in one pass would never see it there. */
+  }
+  <input
+    class="feedback__trap"
+    type="text"
+    name="feedback-website"
+    tabindex="-1"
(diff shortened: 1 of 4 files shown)


enf0rc3:
> Posted by Claude (an AI agent) on Will's behalf.

Closing alongside #3391. These guards are aimed at our own security scanner, and the fix belongs with SecOps rather than in client-side JS. The findings are written up on #3391.

PR #3394, opened 2026-08-19
Experiment: stronger bot gating on the feedback widget

enf0rc3:
An automated scanner loads docs pages with a payload in the URL fragment and clicks every button it finds, which submits a thumbs-down response carrying that payload in the page field. It has been doing this from every environment, dev and staging included, since #3324 merged.

Votes and sends now need a trusted event, a browser that does not report itself as automated, a few input events ahead of the click, and enough time on the page. The send button ships disabled so it cannot be clicked into. The fragment is kept only where it names a heading that is on the page.

Draft and experimental. The thresholds are a first guess and want real traffic to confirm, and a scanner driving Chrome over CDP at reading pace would still get through.

For review: an empty comment still submits, which is what lets a run that never types reach the form at all. Requiring text would close that, and would also stop readers sending a bare thumbs-up.

13 Playwright tests pass against the dev server. `npm run build` fails on a missing `html-to-text` dependency, which also fails on `main`.

### DIFF
--- a/src/components/Feedback.astro
+++ b/src/components/Feedback.astro
@@ -58,10 +58,16 @@ stats.stop();
     </label>
     <textarea class="feedback__textarea" id="feedback-comment" rows="3"
     ></textarea>
+    {
+      /* Disabled here and released by the script on a vote it accepted, as the
+        browser dispatches no click at all on a disabled button. Released as
+        this container is revealed, so the state is never on screen. */
+    }
     <Button
       label={_(Translations.octopus_feedback.send)}
       size="small"
       importance="loud"
+      disabled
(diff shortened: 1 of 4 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3394.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3394&status=all).

enf0rc3:
Probably not required right now, but good for reference

PR #3395, opened 2026-08-19
Bounce feedback to the prefilled Google form instead of posting it

enf0rc3:
Back-pocket alternative to #3393. Keeps the sidebar UI and stops the page submitting anything at all.

Send stops being a form post and becomes an `<a>` to the form's own `viewform` URL, with the fields prefilled. The reader presses submit on Google's side, so nothing on our page can be made to submit — which is what protected us before #3324, since scanners scope to the target domain and don't follow off-site links.

## What it does

- Send is a link, not a button. Its `href` carries `usp=pp_url` plus the page, the rating and the comment, rebuilt as the reader votes and types.
- It works without our script. Astro renders the link at build time with the page title already prefilled, exactly as the pre-#3324 button did. The script only enriches it.
- It opens in its own tab. `setExternalLinkAttributes()` already adds `target="_blank"` and `rel="noopener"` to any off-site link, so the reader keeps their place in the docs. That's better than the old button, which navigated away.
- The thanks message is gone. We can't know whether someone submitted on Google's side, so claiming it would be a lie.
- Label changed from "Send" to "Send feedback", since a button that opens a form elsewhere shouldn't say "Send".
- A comment over 1500 characters is cut, because a prefill travels in the URL. The reader sees the cut text in the form and can finish it there.

## Verified against the live form

I loaded a prefilled URL in a browser and checked all three fields land (viewing only, nothing submitted):

| Field | Prefilled |
|---|---|
| Which page did you view? | `Deploy with Kustomize - https://octopus.com/docs/kubernetes/steps/kustomize` |
| How useful was the content? | `5` selected on the 1–5 scale |
| Please tell us what we did well… | the comment text |

`npx playwright test tests/feedback.spec.ts` — 8 passing, including one that asserts no request reaches `docs.google.com` however the widget is worked.

## Worth knowing before choosing between this and #3393

**Both models require a comment.** The form marks "Please tell us what we did well, or what we could improve" as **required**. So under this PR a reader who clicks a thumb still has to type something on Google's page before their rating counts. That is almost certainly what kept the old flow's volume low, and it means the two approaches converge for the reader: no comment, no feedback.

If we want vote-only feedback in either model, the form field has to be made optional first. That's a Forms settings change, not a code one.

**The remaining cost here is friction.** Every response now needs a page load on another domain and a second click to submit. #3393 keeps the submission inline and pays for it with a filter that is targeted at the scanner's current behaviour.

Draft on purpose — this is the fallback if the required-comment approach in #3393 stops working, or if we decide we want the structural guarantee now rather than later.

### DIFF
--- a/src/components/Feedback.astro
+++ b/src/components/Feedback.astro
@@ -3,7 +3,11 @@ import { accelerator } from '@lib/accelerator';
 import type { Frontmatter } from 'astro-accelerator-utils/types/Frontmatter';
 import { Lang, Translations } from '@util/Languages';
 import Button from './Button.astro';
-import { RATING_NO, RATING_YES } from '../scripts/modules/feedback-form.js';
+import {
+  prefillUrl,
+  RATING_NO,
+  RATING_YES,
+} from '../scripts/modules/feedback-form.js';
 
 const stats = new accelerator.statistics('components/Feedback.astro');
 stats.start();
@@ -20,6 +24,11 @@ const pageTitle = await accelerator.markdown.getTextFrom(frontmatter.title);
 
 // Language
 const _ = Lang(lang);
+
+// The link the reader follows before any of the widget's own script has run,
+// carrying what is known at build time. The script adds the rating, the comment
+// and the full address once someone starts answering.
+const feedbackUrl = prefillUrl(pageTitle, null, '');
 
 stats.stop();
 ---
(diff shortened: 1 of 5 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3395.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3395&status=all).

PR #3398, opened 2026-08-19
Introduce an "area" concept, so we aren't locked into a single layout under the "API" section of the docs site.

borland:
# Background

The docs site now has two distict sections; the main docs, and the api docs.

API docs have a different layout with different column sizing, and they also have their own specific navigation tree, which is tailored to the API, and does things like rendering icons next to GET/POST/PUT requests.

Previously, the layout dictated which area/navigation tree a page was put in. If a page wanted to be in the API tree, then it was forced to use the API's column structure. We want to mix in user-written pages along with the auto-generated ones, and forcing user-written pages to use the API column structure just looks bad.

Also, in future, we intend to have areas for "Learn" and "CLI", which would have their own nav trees

# Results

This PR adds the formal concept of an `area`. The area is now what dictates which navigation tree a page appears in, and we can use different layouts within any given area.

- Pages can now specify an area explicitly via their yaml frontmatter. 
- If they don't, an area will in inferred from the path. Things in the `docs/api` folder get the API area, everything else gets the Docs area. 
- We can extend this to make areas for CLI/Learn later

**Note**: The TopNav hasn't been touched. We should update it so when you click "API" it takes you to the API area's landing page, but that will come later

### DIFF
--- a/src/lib/areas.ts
+++ b/src/lib/areas.ts
@@ -1,0 +1,156 @@
+import type { MarkdownInstance } from 'astro-accelerator-utils/types/Astro';
+import { SITE } from '@config';
+import { buildCrumbs, type Crumb } from '@util/breadcrumbs';
+import { accelerator } from './accelerator';
+
+// An "area" is a slice of the site with its own navigation tree: the docs, the
+// API reference, and — as they arrive — Learn and the CLI reference. It is
+// deliberately not the layout. The layout decides what the middle column looks
+// like; the area decides which nav sits beside it and how the breadcrumbs read,
+// so a page can be in the API area and still be laid out like a docs page.
+//
+// A page's area comes from where it lives, which is what the section already
+// encodes: every page under /docs/api is in the API area without being told so,
+// including the ~100 generated ones. A page that needs to say otherwise puts
+// `area:` in its frontmatter and that wins.
+//
+// Adding an area takes three things: an entry below, a case in
+// components/AreaNavigation.astro, and the nav component it points at.
+
+export type Area = 'docs' | 'api';
+
+type AreaDefinition = {
+  // The path the area occupies, under SITE.subfolder. The docs area is
+  // everything left over, so it claims no path of its own.
+  path: string | null;
+  // A crumb to splice in for the section, for an area with no landing page for
+  // the generic breadcrumb walk to find. null when the walk finds one.
+  sectionCrumb: string | null;
+};
+
+export const AREAS = {
+  docs: { path: null, sectionCrumb: null },
+  // /docs/api has no page of its own yet: `_index.md` is underscore-prefixed so
+  // Astro does not route it, which is why the crumb has to be spliced in.
+  api: { path: '/api', sectionCrumb: 'Api' },
+} as const satisfies Record<Area, AreaDefinition>;
+
+const DEFAULT_AREA: Area = 'docs';
+
+// Longest path first, so an area nested inside another one still wins.
+const PATH_AREAS = (Object.entries(AREAS) as [Area, AreaDefinition][])
+  .filter(([, definition]) => definition.path !== null)
+  .sort((a, b) => (b[1].path as string).length - (a[1].path as string).length);
+
+const trimSlash = (path: string) => path.replace(/\/$/, '');
+
+function isArea(value: unknown): value is Area {
+  return typeof value === 'string' && value in AREAS;
+}
+
+/**
+ * The area a path falls in, before any frontmatter has its say.
+ *
+ * The prefix has to end on a path boundary, or /docs/api-and-integration would
+ * read as part of the API reference.
+ */
+export function areaFromPath(path: string): Area {
+  const pathname = trimSlash(path);
+
+  for (const [area, definition] of PATH_AREAS) {
+    const prefix = SITE.subfolder + definition.path;
+    if (pathname === prefix || pathname.startsWith(prefix + '/')) return area;
+  }
+
+  return DEFAULT_AREA;
+}
+
+/**
+ * The area for a page, given whatever its frontmatter declared and the path it
+ * is served from. An unrecognized `area:` falls back to the path rather than
+ * dropping the page out of every nav on the site.
+ */
+export function resolveArea(declared: unknown, path: string): Area {
+  if (declared != null) {
+    if (isArea(declared)) return declared;
+    console.warn(
+      `Unknown area "${String(declared)}" on ${path}. Falling back to the area its path implies. Known areas: ${Object.keys(AREAS).join(', ')}.`
+    );
+  }
+
+  return areaFromPath(path);
+}
+
+/**
+ * The area for a page read back from the post list.
+ *
+ * Astro only gives a page module a `url` when its file resolves inside
+ * src/pages, and the generated API pages are symlinked in during local
+ * development — so those modules arrive with no url at all. `urlHint` is for
+ * the caller that can rebuild it (see apiNavigation.urlsByFile).
+ */
+export function pageArea(post: MarkdownInstance, urlHint?: string): Area {
+  return resolveArea(post?.frontmatter?.area, post?.url ?? urlHint ?? '');
+}
+
+// Pages whose frontmatter puts them in a different area than their path would.
+// The site nav works from a tree of urls with no frontmatter attached, so the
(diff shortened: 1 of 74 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3398.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3398&status=all).

enf0rc3 (commented):
> Drive-by review: I pointed Claude Code at this PR, so the findings below are an AI's rather than mine. I checked the line references and grepped the counts it cites, and left the rest unverified. Ignore whatever is wrong or not worth the effort.
> 
> The design reads well. Things it checked that could have broken quietly and didn't: all 106 pages under `src/pages/docs/api` use `Api.astro` and no `Api.astro` page lives outside that path, so the layout→path switch moves no page today; `areas.ts` has no top-level `posts.all()`, so it doesn't reopen the import-cycle trap that `urlsByFile()` documents; `ApiNavigation` tolerates absent headings, so `Default.astro` rendering it is safe.
> 
> One caveat on all of it: nothing on the branch declares `area:` yet (0 pages, and 0 `paged:` pages), so the three correctness items are latent. They land on whichever page uses the escape hatch first.

EDIT: Ignore the Claude review, time for a human review, (i removed the comments)

src/lib/areas.ts
```diff
@@ -0,0 +1,171 @@
...
+export function buildAreaCrumbs(
+  currentUrl: URL,
+  area: Area,
+  extraCrumbs?: ReadonlyArray<Crumb> | null
+): Crumb[] {
+  const crumbs = buildCrumbs(currentUrl, extraCrumbs);
+  const { path, sectionCrumb } = AREAS[area];
+
+  if (path === null || sectionCrumb === null) return crumbs;
+  if (!currentUrl.pathname.startsWith(SITE.subfolder + path)) return crumbs;
```
enf0rc3: `buildAreaCrumbs` receives `area` and then decides from the path, so the nav and the breadcrumbs can disagree.

Two cases:

- `/docs/api-and-integration/webhooks` with `area: api` gets an "Api" crumb spliced in, because this line has no path-boundary check — the one `areaFromPath` makes at line 62 (`pathname === prefix || pathname.startsWith(prefix + '/')`).
- `/docs/guides/api-tour` with `area: api` gets the API nav from `AreaNavigation` and no "Api" crumb, since `startsWith('/docs/api')` is false.

The second case is the one that bites. A page under `/docs/api` already gets the API area from path inference, so `area:` frontmatter only earns its keep when the path disagrees with the area — and that is exactly when this line drops the crumb. For the pages the override exists to serve, the `area` argument has no effect.

Deciding lines 158-159 from `area`/`AREAS[area]` alone (or reusing `areaFromPath`'s boundary rule) would cover both.

src/lib/areas.ts
```diff
@@ -0,0 +1,171 @@
...
+function areaOverrides(): Map<string, Area> {
+  if (overrides !== null && import.meta.env.PROD) return overrides;
+
+  const map = new Map<string, Area>();
+  for (const post of accelerator.posts.all()) {
+    const declared = post.frontmatter?.area;
+    if (declared == null || post.url == null) continue;
+
+    const area = resolveArea(declared, post.url);
+    if (area !== areaFromPath(post.url)) map.set(trimSlash(post.url), area);
```
enf0rc3: This keys the map on raw `post.url`, while `areaResolver` is queried with NavPage urls that `mapNavPage` has already rewritten: it appends `/1/` when `frontmatter.paged` is set and replaces the url with `frontmatter.redirect` for `layout: src/layouts/Redirect.astro` (astro-accelerator-utils `navigation.mjs:377-391`).

There are 1402 `Redirect.astro` pages on this branch, so the first one to declare `area:` misses its own override and `areaFromPath` decides instead. Concretely, a paged page at `/docs/foo` with `area: api` is looked up as `/docs/foo/1`, so `withoutOtherAreas` keeps it in the docs nav while `apiMenu()` (reading frontmatter directly via `pageArea`) also lists it — the page appears in both trees, which the `apiNavigation.ts` header says never happens.

Smaller, same path: `isArea` at line 48 uses `value in AREAS`, so `area: constructor` (or `toString`, `valueOf`, `hasOwnProperty`) validates silently. `AREAS[area]` then yields `Object.prototype.constructor`, `path` destructures to `undefined`, the `=== null` guards at 158 fall through, and 159 compares against the string `'/docsundefined'`. `areaOverrides` still records the bogus value, so the page is pruned from the docs nav and never matches `'api'` for the API nav — it drops out of both, which is the outcome the comment at line 70 promises to avoid. `Object.hasOwn(AREAS, value)` closes it.

src/lib/areas.ts
```diff
@@ -0,0 +1,171 @@
...
+ * A resolver for a caller that has urls and nothing else; anything holding the
+ * page itself should use pageArea() and skip the lookup.
+ *
+ * One resolver per pass over a set of urls. The overrides behind it are read
+ * once per resolver, which is the point: collecting them walks the whole page
+ * set, and in dev that set is re-read from disk every time, so a url-at-a-time
+ * lookup turns a tree walk into thousands of them.
+ *
+ * They are collected on first use rather than up front, because reading the
+ * page set is also what pulls the nav tree's lazy children into being, and
```
enf0rc3: Both halves of this justification look wrong, which matters because the comment is what a future edit will trust.

`autoMenu` assigns children eagerly — `page.children = this.getChildren(page, pageList)` at `navigation.mjs:168`, which runs before `withoutOtherAreas` ever sees the tree. And `Posts.all()` is backed by a disk JSON cache (`posts.mjs:25`, `cache.mjs:19`), so it hands back the same order on every call. On top of that, the first lookup lands in `prune`'s `.filter` over the top-level array, which completes before any `node.children` is touched, so even genuinely lazy children would not be pulled in early enough to reorder navOrder ties.

The deferral itself does no harm. It is the stated reason that would send the next reader down the wrong path.

src/lib/areas.ts
```diff
@@ -0,0 +1,171 @@
...
+// deliberately not the layout. The layout decides what the middle column looks
+// like; the area decides which nav sits beside it and how the breadcrumbs read,
+// so a page can be in the API area and still be laid out like a docs page.
+//
+// A page's area comes from where it lives, which is what the section already
+// encodes: every page under /docs/api is in the API area without being told so,
+// including the ~100 generated ones. A page that needs to say otherwise puts
+// `area:` in its frontmatter and that wins.
+//
+// Adding an area takes three things: an entry below, a case in
```
enf0rc3: Three places in this PR give three different answers to "what does adding an area cost", and all three leave out `frontmatter.json`:

- here: three things (an entry, a case in `AreaNavigation.astro`, the nav component)
- `AreaNavigation.astro`: "a new area adds a branch here and nothing else"
- neither mentions the `area` field's `choices` in `frontmatter.json` (lines 79-82), which also needs the new value — miss it and the build succeeds while authors simply cannot pick the area in Front Matter CMS, with nothing to indicate why

Since the description commits to Learn and CLI areas, this checklist will get followed for real. Worth settling on one version. Also `AreaNavigation.astro` is a ternary today, so the third area forces a restructure before it can add its branch.

src/lib/areas.ts
enf0rc3: Narrowing this one — the second case should come off. Breadcrumbs tracking the URL while the nav tracks the area is a coherent split, and the description's inference rule reads as deliberate: a page at `/docs/guides/api-tour` genuinely sits under Docs > Guides whatever nav it shows. My "the `area` argument has no effect" line overstated it.

The first case still stands, and it is narrower than I wrote. `areaFromPath` requires the prefix to land on a path boundary and its doc comment gives the reason outright — "or /docs/api-and-integration would read as part of the API reference". Line 159 does a bare `startsWith(SITE.subfolder + path)`, so a page at `/docs/api-and-integration/webhooks` declaring `area: api` picks up the "Api" crumb that line 62 exists to stop. One rule, applied in one place and skipped in the other. Adding the boundary check here would line them up.

enf0rc3 (approved):
Looks good, i see the benefits of this for nav, really cleans it up

PR #3400, opened 2026-08-20
Stop the Playwright install step hanging the build

enf0rc3:
`playwright install --with-deps` shells out to apt-get, which stalls when the Ubuntu mirrors are slow on a GitHub-hosted runner. Nothing bounds it, so the job runs to the 6 hour Actions ceiling. 

Only Chromium is configured, so install that alone and drop the apt call. This also skips downloading Firefox and WebKit, which were never used.

Also cuts the per-test timeout from 30 minutes to 60 seconds. The whole suite runs in 34 seconds, with the slowest test at 14.

The `build` job in OctopusDeploy/microsite-deployment has no `timeout-minutes`, so any future stall there will still run for 6 hours. Worth setting separately.

# Before

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
         "renovate": "renovate",
         "spellcheck": "git fetch origin main:refs/remotes/origin/main && git diff origin/main --name-only --diff-filter=ACMRTUXB | cspell --no-must-find-files --file-list stdin",
         "spellcheck-all": "cspell \"**/*.{md,mdx}\"",
-        "test": "astro build && npx playwright install --with-deps && npx playwright test",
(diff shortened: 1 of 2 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3400.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3400&status=all).

rosslovas (approved):

PR #3401, opened 2026-08-20
Separate generated API docs from handwritten ones

rosslovas:
# Summary

We intend to have both generated and handwritten files in the `/docs/api` folder.

Most of the files are generated. Having them side-by-side would be confusing:
- It's hard to tell what's handwritten and what's not
- Difficult for the sync process: it needs special logic to not mess with handwritten files
- Difficult for humans: accidentally making edits to generated files not knowing they'll be blown away

This PR allows us to put the generated files in a `_generated` subfolder which makes managing them much easier, and makes it more obvious that handwritten changes to these files won't persist.

# Results

https://stoctodocspr3401.z22.web.core.windows.net/docs/api/channels works,

https://stoctodocspr3401.z22.web.core.windows.net/docs/api/_generated/channels 404s.

Effectively, no visible changes, just files moved in the repo with no effect on the built site.


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3401.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3401&status=all).

src/pages/docs/api/[...generatedFileName].astro
Comment (rosslovas) [inline src/pages/docs/api/[...generatedFileName].astro:1]: This file is the crux of this PR, it's what enables the path `/docs/api/channels` to use the content of `/docs/api/_generated/channels.md` on disk for example.

Note that folders that start with `_` like `_generated` aren't accessible by convention, so the path `/docs/api/_generated/channels` 404s without me having to do anything.

src/lib/accelerator.ts
```diff
@@ -37,23 +41,40 @@ const readAll = posts.all.bind(posts);
...
+// Astro's own routing rule is the filter, with one exception: /docs/api/_generated
+// contains generated API docs which is published from /docs/api/ by [...generatedFileName].astro.
 const PAGES_ROOT = '/src/pages/';
-const isRouted = (post: { file?: string }) => {
+
+// `file` is an absolute path. We only care about the part below `src/pages`.
+const routePathOf = (post: { file?: string }) => {
   const file = post.file ?? '';
-  // Only the path below src/pages decides routing. In local development the
-  // generated API pages are symlinked in, so `file` resolves to wherever the
```
rosslovas, line 45: I assume symlinks were used at some point during development but not in practice.

src/lib/apiNavigation.ts
```diff
@@ -12,49 +12,14 @@ import { pageArea } from './areas';
...
-function urlFromSourcePath(path: string): string {
-  return path
-    .slice(PAGES_ROOT.length)
-    .replace(/\.md$/, '')
-    .replace(/\/index$/, '');
-}
-
-// Astro only gives a page module a `url` when its file resolves inside
-// src/pages. The generated pages are usually copied in, but they are symlinked
-// in during local development, and a symlink resolves to wherever the generator
```
rosslovas, line 26: This was also symlink related and I don't believe is necessary any more

borland (approved):
Tested on staging site, comments in the file explain it nicely. Approved

PR #3406, opened 2026-08-20
Document Tentacle restart requirement for poll-server

harrisonmeister:
The poll-server page didn't mention that the Tentacle service needs restarting before a new server address takes effect, so add a warning callout and show the restart in the basic example.

The HA page attached the restart requirement only to the Tentacle.config method and offered the command line as the alternative that avoided it, which reads as though poll-server applies immediately. Move the restart note so it covers both methods, add the restart step to the command line example, and discourage the Tentacle.config method on the grounds that actually apply to it.

### DIFF
--- a/src/pages/docs/administration/high-availability/polling-tentacles-with-ha.mdx
+++ b/src/pages/docs/administration/high-availability/polling-tentacles-with-ha.mdx
@@ -1,7 +1,7 @@
 ---
 layout: src/layouts/Default.astro
 pubDate: 2023-01-01
-modDate: 2023-01-01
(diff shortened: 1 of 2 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3406.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3406&status=all).

adamoctoclose (approved):
LGTM

PR #3408, opened 2026-08-20
Speed up search by ranking from a title map

enf0rc3:
## The problem

🤖 AI Generated PR 🤖 

Putting results in order needs each result's URL and title. Pagefind keeps both inside that result's own file, one file per result.

That tied two things together: **how many results we could order was the same number as how many files we downloaded.** We downloaded 30 files on every search to order 30 results, before a single row could be drawn.

## The change

The build writes one map of result id → URL and title, beside the index.

Ordering reads the map, so it can order every match without downloading anything. Result files are then downloaded only for the rows actually on screen.

## Why the map, and not just fewer downloads

Cutting downloads from 30 to 10 is the obvious fix. On its own it drops top-five accuracy from 84% to 71%, because only the results you download can be ordered.

The map breaks that link. Ordering reads the map, so every match is ordered however few files come down. This PR downloads 10 and holds 84%.

## What does not change

Matching. Pagefind's word index is untouched, so the full body text of every page is searched exactly as before — a phrase from the middle of a page still finds it. The map holds only a URL and a title per page, and is read after matching to decide what order the matches go in.

## Effects

| | before | after |
| --- | --- | --- |
| requests per search | 36 | 16 |
| first search, Slow 4G | 8.8 s | 7.3 s |
| second search, same session | 5.9 s | 4.1 s |
| relevance, both query sets | 57/84, 90/98 | unchanged |

The map is 29 KB gzipped, loaded once per page. The first screen draws 10 rows where it drew 30, so 6 of 155 measured queries need one scroll to reach their answer — none ranked any lower.

### DIFF
--- a/src/integrations/pagefind-index.ts
+++ b/src/integrations/pagefind-index.ts
@@ -87,7 +87,7 @@ export default function pagefindIndex(): AstroIntegration {
           const titles = await writeTitleMap(
             path.join(distDir, 'docs', 'pagefind')
           );
-          logger.info(`wrote ${titles} titles to ${TITLE_MAP}`);
+          logger.info(`wrote ${titles.pages} titles to ${titles.file}`);
         } finally {
           await close();
         }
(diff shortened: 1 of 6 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3408.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3408&status=all).

borland (changes requested):
While this is technically very amazing, and I can't fault the improved outcome, we're essentially writing code that should be private inside pagefind. The fact that we have to unzip the compressed pagefind fragments, understand their structure, and more makes this incredibly brittle. If pagefind vNext moves things even slightly, it's all going to blow up. 

That's an awful lot of complexity and risk to be handing over to the web team; I'm not comfortable with it.

How about, leave pagefind alone, but instead of loading 30 rows in the search dialog, load 20. That will get us some of the improvement, and I'd guess people probably don't read beyond the first 20 anyway

enf0rc3:
Sounds fair to me! thanks for taking a look @borland

PR #3425, opened 2026-08-27
Fix: document the account saved on discovered cloud targets

HuyPhanNguyen:
# Background

Cloud target discovery saves the account it authenticated with onto every target it creates, and that account is what Octopus then uses to deploy to the target and run its health checks. The page never said so. It describes `Octopus.Azure.Account` as an account "to use when discovering cloud targets", which reads like the account is only used for the scan.

FD-731 came in because of that. Someone took the account out of their deployment and couldn't work out why an Azure Web App target still pointed at it, or whether they had to go and change the target themselves. Two other spots on the page steer you wrong: the switch-off instructions stop at "discovery is now off" and never mention existing targets, and the stale-target warning says Octopus removes them automatically. It only does that when a target fails its health checks.

# Results

Says where a discovered target's account comes from, and what happens to it when you change or remove the discovery account. Also fixes a broken in-page link.

Fixes FD-731

## Before

- `Octopus.Azure.Account` and `Octopus.Aws.Account` read as credentials for the discovery scan only.
- "Switching off ... cloud target discovery" ends once you delete the variable. Nothing about targets already discovered.
- "Previously discovered targets" says the target "will simply be updated" without saying what gets replaced.
- The stale-target warning implies Octopus cleans up: "the old target will become unhealthy and be removed automatically".
- `[Cleaning up unhealthy targets]` had no link destination.

## After

- Both variable tables say the account is saved on each discovered target and used for that target's deployments and health checks.
- Both switch-off sections warn that existing targets keep their account.
- New "Changing or removing the discovery account" section: rediscovery under a different account updates the target, because the account isn't part of the target name; removing the account switches discovery off, so nothing updates the existing targets; and a target holding an account blocks deleting that account.
- The auto-removal sentence now says removal only happens on health check failure.
- Link fixed to `#cleaning-up-unhealthy-targets`.

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ ] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [ ] I have considered the [appropriate target version for this PR](https://octopushq.atlassian.net/wiki/spaces/RND/pages/1145896982/Which+versions+should+I+patch#Process).
- [ ] I have considered appropriate testing for my change.
  - Automated testing / Exploratory testing / Nothing required?
- [ ] I have considered manually testing my changes on a [branch instance](https://octopushq.atlassian.net/wiki/spaces/RND/pages/1262256152/Use+a+Core+Feature+Branch+Instance) to check correctness, stability and performance on Octopus Cloud.
- [ ] I have considered safety nets to reduce any time-to-recovery for my change.

### DIFF
--- a/src/pages/docs/infrastructure/deployment-targets/cloud-target-discovery/index.md
+++ b/src/pages/docs/infrastructure/deployment-targets/cloud-target-discovery/index.md
@@ -1,7 +1,7 @@
 ---
 layout: src/layouts/Default.astro
 pubDate: 2023-01-01
-modDate: 2024-06-27
+modDate: 2026-08-27
 title: Cloud Target Discovery
 description: Cloud resources can be discovered and registered as deployment targets by Octopus
 navOrder: 90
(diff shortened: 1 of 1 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3425.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3425&status=all).

grace-rehn (approved):

PR #3426, opened 2026-08-29
docs: add a Buildkite build server page

NickJosevski:
## Why

`getting-started/index.mdx` lists **BuildKite** among the CI systems we "provide first-class integrations with" and give guides for. There is no such page — `build-servers/` documents nine other systems and not this one. Buildkite users following that promise land nowhere.

## What's in it

Everything here was verified end to end against a Buildkite **hosted** agent and an Octopus Cloud instance, not inferred from the API:

- **Installing the CLI.** It is *not* preinstalled on Buildkite agents, including hosted ones, and there's no Buildkite equivalent of `install-octopus-cli`. The page gives the tarball recipe that was actually used.
- **OpenID Connect**, with the full token-exchange script, since there's no login plugin for Buildkite yet.
- **Creating a release**, via the `create-release` plugin or the CLI directly.
- **Build information** — honestly marked as not yet having a plugin.
- **Troubleshooting**, drawn from the failures actually hit.

## The bit I'd most want reviewed

Buildkite's subject claim pins the commit SHA:

```
organization:ORG:pipeline:PIPELINE:ref:REF:commit:COMMIT:step:STEP_KEY
```

So a subject copied verbatim from a working build authorises **exactly that build** and then silently never matches again — with an error that gives no hint which half is wrong. `commit:*` is effectively mandatory. That's in a warning callout, and it's the single thing most likely to cost a customer an afternoon.

It's also the strongest argument for eventually giving Buildkite a guided issuer type in the portal like GitHub Actions has, rather than leaving people to hand-write a five-segment subject in a free-text field.

## Also

- Adds Buildkite to the list on `build-servers/index.md` (alphabetical, `navOrder: 45` between BitBucket Pipelines and Codefresh).
- All 7 internal `/docs/...` links verified to resolve to real pages.

## Known gaps this page documents rather than fixes

No login plugin, no build-information plugin, no CLI-install plugin for Buildkite. The page tells the truth about each rather than omitting them.

### DIFF
--- a/src/pages/docs/packaging-applications/build-servers/buildkite.md
+++ b/src/pages/docs/packaging-applications/build-servers/buildkite.md
@@ -61,13 +61,13 @@ Give the service account the permissions it needs. For creating releases, the bu
 
 Buildkite builds the subject claim from the job:
 
-```
+```text
 organization:ORG:pipeline:PIPELINE:ref:REF:commit:COMMIT:step:STEP_KEY
 ```
 
 For example:
 
-```
+```text
 organization:acme:pipeline:my-app:ref:refs/heads/main:commit:9f3182061f1e2cca4702c368cbc039b7dc9d4485:step:build
 ```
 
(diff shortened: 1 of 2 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3426.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3426&status=all).

octonautcal (approved):

PR #3428, opened 2026-08-30
Fetch 20 search results a page instead of 30

enf0rc3:
Each row the search overlay draws costs a fragment fetch, so `PAGE_SIZE` is what a query pays before it can show anything. Dropping it from 30 to 20 cuts that by a third.

Twenty still leaves `byNameThenDepth` enough of the list to reorder, and the shallow-page promotion that rescues a page ranked past the page size is untouched. Two comments that quoted the old number are corrected.

Pagefind is left alone. This is the cheap part of the search speed-up, split out from #3408 so it can land on its own.

All 30 tests in `tests/docs-search.spec.ts` pass, including the three that depend on the page size: "scrolling to the end of the results loads more", "a page pulled forward is not listed again further down", and "the page a query names comes first".

### DIFF
--- a/src/scripts/search-engine-pagefind.ts
+++ b/src/scripts/search-engine-pagefind.ts
@@ -12,10 +12,12 @@ import {
   type SearchResult,
 } from './search-engine';
 
-// Rows fetched at a time. Pagefind's own UI shows five and offers the rest on
-// demand; thirty, because `byNameThenDepth` reorders within a page and needs
-// enough of the list to have something to reorder.
-const PAGE_SIZE = 30;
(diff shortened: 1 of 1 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3428.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3428&status=all).

borland (approved):

PR #3429, opened 2026-08-30
Find the page a query names from a title list, not a second search

enf0rc3:
A section's own page can rank far below the pages inside it: `/docs/infrastructure/deployment-targets/` is 36th on raw score for "deployment targets". The overlay rescues it by promoting the page a query names, and `claimsName` decides that from a URL and a title alone.

Pagefind holds both, but only inside a result's fragment, one fetch each. That is why the promotion needed a second search narrowed to shallow pages by a `landing:true` filter, and why it only ever reached pages three segments deep.

Both are known from the site's own markdown at build time. `/docs/search-titles.json` now carries them for every indexed page, built from the same `import.meta.glob` that `llms.txt.ts` and `sitemap.xml.ts` already walk. The lookup needs no fetches and no second search, and any page can be promoted from any depth.

Removed with it: the `landing:true` filter, its `data-pagefind-filter` in both layouts, the third `api.search` per query, and `LANDING_CANDIDATES`.

Pagefind is used only through its public client API. Nothing here reads its index format.

### Measured from the build

- 1258 entries, 112 KB raw, 19 KB gzipped, fetched once alongside the index
- 0 list entries are absent from the pagefind index, so a promoted row always has a page behind it
- The only indexed page with no entry is `/docs/` itself, which has no frontmatter title

### Notes for review

- The promoted row has no excerpt, because it has no fragment. `DocsSearch.astro` already collapses an empty excerpt, so the row shows its title and breadcrumb. Carrying a description per page would have roughly tripled the file for one row on a minority of queries.
- `namedPage` rejects a page whose facet does not match the active tab. The list covers the whole site while a tab narrows results to one section.
- Redirect stubs are excluded. Pagefind drops them for having no article, so promoting one would link somewhere the search cannot reach.
- URLs are normalised to a trailing slash to match what Pagefind returns. The two are compared as strings.

All 30 tests in `tests/docs-search.spec.ts` pass.

Follows #3428. Replaces the approach in #3408.

### DIFF
--- a/src/layouts/Api.astro
+++ b/src/layouts/Api.astro
@@ -99,7 +99,7 @@ const searchIndex = searchIndexAttributes(Astro.url.pathname, frontmatter);
               /* Copy as markdown temporarily disabled until we can get it working with the API docs. Deliberately no "Edit on GitHub" because these are generated and should not be hand edited */
             }
           </div>
-          <div class="page-content anim-show-parent" {...searchIndex.content}>
+          <div class="page-content anim-show-parent">
             <slot />
             <Authors frontmatter={frontmatter} lang={lang} />
             <Taxonomy frontmatter={frontmatter} lang={lang} />
--- a/src/layouts/Default.astro
+++ b/src/layouts/Default.astro
@@ -100,7 +100,7 @@ const searchIndex = searchIndexAttributes(Astro.url.pathname, frontmatter);
               lang={lang}
             />
           </div>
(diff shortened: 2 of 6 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3429.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3429&status=all).

PR #3430, opened 2026-08-30
Move administrative pages out of the Octopus REST API section

enf0rc3:
## Background

The Octopus REST API section is being split up, with the API and the CLI moving to their own top level sections. These pages are about administering Octopus with command line tools, so they don't belong in either.

Migration API is left where it is. Its content is already folded into `/docs/api/migrations`, so it gets deleted rather than moved.

## Before

- `/docs/octopus-rest-api/octopus.server.exe-command-line`
- `/docs/octopus-rest-api/tentacle.exe-command-line`
- `/docs/octopus-rest-api/calamari`
- `/docs/octopus-rest-api/octopus.migrator.exe-command-line`

## After

- `/docs/administration/octopus.server.exe-command-line`
- `/docs/administration/tentacle.exe-command-line`
- `/docs/administration/calamari`
- `/docs/administration/octopus.migrator.exe-command-line`

Redirects are left at every old URL, and the existing `api-and-integration` stubs now point straight at the new destinations instead of chaining.

## Everything in this PR that is not the move

All 61 moved pages are byte-identical to their old copies apart from the link rewrites. On top of that:

- `navOrder` on the four landing pages, to place them in the Administration strip: server.exe 110, tentacle.exe 115, Calamari 120, migrator.exe 125.
- `administration/data/data-migration.md` gains one link, to the migrator command line reference.
- `tests/code-block.spec.ts` pins `octopus.server.exe-command-line/configure` as its long-code-block page; the path was updated.
- `tests/docs-search.spec.ts` searched `kubernetes` to prove the API tab shows rows behind its count. Calamari was the only page in the API facet matching that term, so the query is now `runbook`, which several pages in every tab match.
- `infrastructure/deployment-targets/tentacle/windows/index.mdx` shows as a whole-file diff. Two links changed; the rest is CRLF normalising to LF, which is what `.gitattributes` asks for.

`api/_generated/migrations.md` still points at the old migrator URLs. It is generated from the Octopus Server XMLdoc, so it is fixed upstream in OctopusDeploy/OctopusDeploy#46971; the links resolve through the redirects until that lands.

### On the secret scanner

No key was removed or added. `API-YOUR-KEY` appears 513 times before and after. The moved pages read as new files, because the old paths still exist as redirect stubs, so every `--apiKey=API-YOUR-KEY` example counted as an added line.


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3430.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3430&status=all).

borland:
Spot-checked some things on the review site. LGTM

borland (approved):

PR #3431, opened 2026-08-31
Move the CLI reference to its own top level section

enf0rc3:
Moves the CLI reference out of the "Octopus REST API" section and into its own top level area at `/docs/cli`, alongside the work already done for `/docs/api`.

Pairs with **OctopusDeploy/cli#715**, which points the docs sync at the new location.

### What moves

| From | To | Count |
|---|---|---|
| `/docs/octopus-rest-api/cli` | `/docs/cli` | overview |
| `/docs/octopus-rest-api/cli/octopus-*` | `/docs/cli/octopus-*` | 185 |
| `/docs/octopus-rest-api/octopus-cli/*` | `/docs/cli/octopus-cli/*` | 40 |

The side nav follows the API reference's shape — hand-written pages, a rule, then the generated reference:

```
Overview
Octo Command Line (CLI) (Deprecated)   ← 40 pages
────────────────────────────────────
octopus
octopus account
...  185 commands
```

### Redirects

Two redirect pages, using `src/layouts/Redirect.astro` the same as the rest of the site. Everything that used to sit under them is deleted rather than redirected — these are reference pages for a CLI that moved, and they are not worth a file each to keep alive.

| URL | Result |
|---|---|
| `/docs/octopus-rest-api/cli` | → `/docs/cli` |
| `/docs/octopus-rest-api/cli/<any command page>` | 404 |
| `/docs/octopus-rest-api/cli/octopus_account` (legacy underscore) | 404 |
| `/docs/octopus-rest-api/octopus-cli` | → `/docs/cli/octopus-cli` |
| `/docs/octopus-rest-api/octopus-cli/<any sub page>` | 404 |
| `/docs/octopus-rest-api/octo.exe-command-line/*` | 404 |
| `/docs/api-and-integration/octo.exe-command-line/*` | 404 |

That last row is a 2023 folder of 44 stubs pointing at the Octo CLI pages. It goes for the same reason as the rest.

427 stub files removed in total, which takes the build from 3,017 pages to 2,653.

### What changed

Aligned the "generated docs" with how we do "API" generated docs, e.g. `_generated` folder, routing etc, adding new layout  `src/layouts/Cli.astro` and update the areas.ts to support having a new `area`.

We had to make some changes to the `apiNavigation` and other menu components to make it generic for "areas", this will be valuable in the future if we move other automated gen docs, e.g. [integrations ](https://octopus.com/integrations)

### The overview page

Previously the "overview" page was generated in the CLI repo and was created when the doc files were generated. I believe this should be generated in the "docs" repo instead of CLI, to be closer to how the Api docs work, where top level pages like overview that has a list of the generated docs, is owned by docs, this should simplify things and make it easier to update and change, we can generated the same overview at build.

This means we will also need to update the CLI repo to no longer generate this file too, thats in a seperate pr to be merged after.  [CLI#715](https://github.com/OctopusDeploy/cli/pull/715)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3431.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3431&status=all).

src/lib/searchIndexing.ts
```diff
@@ -59,7 +60,20 @@ export function searchIndexAttributes(
...
+  // exists to find the section a reader has named, and a command page is not one:
+  // /docs/cli/octopus-deployment-target-list is three segments, and there are 185
+  // like it, enough to crowd out the sections they match in a pool this small.
+  // Before this, "deployment targets" put the CLI's deployment-target commands
+  // ahead of /docs/infrastructure/deployment-targets.
+  //
+  // Keyed on the file rather than on the area, because the pages somebody wrote
+  // still belong in the pool: /docs/cli is the section a reader searching "cli"
+  // means, the same way it was back when it lived under /docs/octopus-rest-api.
+  const isLanding =
```
enf0rc3: **Why this line exists.**

`landing:true` is not "is this page in search". Every CLI page is indexed and findable — they carry `section:cli` and no `data-pagefind-ignore`, they are in the sitemap (226 entries) and in `llms.txt`. The under-construction exclusion applies to `/docs/api` only, and CLI never gets it.

This filter marks a much smaller shortlist: the pages the overlay's *second* search draws on to rescue a page that names the query but ranks past `PAGE_SIZE`. See `search-engine-pagefind.ts`:

```js
const named = hits.some((hit) => claimsName(hit, query))
  ? null
  : await namedPage(landing.results, query, hits);
```

It is only consulted when nothing on the first page already names the query.

**What went wrong without it.** The command pages moved from four URL segments to three, which put all 185 of them under `LANDING_DEPTH` for the first time. At their old URL they were never in this pool. The result: searching `deployment targets` returned the CLI's `deployment-target` commands above `/docs/infrastructure/deployment-targets`, and `docs-search.spec.ts` failed on it.

**Why keyed on the file, not the area.** My first attempt tested `areaFromPath(pathname) === 'docs'`, which threw out the hand-written pages too — `/docs/cli` stopped being promotable, and it was promotable at its old URL. Testing the file keeps everything somebody wrote in the pool and takes only the generated reference out.

**This is consistent with what the file already did.** `search-engine-pagefind.ts:136` predates this PR:

```js
const REFERENCE_FACETS = new Set(['api', 'cli']);
```

Reference pages are already discounted by `REFERENCE_PENALTY = 1.3` and already excluded from claiming a name unless `ASKS_FOR_REFERENCE` matches. That constant was swept over two traffic-weighted query sets.

**Measured, not argued.** Driving the real overlay against this build:

| Query | Top result |
|---|---|
| `deployment targets` | `/docs/infrastructure/deployment-targets/` |
| `cli` | `/docs/cli/` |
| `octopus cli` | `/docs/cli/` |
| `octopus release create` | `/docs/cli/octopus-release-create/` |
| `octopus account` | `/docs/cli/octopus-account/` |
| `create release` | `/docs/releases/creating-a-release/` |
| `worker pools` | `/docs/infrastructure/workers/worker-pools/` |
| `octopus deployment-target list` | `/docs/cli/octopus-deployment-target-list/` |

The last row is the one that matters here: a generated page, excluded from this pool, still ranks first when the query names it. It wins on raw score and never needed rescuing.

src/scripts/search-engine.ts
```diff
@@ -74,18 +74,20 @@ export const FACETS: Facet[] = [
...
+// Longest-match-first: the .exe command line references still sit under the REST
+// API tree, so that CLI prefix has to be tested before the API one it is nested
+// in. The Octopus CLI itself has its own section at /docs/cli and nests in
+// nothing.
 const SECTIONS: {
   facet: string;
   kind: SearchResult['kind'];
   prefix: RegExp;
 }[] = [
+  { facet: 'cli', kind: 'command', prefix: /^\/docs\/cli(\/|$)/ },
```
enf0rc3, line 82: **Why the facet rules changed shape.**

Before, one pattern caught everything CLI because it all lived under the REST API tree:

```js
/^\/docs\/octopus-rest-api\/(cli|octopus-cli|[a-z.]+-command-line)(\/|$)/
```

The Octopus CLI has left that tree, so it gets a prefix of its own that nests in nothing. `/docs/cli` covers the overview, the 185 commands and the deprecated Octo CLI pages in one.

The `.exe` command line references (`octopus.server.exe`, `tentacle.exe`, `octopus.migrator.exe`, `octo.exe`) are still under `/docs/octopus-rest-api`, so they keep a pattern there — and it still has to be tested before the `api` prefix it nests inside. That ordering constraint is why the comment above was reworded rather than deleted.

Net effect: CLI results keep landing under the CLI tab, from the new URLs. `docs-search.spec.ts` covers it — the tab-count assertion (`All (4)`, `Docs (2)`, `API (1)`, `CLI (1)`) still passes.

Worth flagging for whoever finishes the API move: `/docs/api` matches no rule here, so those pages classify as `docs`. That is invisible today because the section is excluded from the index by `underConstruction.ts`, but it will need a prefix when the section goes live.

enf0rc3 (commented):
Notes on the two search-index changes, since they are the least obvious part of this PR.

src/layouts/Cli.astro
```diff
@@ -0,0 +1,141 @@
...
+    <TopNav lang={lang} showSearch={showSearch} />
+    <div class="content-group">
+      <main id="site-main">
+        <Breadcrumbs lang={lang} crumbs={crumbs} />
+        <article {...searchIndex.article}>
+          <ArticleHeader lang={lang} subtitle={subtitle} title={title} />
+          <div class="page-actions">
+            <CopyAsMarkdown lang={lang} />
+            <OpenInLlm lang={lang} />
+          </div>
```
enf0rc3, line 110: Same as Default.astro, but drops the edit in github button, opted for brand new template, because eventually i assume this will have more layout changes

src/scripts/search-engine.ts
```diff
@@ -74,19 +74,12 @@ export const FACETS: Facet[] = [
...
 const SECTIONS: {
   facet: string;
   kind: SearchResult['kind'];
   prefix: RegExp;
 }[] = [
-  {
-    facet: 'cli',
-    kind: 'command',
-    prefix:
-      /^\/docs\/octopus-rest-api\/(cli|octopus-cli|[a-z.]+-command-line)(\/|$)/,
```
enf0rc3, line 88: This can just be `cli` now we have moved the rest of the pages out.

src/scripts/search-engine-pagefind.ts
```diff
@@ -359,6 +359,27 @@ export function pagefindEngine(bundlePath: string): SearchEngine {
...
+        // right section, so use it as it is. Without one, exclude the API and
+        // CLI results: `namedPage` reads only the first few and skips reference
+        // pages anyway, so left in they fill every slot and it finds nothing to
+        // promote. Unless the query asks for the reference, which can promote
+        // one.
+        //
+        // Excluding, rather than naming the two sections to keep, because
+        // Pagefind reads a plain array as "all of these at once". Measured for
+        // "deployment targets": section: ['docs', 'integrations'] returns 0,
+        // section: { not: { any: ['api', 'cli'] } } returns 97.
```
enf0rc3, line 374: This one was a mission to work through, here is the claude dump of what it does:

```
To run it: pnpm dev, open http://localhost:4321/docs, click the search field. Stay on the All tab unless a row says otherwise.

The one that proves the fix

┌────────────────────┬──────────────────────────────────────────┐
│       Search       │                Expect #1                 │
├────────────────────┼──────────────────────────────────────────┤
│ deployment targets │ /docs/infrastructure/deployment-targets/ │
└────────────────────┴──────────────────────────────────────────┘

Without the fix this returns /docs/getting-started/first-deployment/add-deployment-targets/. That's the whole regression — if you see the getting-started page, the fix isn't working.

CLI must still win when you ask for it

┌────────────────────────────────┬───────────────────────────────────────────┐
│             Search             │                 Expect #1                 │
├────────────────────────────────┼───────────────────────────────────────────┤
│ octopus account                │ /docs/cli/octopus-account/                │
├────────────────────────────────┼───────────────────────────────────────────┤
│ octopus release create         │ /docs/cli/octopus-release-create/         │
├────────────────────────────────┼───────────────────────────────────────────┤
│ octopus deployment-target list │ /docs/cli/octopus-deployment-target-list/ │
├────────────────────────────────┼───────────────────────────────────────────┤
│ octopus package upload         │ /docs/cli/octopus-package-upload/         │
├────────────────────────────────┼───────────────────────────────────────────┤
│ octopus login                  │ /docs/cli/octopus-login/                  │
├────────────────────────────────┼───────────────────────────────────────────┤
│ cli or octopus cli             │ /docs/cli/ (the overview)                 │
└────────────────────────────────┴───────────────────────────────────────────┘

These are the cases you were worried about — CLI pages excluded from the shortlist but still top of the All tab. They rank there on merit; the shortlist never comes into it.

Docs must still beat CLI when you don't ask

┌────────────────┬────────────────────────────────────────────────────────────────────────────────────────────┐
│     Search     │                                         Expect #1                                          │
├────────────────┼────────────────────────────────────────────────────────────────────────────────────────────┤
│ create release │ /docs/releases/creating-a-release/ — the guide, not octopus release create                 │
├────────────────┼────────────────────────────────────────────────────────────────────────────────────────────┤
│ worker pools   │ /docs/infrastructure/workers/worker-pools/ — the section, not the CLI worker-pool commands │
└────────────────┴────────────────────────────────────────────────────────────────────────────────────────────┘

Tabs

Click the CLI tab and check the count is non-zero and results are CLI pages:

┌────────────────────┬──────────┐
│       Search       │ CLI tab  │
├────────────────────┼──────────┤
│ account            │ CLI (66) │
├────────────────────┼──────────┤
│ deployment targets │ CLI (48) │
├────────────────────┼──────────┤
│ worker pool        │ CLI (22) │
└────────────────────┴──────────┘

One caveat on those counts: they lag by a beat when you click the tab, so give it a second before reading the rows — my first probe read stale rows and briefly made it look like docs pages were showing under CLI.

One oddity worth knowing, unrelated to this PR: octopus runbook run returns /docs/octopus-rest-api/examples/runbooks/create-and-publish-runbook/ at #1 rather than /docs/cli/octopus-runbook-run/. That's the only CLI query out of eight that doesn't land on its command page. It behaves the same way before this change, so I've left it alone — but it's the sort of thing worth a look if you're auditing REFERENCE_FACETS anyway.
```

src/pages/docs/cli/_generated/octopus-account-aws-create.mdx
borland, line 1: Nice touch copying the `_generated` pattern from the API stuff so people know not to make edits

src/pages/docs/octopus-rest-api/cli/index.md
borland, line 1: We want to keep this redirect at the folder root, but not all the rest!

I understand it would have been claude adding them, and easy enough to do, but they massively clutter up the site, slow the build process down and they're not important enough to keep, except for the root.

src/pages/docs/octopus-rest-api/octo.exe-command-line/index.md
borland, line 1: Woof. I reckon we delete this whole folder of redirects

src/pages/docs/octopus-rest-api/octopus-cli/index.mdx
borland, line 1: As per the `cli` folder, I think we should keep the redirect on `index` but delete the rest to declutter the site

borland (approved):
I looked at all the actual code and it all looked sensible and reasonable.
Tested on the staging site and everything looks good as far as I can tell.

BTW: I merged main into your branch because it merge-conflicted against Shaun's scrolling fix

Approved so you can merge without second approval, but I'd like it if we cleaned up the excessive redirect files.

If you disagree or think my opinion might be controversion, then merge as-is with all the redirects and we can worry about cleaning house later.

PR #3432, opened 2026-08-31
Ensure the area navigation always keeps the active item visible on page load

shaunmarx:
When visiting a page that's deeply nested in the docs the item may not always be visible in the area navigation list. We therefore scroll the active element into view on page load to ensure user's don't have to actively scroll each and every time a page is navigated to.

# Before
https://github.com/user-attachments/assets/0305bcd8-c9af-415b-a5a0-9662ec5ca755

The area navigation list will jump to the top of the list after every page navigaiton.

# After

https://github.com/user-attachments/assets/c3a1c17d-ef2a-4b2b-87bb-7b002d4b2873

The area navigation list will now ensure the active item is visible when the page reloads

### DIFF
--- a/src/components/ApiNavigation.astro
+++ b/src/components/ApiNavigation.astro
@@ -81,7 +81,7 @@ stats.stop();
   aria-label={_(Translations.aria.site_navigation)}
 >
   <h2 class="site-nav-title">{_(Translations.navigation.title)}</h2>
-  <ul class="site-nav__list">
+  <ul class="site-nav__list" data-site-nav-scroll>
     {
       authored.map((page) => (
         <ApiNavigationItem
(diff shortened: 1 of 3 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3432.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3432&status=all).

borland (approved):
Tested on the staging site. So much better!

(Normally it's impolite to merge other people's PR's but because this fix is so simple and because it's so annoying I'm going to merge it now, just to stop innocent customers suffering in the meantime. Thanks!)

PR #3434, opened 2026-08-31
Make this repo the source of truth for docs content standards

samironsoctopus:
## What this does

Moves the Octopus documentation content standards into this repository and makes it their single source of truth. They currently live in both the `octopus-docs-standards` Claude plugin skill and Confluence, and the copies drift.

### Task 1 — the standards

Adds ten standards to `contributing/standards/`, migrated verbatim from the skill.

One difference between the installed skill copies needed reconciling: a stale local copy carried an extra "3 years old or newer" guideline in `version-notes.md` that the committed plugin version and the shipped marketplace build had already dropped. This takes the current version, so that line is not present here.

Also adds `contributing/standards/README.md`, which carries the routing and orientation that previously existed only inside the skill's `SKILL.md` — the content model, how to pick a page type, the guide-versus-tutorial boundary, and a table of which standards apply to which task. Without it, a contributor not using Claude Code got an alphabetical file list and no way in. It routes and orients only, and never restates substance from the ten standards.

### Task 2 — top-level README

Adds a short `Documentation standards` section, and repoints the existing `Docs page layout guidelines` section at the folder. That section previously sent people to the skill plus a Confluence search, which is now the wrong direction.

### Task 3 — CONTRIBUTING

- External contributor instructions now come before Octopus staff instructions.
- Adds a line pointing contributors at the standards before they submit.
- Replaces both `INSERT-LINK-HERE` placeholders with a Confluence search reference, so no broken links ship.

## Lint and spellcheck exclusions

`contributing/standards/**` is excluded from markdownlint and cspell, following the `src/pages/docs/api/**` precedent already present in both configs.

CI checks every changed `.md` file repo-wide, and the migrated files intentionally contain fences without a language, a compact-style example table, and a repeated "Version notes" heading demonstrating two format variants. Linting them would mean editing illustrative content rather than fixing defects, and would break the verbatim migration.

A directory glob was chosen over enumerating the ten filenames deliberately: an enumerated list would need updating in two configs every time a new page-type standard ships, or CI fails on verbatim content.

`README.md` in that folder is our own prose rather than migrated content, so it was checked against both markdownlint and cspell out-of-tree and passes cleanly. It is not covered by CI going forward, which is the tradeoff of the directory glob.

## Sequencing

This needs to merge before the `octopus-docs-standards` skill is repointed at these files, otherwise the skill's references break.

## Reviewer notes

- The ten standards are byte-for-byte identical to the plugin source, verified with `diff`. The only judgment call in them is the `version-notes.md` reconciliation described above.
- The CONTRIBUTING reorder was verified line-by-line to confirm it changed only the four intended lines and dropped nothing.
- All relative links resolve.
- Pre-existing and left alone: CONTRIBUTING refers to "Required checks" and "Deploying to preview environment" as being "below", but both sections live in the top-level README. Happy to fix in this PR or separately.

### DIFF
--- a/.markdownlint-cli2.jsonc
+++ b/.markdownlint-cli2.jsonc
@@ -1,3 +1,3 @@
 {
-    "ignores": ["src/pages/docs/api/**"]
+    "ignores": ["src/pages/docs/api/**", "contributing/standards/**"]
 }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,12 +1,25 @@
 # Contributing to the Octopus docs
 
-Contributions to help improve this documentation are welcome. How you contribute depends on whether you're an Octopus staff member or contributing from outside Octopus.
+Contributions to help improve this documentation are welcome. How you contribute depends on whether you're contributing from outside Octopus or you're an Octopus staff member.
+
+Before you submit, check your change against the [documentation content standards](contributing/standards).
+
+## External contributors
+
+- The `main` branch has the latest version of the docs.
+- Fork this repo and create a branch for your changes.
+- Make the changes you'd like to contribute.
+- Submit a pull request (PR) to `main` with your changes and include a comment explaining the changes.
+- Sign the [Contribution License Agreement (CLA)](https://cla-assistant.io/OctopusDeploy/docs).
+- We'll review your PR and accept it or suggest changes.
+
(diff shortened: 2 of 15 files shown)


team-marketing-branch-protections:
Pull request environment is available at https://stoctodocspr3434.z22.web.core.windows.net.

You can view the [ephemeral environment status in Octopus Deploy](https://deploy.octopus.app/app#/Spaces-2095/projects/docs-microsite/ephemeral-environments?page=1&environmentName=pr3434&status=all).

borland (approved):

